PR #7, opened 2015-05-08
Concatenating structs concatenates members instead of overwriting them

poppolopoppo:
Example :
```
// struct concatenation
{
	.StructA =
	[
		.MemberA = "member A"
	]
	.StructB =
	[
		.MemberA = "zzz" // this will be replaced <- NOW MORE, SEE BELOW
		.MemberB = "member B"
	]
	.StructB + .StructA // NOW .StructB has .MemberA == "member Azzz"

	Using( .StructB )
	Print( "$MemberA$ + $MemberB$" )
}
```

This modification allows the following pattern, which is usefull to reduce BFF verbosity :
```
.DebugConfig 		[ .CompilerOptions = ' /D_DEBUG'  ]
.ReleaseConfig 	[ .CompilerOptions = ' /DNDEBUG'  ]

.X86Platform 		[ .CompilerOptions = ' /D_MX86_' ]
.X64Platform 		[ .CompilerOptions = ' /D_MX64_' ]

.X86Debug 		= .X86Platform + .DebugConfig   // ' /D_MX86_ /D_DEBUG'
.X86Release 		= .X86Platform + .ReleaseConfig // ' /D_MX86_ /DNDEBUG'

.X64Debug 		= .X64Platform + .DebugConfig   // ' /D_MX64_ /D_DEBUG'
.X64Release 		= .X64Platform + .ReleaseConfig // ' /D_MX64_ /DNDEBUG'
```

Signed-off-by: poppolopoppo <frostiebek@gmail.com>


ffulin:
This behaviour is more useful than the current behaviour, which is cool.  The only concern is that we'll be breaking backwards compatibility, which I generally never want to do.

I need to think about how best to manage this - the break might be worth it for the syntactic enhancement this provides.

PS - please create pull requests against 'dev' only - pull requests into 'master' will not be accepted as this is the canonical git repo as imported from Perforce.

poppolopoppo:
Sorry for the wrong pull request, I am still learning how to use github.

Maybe introducing a new operator ?
```
.C = .A | .B
```
But + feels more natural, and ironically | better represents the current behavior.

When I tried to factorize my BFF I also tried something like this :
```
.MSVC   = [  .CompilerOptions = ' /SomeFlags' ] ; declares $CompilerOptions$
.Debug  = [  .CompilerOptions + ' /D_DEBUG'   ] ; tries to append

.X86Debug = 
[
    Using( .MSVC  )                     ; declares $CompilerOptions$
    Using( .Debug )                    ; tries to append
    .CompilerOptions + ' /D_MX86_'     ; still available for modification in this scope
]
```
This was the way I would have like to write my config at first, but the declaration of .Debug is invalid and would be tricky to implement.

Maybe a way of integrating this in the language could be :
```
Mixin( .Debug ) ; declares mixin .Debug
{
    ; inside callee scope here
   .CompilerOptions + ' /D_DEBUG'
}
.X86Debug = 
[
    Using( .MSVC )
    Using( .Debug ) ; not introducing another function
]
```

Which would enable this :
```
; myproject.bff

Mixin( .ProjectConfig )
{
    ; myproject is the only project using this external
    .CompilerOptions + ' /I"path/to/external"'
    .LinkerLibraries + { "External.lib" }
}
#include "binary.bff"
```
```
; binary.bff

; each call to Mixin concatenates to previous declarations
Mixin( .ProjectConfig )
{
    ; still appending some global config
    .CompilerOptions + ' /DSomeFlag'
}
ForEach( .Config in .Configs )
{
    Using( .Config )
    Using( .ProjectConfig ) ; now able to append here
    Library( '$ProjectName$-Lib-$Platform$-$Config$' )
    {
       ; ...
    }
}
```
This is more intrusive and the language is getting trickier with that addition -that's why I finally chose the struct concatenation way- but also more powerful (and maybe more like your approach with ForEach).

ffulin:
I hope you didn't close this because it's taking too long to integrate :)

I am planning on taking your change - I think the change in behaviour makes it more sensible to the point where it can almost be considered a bug fix.

poppolopoppo:
Not at all ;) 
The pull request was automatically closed when I broke up my fork in several branches.
I'll make a new pull request with this commit, but this time descending from your dev branch.

PR #9, opened 2015-05-22
Concatenating structs concatenates members instead of overwriting them

poppolopoppo:
Same commit than https://github.com/fastbuild/fastbuild/pull/7, but descending from you dev branch and with the merge of the version 0.81.


ffulin:
I've merged this into dev.  I've also marged it into the main Perforce depot, cleaned up the formatting a little and added some tests. This should be in the v0.84 release.

I consider this change in behaviour to be a fix, and v0.84 will carry a note about that.

Thanks again for the change :)

PR #10, opened 2015-06-08
Make bootstrap compilation script work again

inequation:
I've updated compile.sh to be able to bootstrap FASTBuild on a machine without previous FB install. Also updated .gitignore to skip the bin/ directory.


ffulin:
Hi there.  

Thanks for these changes, but actually the compile.sh is deprecated. It should have actually been removed from the github repo, but due to a mistake on my part, it wasn't deleted.

FASTBuild can be compiled using itself on all platforms.

Cheers!

- Franta

inequation:
Hehe, in that case, I've adjusted the pull request accordingly. :)

PR #12, opened 2015-06-22
Fixing incorrect compiler message error 1106, stating /c is missing

inbilla:
Under msvc when -c is used, the incorrect error 1106 message appears. 
MSVC supports both - and / as parameter flags. 

Cmake makes heavy use of this assumption, so support isn't possible without resolving this issue.


ffulin:
Looks good - thanks for the fix.

PR #13, opened 2015-06-25
New function Export()

poppolopoppo:
Export( .var ) is the counter part of Using( .var ).
It will update the previous definition of .var in the first ancestor scope declaring it.
If a previous declaration of .var is not found the parent scope will be used to declare a new .var.

This addition allows the following idiom :
```
.Defines = { '_DEBUG', 'SOMETHING=1' }
.CompilerOptions = ''

#if __WINDOWS__
ForEach( .name in .Defines )
{
   .CompilerOptions + ' /D "$name$"'
   Export( .CompilerOptions )
}
#endif

#if __LINUX__
ForEach( .name in .Defines )
{
   .CompilerOptions + ' -D$name$'
   Export( .CompilerOptions )
}
#endif

Print( "$CompilerOptions$" ) ; windows : ' /D "_DEBUG" /D "SOMETHING=1"'
                             ; linux   : ' /D_DEBUG -DSOMETHING=1'
```

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/errors.html
+++ b/Code/Tools/FBuild/Documentation/docs/errors.html
@@ -676,6 +676,24 @@ Unity( "name" )
 </div>
 </div>
 
+<!--------------- 1060 --------------->
+    <div class='newsitemheader'>1060 - Can't modify frozen variable '%s'.</div>
+    <div class='newsitembody'>
+An attempt was made to assign a value to a frozen variable.
+<h4>Example Config:</h4>
+<div class='code'>.Array = { 'a', 'b', 'c' }
+ForEach( .Item in .Array )
+{
+  ^Array + 'loop variables are frozen'
+}
+</div>
+<h4>Example Output:</h4>
+<div class='output'>C:\Test\fbuild.bff(4,3): FASTBuild Error #1060 - Can't modify frozen variable '.Array'
+  ^Array + 'loop variables are frozen'
+  ^
+  \--here
+</div>
+</div>
 
 
 <h2 id='1100'>1100 - 1199 : Function Parameter Errors</h2>
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
@@ -36,6 +36,7 @@
 //------------------------------------------------------------------------------
 BFFParser::BFFParser()
 : m_SeenAVariable( false )
+, m_LastVarFrame( nullptr )
 {
 	++s_Depth;
 }
@@ -99,6 +100,7 @@ bool BFFParser::Parse( BFFIterator & iter )
 		switch ( c )
 		{
 			case BFF_DECLARE_VAR_INTERNAL:
+			case BFF_DECLARE_VAR_PARENT:
 			{
 				if ( ParseNamedVariableDeclaration( iter ) == false )
 				{
@@ -160,10 +162,14 @@ bool BFFParser::Parse( BFFIterator & iter )
 
 // ParseNamedVariableName
 //------------------------------------------------------------------------------
-/*static*/ bool BFFParser::ParseVariableName( BFFIterator & iter, AString & name )
+/*static*/ bool BFFParser::ParseVariableName( BFFIterator & iter, AString & name, bool & parentScope )
 {
 	// skip over the declaration symbol
-	ASSERT( *iter == BFF_DECLARE_VAR_INTERNAL );
+	ASSERT( *iter == BFF_DECLARE_VAR_INTERNAL ||
+			*iter == BFF_DECLARE_VAR_PARENT );
+
+	parentScope = ( *iter == BFF_DECLARE_VAR_PARENT );
+
 	const BFFIterator varNameStart = iter; // include type token in var name
 	iter++;
 
@@ -196,6 +202,13 @@ bool BFFParser::Parse( BFFIterator & iter )
 	// store variable name
 	name.Assign( varNameStart.GetCurrent(), varNameEnd.GetCurrent() );
 
+	if ( parentScope )
+	{
+		// exchange '^' with '.'
+		ASSERT( BFF_DECLARE_VAR_PARENT == name[0] );
+		name[0] = BFF_DECLARE_VAR_INTERNAL;
+	}
+
 	return true;
 }
 
@@ -212,14 +225,17 @@ bool BFFParser::ParseUnnamedVariableConcatenation( BFFIterator & iter )
 		return false;
 	}
 
-	return ParseVariableDeclaration( iter, m_LastVarName );
+	return ParseVariableDeclaration( iter, m_LastVarName, m_LastVarFrame );
 }
 
 // ParseNamedVariableDeclaration
 //------------------------------------------------------------------------------
 bool BFFParser::ParseNamedVariableDeclaration( BFFIterator & iter )
 {
-	if ( ParseVariableName( iter, m_LastVarName ) == false )
+	const BFFIterator varNameStart( iter );
+
+	bool parentScope = false;
+	if ( ParseVariableName( iter, m_LastVarName, parentScope ) == false )
 	{
 		return false; // ParseVariableName() would have display an error
 	}
@@ -232,12 +248,23 @@ bool BFFParser::ParseNamedVariableDeclaration( BFFIterator & iter )
 		return false;
 	}
 
-	return ParseVariableDeclaration( iter, m_LastVarName );
+	// check if points to a previous declaration in a parent scope
+	m_LastVarFrame = ( parentScope )
+		? BFFStackFrame::GetParentDeclaration( m_LastVarName, nullptr )
+		: nullptr;
+
+	if ( parentScope && nullptr == m_LastVarFrame )
+    {
+    	Error::Error_1009_UnknownVariable( varNameStart, nullptr );
+    	return false;
+    }
+
+	return ParseVariableDeclaration( iter, m_LastVarName, m_LastVarFrame );
 }
 
 // ParseVariableDeclaration
 //------------------------------------------------------------------------------
-bool BFFParser::ParseVariableDeclaration( BFFIterator & iter, const AString & varName )
+bool BFFParser::ParseVariableDeclaration( BFFIterator & iter, const AString & varName, BFFStackFrame * frame )
 {
 	m_SeenAVariable = true;
 
@@ -316,7 +343,7 @@ bool BFFParser::ParseVariableDeclaration( BFFIterator & iter, const AString & va
 			Error::Error_1018_IntegerValueCouldNotBeParsed( startIntValue );
 			return false;
 		}
-		return StoreVariableInt( varName, i );
+		return StoreVariableInt( varName, i, frame );
 	}
 	else if ( ( *iter == 't' ) || ( *iter == 'f' ) )
 	{
@@ -335,7 +362,7 @@ bool BFFParser::ParseVariableDeclaration( BFFIterator & iter, const AString & va
 						Error::Error_1027_CannotConcatenate( operatorIter, varName, BFFVariable::VAR_ANY, BFFVariable::VAR_BOOL );
 						return false;
 					}
-					return StoreVariableBool( varName, true );
+					return StoreVariableBool( varName, true, frame );
 				}
 				else if ( value == "false" )
 				{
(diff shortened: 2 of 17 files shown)


kylawl:
This is a great feature, helps a lot.

ffulin:
There is a definite need for something like this - it greatly improves the flexibility of the bff syntax. Thanks as always to Poppolopoppo for the pull request!

The delay in integrating this change is caused by my uncertainty of the syntax. Expressing it as a "function" doesn't sit right with me for some reason, though I don't have a clear better idea.

I try to keep FASTBuild backwards compatible, so introducing something like this generally means it can't be changed, so whatever the final choice, it needs to be very done carefully. I will continue to stew on this!

poppolopoppo:
Maybe a new syntax for global variables :
````
.Local = 'local'
@Global = 'global'
{
    .Local = 'sublocal'
    @Local = 'still global'
}
Print( .Local ) ; prints "local"
Print( @Global ) ; prints "still global"
````
This can be easily implemented in BFFStackFrame with a global frame.

One advantage with that syntax is a much more simple code flow behind the scene in :
````
.Defines = { 'A', 'B', 'C' }
@CompilerOptions = ''
ForEach( .name in .Defines )
{
   @CompilerOptions + ' /D "$name$"'
}
````
Than :
````
.Defines = { 'A', 'B', 'C' }
.CompilerOptions = ''
ForEach( .name in .Defines )
{
   .CompilerOptions + ' /D "$name$"'
   Export( .CompilerOptions )
}
````

The first will only parse the concatenation for each loop and append to the variable declared in the global frame, while the second will create a new local variable `.CompilerOptions` copying each time the accumulated result, then searching recursively the scope containing the last `.CompilerOptions` definition and finally copying again the result in it. All of this for each loop. This is not a

On the other hand `Export()` function has the advantage of integrating more gracefully with scopes. It just happen that exporting a variable previously defined in the root frame will promote it as a "global variable" accessible from each scope, but it is not limited to that. On the contrary global variables are, indeed, global : all global variables used at least once will be piling in the global frame, without a way for the user to remove them. This implies to safely clean global variables before using them, always assuming than some other part might have been using them.

**Comment by kylawl** [comment]
I like the idea of a variable decorator like that. Why not give the @ the
same behavior as the existing Export function rather than making everything
global? You could even have 2 different behaviors if you wanted.

@GlobalVariable
^ParentScopeVariable

.Defines = { 'A', 'B', 'C' }
.CompilerOptions = ''
ForEach( .name in .Defines )
{
   ^CompilerOptions + ' /D "$name$"'
}

**Comment by poppolopoppo** [comment]
Yes, good idea :)

**Comment by coder-arn** [comment]
I second kylawl's idea of the @ for global scope and ^ for parent scope.  This combined with poppolopoppo's branch for variable expansion (https://github.com/poppolopoppo/fastbuild/tree/expand_vars) and I would be very happy!

**Comment by ffulin** [comment]
I like where this is going - lots of good idea.  We need to be careful to ensure we've captured all the edge cases and be sure it always feels intuitive. Some thoughts/questions on this:

 **@ as a global variable declarator**
This seems pretty reasonable, but what happens when you do this:
```C++
@Var = 1
{
    .Var = 2
}
```
Is that an error? Does it create a non-global shadow variable within the nested scope? 
I think it would probably need to be an error to not be confusing.

**^ as a scope modifier**
```C++
{
    ^Var = 1
}
```
Does this just push up one scope level?
What about if we're already at the global scope? (Error or just treat as if .Var = 1 ?)
Can you push up multiple scope levels?  Would that look like this?
```C++
{
    {
        ^^Var = 1
    }
}
```

**Impact on local scope:**
```C++
.MyVar = 'A'
{
   ^MyVar = 'B'
   Print( .MyVar ) // This prints "B" as you'd expect
}
```
But is this intuitive?
```C++
.MyVar = 'A'
{
   .MyVar + 'B'
   ^MyVar + 'C'
   Print( .MyVar ) // this would print "AB" (since we added 'C' to the parent scope)
}
Print( .MyVar ) // this would print "AC"
```
and then this:
```C++
.MyVar = 'A'
{
   ^MyVar + 'B'
   .MyVar + 'C'
   Print( .MyVar ) // this would print "ABC" (since we added 'B' to the parent scope *before* creating a new var)
}
Print( .MyVar ) // this would print "AB"
```
Does this behaviour feel consistent and logical?

**Mixed Metaphors**
@ is modifying the declaration
^ is modifying the operator
Is that ok?  I'm not sure - the inconsistency bothers me a little bit, but maybe it's fine.

poppolopoppo:
## @ as a global variable declarator
>````
@Var = 1
{
    .Var = 2
}
````

Globals should not be stored in *BFFStackFrame*, and therefore should be entirely orthogonal to local variables. Making this case not ambiguous as `@Var` and `.Var` won't shadow each other (very much like how [Ruby syntax](https://en.wikibooks.org/wiki/Ruby_Programming/Syntax/Variables_and_Constants) handles local, member, static, global variables and constants separately).

## ^ as a scope modifier
>````
{
    ^Var = 1
}
````
Does this just push up one scope level?
What about if we're already at the global scope? (Error or just treat as if .Var = 1 ?)

The current behavior is to fetch *the first previous definition in a parent scope*.
Those 2 examples should trigger the same error (like `Export()` at the moment). Not specifically because of the top level scope, but because a parent scope containing a previous declaration of `.Var` won't be found.

> Can you push up multiple scope levels? Would that look like this?

This does not match the current behavior. `^Var` may be confusing, as the variable will be searched recursively in parent scopes, not just in the first parent. This is why `^^Var` would not make any sense. Having to handle explicit scope levels would be a useless pain and make reusable configurations much harder to write.

Using the C dereferencing operator like `*Var` instead of `^Var` might be less misleading as it is syntactically closer to the actual behavior :
````
.var = "foo"
{
    {
        *var + "bar"
    }
}
````

## Impact on local scope:
>````
.MyVar = 'A'
{
   ^MyVar = 'B'
   Print( .MyVar ) // This prints "B" as you'd expect
}
````
But is this intuitive?

No you are right, this is confusing. But I guess it might get better again by using the `*Var` syntax coupled with a slight change of behavior : from "*the first previous definition in a parent scope*" to "*the first previous definition*".

If I apply these to your examples :
````
.MyVar = 'A'
{
   *MyVar = 'B'
   Print( .MyVar ) // This still prints "B" as you'd expect
}
````
````
.MyVar = 'A'
{
   .MyVar + 'B'
   *MyVar + 'C'
   Print( .MyVar ) // this would now print "ABC" (since we added 'C' to the local scope)
}
Print( .MyVar ) // this would now print "A"
````
````
.MyVar = 'A'
{
   *MyVar + 'B'
   .MyVar + 'C'
   Print( .MyVar ) // this would still print "ABC" (since we added 'B' to the 
             // parent scope *before* creating a new var)
}
Print( .MyVar ) // this would still print "AB"
````
Which, in my opinion, feel more intuitive and consistent.

## Mixed Metaphors
>@ is modifying the declaration
^ is modifying the operator

I am not sure to understand what you mean. For me both are about variable declaration scope, and everything behind remind the same. Actually it seems implemented like this, most of the code knowing only about *BFFVariables* but nothing about their corresponding *BFFStackFrame*.

I found one issue with the actual syntax though :
````
.var = 42 ; local
@var = 69 ; global
{
    *var = 13 ; what am I doing ?
}
````
Now this is very ambiguous, even if  `.var` and `@var` won't shadow each other. 
Maybe we should extend the syntax like this :
````
.var = 42 ; local
@var = 69 ; global
{
    *.var = 13
    *@var = 777
}
Print( .var ) ; prints "13"
Print( @var ) ; prints "777"
````

Sadly that won't fix this ambiguity :
````
.var = 42
@var = 69
Print( "$var$" ) ; again, what am I doing ?
````

These syntax changes seem to break the existing language in several places ultimately.
It might be better to completely remove the `@var` global variable concept and just keep `*var` syntax, therefore not adding a global scope (and all those ambiguities) and keeping global variables defined as *variables defined at top level scope*. This is probably more elegant anyway.

## ForEach() & Using()
Two weird additional examples :
````
ForEach( .A in .ArrayA )
{
    *A = 42
}
````
Not sure about the implications on that one (I think function headers have their own scope).
Maybe not weirder (albeit different) than :
````
ForEach( .A in .ArrayA )
{
    .A = 42
}
````


Also :
````
Using( *Struct )
````
Which may be considered valid, as the behavior is staying equivalent to `Using( .Struct )` I guess.

## Additional thoughts
BFF has no concept of constness, which is a very handy tool for both the user and the language. Environment variables should be imported as constants for instance, and the parser could also expose some builtin constants like `__FILE__` or `__DIRNAME__` (this would be so useful to deduce absolute paths).
This should also be exposed to the user with a specific syntax, note sure about which one though.
````
const .local = 42 ; const is arbitrary here, waiting for a better idea
const @global = 42
const *global = 42 ; error this is an invalid syntax
````
Or if we decide to remove the global scope :
````
.variable = 42
.variable = 69
@constant = 42
@constant = 69 ; boom, already defined constant
.constant = 69 ; boom, trying to assign to a constant
{
    .variable = 42
    .constant = 42 ; boom again, last definition is locked by '@' constant prefix
}
````
---
> This combined with poppolopoppo's branch for variable expansion (https://github.com/poppolopoppo/fastbuild/tree/expand_vars) and I would be very happy!

@coder-arn: I'll add a pull request for [expand_vars](https://github.com/poppolopoppo/fastbuild/commit/596e4d9c64a2ef333db13b7d6da4ce5777a352ab), which emulates associative containers : 
````
.key = "abc"
."values_$key$" = [ .name = 'john' .age = 42 ]
{
    Using( ."values_$key$" )
    Print( "$name$, $age$" ) ; output "john, 42"
}
````
This has been in the oven for a while, but I was hospitalized recently and had to take a break. I'll get back to work soon though, many things I have to finish (and add :wink:)

**Comment by kylawl** [comment]
I'm perfectly fine not having the @ globals. Currently with Export(), we're getting the behavior we need anyway.

**Comment by ffulin** [comment]
Thinking about this some more, it seems @ is really not that necessary, since declaring something once globally with the normal syntax, and then always manipulating it with ^ would give you global behaviour.

With that in mind, if we only support ^, and make it reference the variable at closest parent scope, I think we have something fairly simple and understandable.  Thoughts?

BTW - one thing we'll need to make sure behaves sensibly (either looping through the original, or by generating an error) is this:
```C++
ForEach( .Thing in .Things )
{
    ^Things += 'Infinte loop'
}
```

poppolopoppo:
Ok I'll remove the function and add the syntax `^Var` which will point to the last definition 
**starting from the parent scope**.

I'll also handle the case you described by adding a new frozen state to BFFVariables. A new error will be thrown if the user tries to write to a frozen variable. `ForEach()` will freeze the variables for the duration of the loop. This should be reusable for other functions/features.

This will be still valid :
````csharp
ForEach( .Thing in .Things )
{
    .Things += 'Infinte loop'
}
````

Everybody ok with that ?

poppolopoppo:
You can see the results at https://github.com/poppolopoppo/fastbuild/commits/parent_scope.
I let all of you review the code, and if you are ok I'll force push on this pull request.

kylawl:
The change sounds good, thanks for hamming all this awesome work in!

ffulin:
I looked at the branch - at a glance it seems good!  I've added a few comments on some minor things I spotted.

Given that this changes variable parsing in a significant way, we're going to need to extend the unit tests. Is that something you're comfortable in trying?

poppolopoppo:
Thanks for your comments, I am fixing the issues.
I'll also add more unit tests in TestBFFParsing ([already added some](https://github.com/poppolopoppo/fastbuild/commit/4e262220fc7e983e634f9e4a8151c110c90f4e5b) for the change of behavior of the stack).

I actually use your tests to validate my commits, but I add to fix some issues before getting them to work at home. I have a [branch](https://github.com/poppolopoppo/fastbuild/commits/fix_unit_tests) with those fixes, and [another](https://github.com/poppolopoppo/fastbuild/commits/configure_externals) with a script to configure externals using mklinks (windows only). I have to admit that I can't run all tests though (distributed builds, osx, consoles, ...) despite having each version of visual studio from 11.0 to 15.0 :sweat_smile:

poppolopoppo:
I updated the branch [parent_scope](https://github.com/poppolopoppo/fastbuild/commits/parent_scope) with the fixs, documentation for error 1060 and new unit tests.
Let me know If it corresponds to what you had in mind.

ffulin:
I took a look at the various commits to the parent_scope branch and they look good.

If you can update this pull request I think we can start getting it integrated.

poppolopoppo:
I updated the pull request.
Let me know if you need anything.

YannickTriqueneaux:
I'm just very happy to know this feature is coming.

As I said to Franta, this ParentScopVariable will unlock lot of possibilities !

Thank you guys for that !

ffulin:
This has been integrated into the Perforce mainline and will be in the next release v0.87. I made some tiny edits (docs and test comments) which I've checked into the git dev branch.

Thanks again for this change!

poppolopoppo:
You're welcome, I am happy that I could contribute to this project :)
Thanks you too for your hard work Franta !

PR #14, opened 2015-06-25
Support for system environment variables expansion

poppolopoppo:
Uses the std::getenv() function, so the behavior is the same on all platform.
A 1009 error will be raised if the variable is not found in system environment.

Allows :
```
.VSBasePath	= '%VS120COMNTOOLS%..\..\'
```

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
@@ -598,6 +598,10 @@ bool BFFParser::ParsePreprocessorDirective( BFFIterator & iter )
 	{
 		return ParseEndIfDirective( directiveStartIter );
 	}
+	else if ( directive == "import" )
+	{
+		return ParseImportDirective( directiveStart, iter );
+	}
 
 	// unknown
 	Error::Error_1030_UnknownDirective( directiveStartIter, directive );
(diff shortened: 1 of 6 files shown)


poppolopoppo:
I forgot to mention one important change necessary for this patch.
**Quoting is now necessary for build-time substitutions** :
````
Library( "mylib" )
{
  .CompilerOptions   = '%1 /Fo%2 /nologo /c' ; substitutions detailed below
}
````
Must be fixed like that :
````
Library( "mylib" )
{
  .CompilerOptions   = '^%1 /Fo^%2 /nologo /c' ; substitutions detailed below
}
````

ffulin:
Hello - thanks for the pull request. This is a long-time requested feature :)

Unfortunately, I think there are a couple of problems that prevent the integration of this implementation:
a) config isn't re-parsed if environment variables change
b) syntax is not backwards compatible  We can't ask all the existing users to update all their bff files to escape % signs. Only in extremely rare circumstances can backwards compatibility of existing configs be broken.

The way I had anticipated I would implement this is as follows:
a) add a new directive, #import ENV_VAR_NAME - for simplicity, I imagined that this would declare a variable .ENV_VAR_NAME (in the future we could extend the syntax to "#import X as .Y" to allow the user to optionally specify the variable).  This avoids the breaking change.
b) Track the list of imported variables, and write the name of the var and the hash of the value in the fdb.  I already do this for the special case LIB env var, so it would involve re-jigging that to generaicly handle a list of imported vars (plus the special cases):  Search for:
				FLOG_WARN( "'%s' Environment variable has changed - BFF will be re-parsed\n", "LIB" );

Additionally, is it possible to use the Env::GetEnvVariable function?

Sorry!

poppolopoppo:
Hi,

I see that you have thought well about that one ;)

I did not thought of problem a), mostly because my current workaround (bff generated by a script) has the same problem. And yeah, maybe breaking syntax is not a good idea and another separator could be used.

Your solution b) seems to be the way to go.
I have yet one last idea (but you probably already thought of it) : if environment variables were not expanded in the parser but later like in the loading of each node, would it work then ?

Sorry for Env::GetEnvVariable() I did not see (or, hum, search for) it :p

No problem though to reject this pull request, I am still glad that this was already requested and that you are thinking of it :)

poppolopoppo:
Hi,

So I finally chose to implement it the way you suggested it.
* There is a new directive `#import ENV_VAR_NAME`. 
An error 1009 will be raised if `ENV_VAR_NAME` was not found in system environment.
* Each imported variables are registered with `FBuild::ImportEnvironmentVar()`. The hash of each variable value is  stored in the singleton.
* When saving `NodeGraph` the list of imported variable names and their hash values are saved.
* When loading `NodeGraph` the list is read and each variable imported again in FBuild. 
If the hash value after import differs from the one saved then a reparsing will occur. If the variable name was not found in system environment a reparsing will also occur.
* This does not impact the behavior of `LibEnvVar`, which can be dependent of `EnvironmentString` unlike explicitly imported variables. The code was not generalized for this reason.
* I use `Env::GetEnvVariable()` :)

NODE_GRAPH_CURRENT_VERSION was incremented to support the FDB file format change.

ffulin:
Very nice!

ffulin:
As well as integrating this into dev, I've integrated it into the Perforce mainline and added documentation.

This change wil be in the next main release (v0.84)

PR #16, opened 2015-06-27
New RemoveDir() function

poppolopoppo:
Remove files in a directory.
Same usage than CopyDir :
````
RemoveDir( 'name' ) // Name (required)
{
  // Basic options
  .Paths              			// Directories to search

  // Filtering options
  .PathsPattern       			// (optional) Wilcard pattern to filter files (default: "*")
  .PathsRecurse       		 // (optional) Recurse into sub-directories? (default: true)
  .ExcludePaths          // (optional) Directories to ignore when recursing

  // Additional options
  .PreBuildDependencies  // (optional) Force targets to be built before this RemoveDir
}
````
Usefull to handle explicit cleaning of a target.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
@@ -18,6 +18,7 @@
 #include "FunctionLibrary.h"
 #include "FunctionObjectList.h"
 #include "FunctionPrint.h"
+#include "FunctionRemoveDir.h"
 #include "FunctionSettings.h"
 #include "FunctionSLN.h"
 #include "FunctionTest.h"
@@ -115,6 +116,7 @@ Function::~Function()
 	FNEW( FunctionForEach );
 	FNEW( FunctionLibrary );
 	FNEW( FunctionPrint );
+	FNEW( FunctionRemoveDir );
 	FNEW( FunctionSettings );
 	FNEW( FunctionSLN );
 	FNEW( FunctionTest );
@@ -388,7 +390,7 @@ bool Function::GetInt( const BFFIterator & iter, int32_t & var, const char * nam
 // GetNodeList
 //------------------------------------------------------------------------------
 bool Function::GetNodeList( const BFFIterator & iter, const char * name, Dependencies & nodes, bool required,
-							bool allowCopyDirNodes, bool allowUnityNodes ) const
+							bool allowCopyDirNodes, bool allowUnityNodes, bool allowRemoveDirNodes ) const
 {
 	ASSERT( name );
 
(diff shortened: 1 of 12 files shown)

### CODE REVIEW

**Comment by ffulin** [comment]
This is interesting, but there are some subtleties to this that I think are problematic, at least with the current implementation:

This snippet of code:
```C++
+            // source file (as a node)
+            Node * srcFileNode = ng.FindNode( srcFile );
+            if ( srcFileNode == nullptr )
+            {
+                srcFileNode = ng.CreateFileNode( srcFile );
+            }
```

Will auto-create a FileNode for a file to be deleted, if nothing has defined how to build that, but ObjectFiles are defined at build time. This means depending on the order of operations, you might incorrectly define a plain FileNode for something that should be an ObjectNode.

For example, having compiled once before, and then doing this order:
* Compile and object xyz.obj is created as ObjectNode
* Deleted fdb manually or update with a version change
* Run Remove-X target, which defined as xyz.obj as FileNode
* Compile, build fails and the only way to recover is to delete the .fdb

Additionally, I think as implemented, running the remove target will cause the target to first be built and then deleted, because the removal node depends on the compiled object. 

We might be able to fix these problems by not explicitly mapping these dependencies, or by making them "weak". This will require some careful thought.

No matter what, we'll definitely need unit tests to ensure we can verify the behaviour of these edge cases.

poppolopoppo:
Wow, I really did not see that coming. Good eye !

Like you said not handling deleted files as dependencies would fix those problems.
Then RemoveNode could be removed and RemoveDirNode would systemically perform the delete in DoBuild().

poppolopoppo:
Hi,

I updated the patch with your remarks.
* RemoveNode does not exist anymore.
* RemoveDirNode has no more dynamic dependencies to FileNodes.
* RemoveDirNode::DoBuild() will delete the files listed by DirectoryListNodes in static dependencies.
* RemoveDirNode has always a 0 timestamp, triggering DoBuild() every time.
* Node::GetNodeList() has a new flag allowRemoveDirNodes, mandatory to support RemoveDirNodes as .Targets in FunctionAlias.

https://github.com/fastbuild/fastbuild/pull/16.diff

ffulin:
I've integrated this into the Perforce mainline. This will be part of the next release (v0.88). 

I've made some additions/changes during the integration:
 - added documentation 
 - updated the Notepad++ and Visual Studio integration files
 - modified option names

I changed to naming of the options to avoid likely conflicts with global variables and for slightly improved consistency with existing conventions. The options are now:
```C++
  // Basic options
  .RemovePaths              // Directories to delete contents of
  .RemovePathsRecurse       // (optional) Recurse into sub-directories? (default: true)

  // Filtering options
  .RemovePatterns           // (optional) Wildcards of contents to delete (default: *)
  .RemoveExcludePaths       // (optional) Directories to ignore if recursing
```

I've integrated these updates into dev.

poppolopoppo:
Good point on the naming, thanks for you hard work !

ffulin:
My pleasure :) Sorry it took so long to get this one integrated!

PR #17, opened 2015-06-27
AString::Match() supports multiple globbing patterns

poppolopoppo:
Mutliple globbing patterns can de defined in one string.
Patterns must be separated by a comma (ex: "*.h,*.cpp").
AString::MatchI() also uses the same behavior.

The algorithm is base on the function szWildMatch7() from Alessandro Cantatore,
because it is said to be up to 60% faster than previous one (szWildMatch5).
http://xoomer.virgilio.it/acantato/dev/wildcard/wildmatch.html

This modification impacts every function using globbing patterns.

Ex:
````
Unity( 'unity' )
{
    .UnityInputPattern = "*.h,*.cpp"
}
````

### DIFF
--- a/Code/Core/Strings/AString.cpp
+++ b/Code/Core/Strings/AString.cpp
@@ -19,6 +19,93 @@
 /*static*/ const char * const AString::s_EmptyString( "" );
 /*static*/ const AString AString::s_EmptyAString;
 
+// CharEqual<>
+//------------------------------------------------------------------------------
+template < bool _CaseSentitive >
+struct CharEqual
+{
+	bool operator ()( char lhs, char rhs ) const
+	{
+		return lhs == rhs;
+	}
+};
+
+template <>
+struct CharEqual< false >
+{
+	bool operator ()( char lhs, char rhs ) const
+	{
+		lhs = ( ( lhs >= 'A' ) && ( lhs <= 'Z' ) ) ? 'a' + ( lhs - 'A' ) : lhs;
+		rhs = ( ( rhs >= 'A' ) && ( rhs <= 'Z' ) ) ? 'a' + ( rhs - 'A' ) : rhs;
+		return lhs == rhs;
+	}
+};
+
+// WildMatch
+// NOTE: This code is based on that in the C/C++ Users Journal (Mike Cornelison)
+//------------------------------------------------------------------------------
+template < bool _CaseSentitive >
+static bool WildMatch( const char *pat, const char *end, const char *str )
+{
+	const CharEqual< _CaseSentitive > equalto;
+
+	bool star;
+
+new_segment:
+	star = false;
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by ffulin** [comment]
This is cool.  A couple of questions:
* Did you profile the new core algorithm versus the old one? Unless it's proven to be faster, I'd be inclined to keep the existing tested algorithm.
* How do you feel about using | as a separator instead of ,.  Since | this is not a valid filename character (on Windows at least). This feels a little safer, since this change will impact right down to directory listing behaviour.

**Comment by poppolopoppo** [comment]
* I did not profile, I can rollback to the previous algorithm if you want.
* I thought about |, but I saw some code using it (mostly automatic name formatting). But if there is no problem then I can change for it.

**Comment by poppolopoppo** [comment]
Hi,

I updated the patch to use '|' instead of ',' and reverted to the previous algorithm.

Ex:
````
Unity( 'unity' )
{
    .UnityInputPattern = "*.h|*.cpp"
}
````

**Comment by ffulin** [comment]
Nice.

You're right about | being used in a special way for DirectoryListNode name generation (which I forgot about :)). I've looked at the code in question and it should be fine - it's just used to ensure a unique name for the DirectoryListNode (we don't parse the strings back out). If it does turn out to be a problem , I can change the name generation serparator, so we're still cool.

**Comment by ffulin** [comment]
In the end I decided that for consistency, a list of patterns should use the ArrayOfStrings syntax. I've made the necessary code changes so all places that take a pattern can now take a single String or an ArrayOfStrings. As a small bonus, this also avoids some small overhead in the string matching as we don't need to find the boundaries of the patterns for every match.

I've made this change in the Perforce mainline and it will be in the next release (v0.87).

So the syntax to specify multiple patterns is:
```C++
.CompilerInputPattern = { "*.cpp", "*.cc" }
```
as well as still being able to specify a single one as before:
```C++
.CompilerInputPattern = "*.cpp"
```

With that in mind, I won't integrate this change. I'll wait for your confirmation before closing the pull request.

poppolopoppo:
You are right this is better :)
I close the pull request.

poppolopoppo:
Apparently this was not implemented for the property UnityInputPattern of Unity function :/
````javascript
Unity( '$ProjectName$-Unity' )
{
    .UnityInputPattern = { '*.cpp', '*.c' }
}
````
````
Unity() - Property '.UnityInputPattern' must be of type <String> (found <ArrayOfStrings>).
````

poppolopoppo:
Oh my bad, it seams that you didn't already merge that from perforce :p

ffulin:
Yeah, this change (and several others) have not been brought into GitHub from the Perforce mainline. That'll happen when 0.87 is released, which should be very soon.

I've double-checked that UnityInputPattern is correctly updated to support multiple wildcards in the incoming change.

PR #18, opened 2015-06-27
Unity timestamp

poppolopoppo:
If any included file was written later than current unity, the last write time of the unity file will be updated.
This will trigger the compiler, avoiding false-positive clean state.

Sadly this cannot handle included headers, since UnityNode only knows about project source files.
But explicitly including project headers in unity files might fix that, and still prevent a costly compiler call.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/UnityNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/UnityNode.cpp
@@ -22,6 +22,17 @@
 #include "Core/Process/Process.h"
 #include "Core/Strings/AStackString.h"
 
+// CONSTRUCTOR
+//------------------------------------------------------------------------------
+struct FileAndOriginCompare
+{
+    bool operator ()(   const UnityNode::FileAndOrigin& lhs,
+                        const UnityNode::FileAndOrigin& rhs ) const
+    {
+        return lhs.GetName() < rhs.GetName();
+    }
+};
+
 // Reflection
 //------------------------------------------------------------------------------
 REFLECT_BEGIN( UnityNode, Node, MetaNone() )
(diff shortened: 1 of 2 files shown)


ffulin:
I'm having a little trouble understanding the problem this change addresses.

The Unity cpp files are a bit of a special case.  The current behavior is that they are generated in-memory every time, but only written to disk when they change.  This updates the last write time automatically.

Can you please describe the case that wasn't working for you in more detail so I can understand the problem?  Was something not compiling when it should?  Or re-compiling when it shouldn't?

PS - Thanks for the bonus file ordering fix :)

poppolopoppo:
My problem is that after compiling once the generated unity files FastBuild won't summon the compiler on these if I do not change the list of files used as inputs. Then only modifying source files won't trigger a compilation.

The code I added won't change the part where you compare in memory generated unity with the version written on the disk. But if this test pass -and only if it pass- and  if any included file has been written since last commit to the disk, the patch updates the last write time of the generated file.  Updating the last write time will force DetermineNeedToBuild() to see the change and not false-skip the compilation.

PS- No problem, it was an easy TODO task :)

ffulin:
Normally, the ObjectNode created to compile each generated Unity*.cpp has dependencies on all the files used (recursively, including headers) to compiler that cpp file.

There's only three reasons I can think of that this wouldn't be the case:
**1) An unsupported compiler is being used.** FASTBuild can only monitor dependencies correctly for the supported compilers listed here: http://fastbuild.org/docs/status.html - If your compiler is not listed here, but works like GCC/Clang/MSVC, support can be added.
**2) A supported compiler is not correctly detected.** FASTBuild uses the executable name to detect the compiler.  If your compiler is supported, maybe the exe name is not being recognized correctly.
**3) The compiler output is not being correctly parsed by FASTBuild.** FASTBuild uses the /E, -E or /showIncludes options in various ways depending on the compilation mode and compiler. In rare cases in the past there have been some compiler options that modified the output in ways that broke FASTBuild's parsing (for example Clang for Windows).  Maybe this is happening here?

Let me know if any of these could be the issue.

poppolopoppo:
I better understand now how UnityNode works, and how commit this is not a fix :p
I'll investigate more the problem with your recommendations, thanks for that.
I close this pull request.

ffulin:
Please be sure to let me know if it turns out to be a problem with output parsing or compatible compiler detection!

PR #19, opened 2015-06-27
Support for user defined macros

poppolopoppo:
BFF support for #define and #undef directives.

Ex:
````
#define USE_FEATURE
#if USE_FEATURE
; active
#endif
#undef USE_FEATURE
#if USE_FEATURE
; inactive
#endif
````

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFMacros.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFMacros.cpp
@@ -1,0 +1,95 @@
+// BFFMacros
+//------------------------------------------------------------------------------
+
+// Includes
+//------------------------------------------------------------------------------
+#include "Tools/FBuild/FBuildCore/PrecompiledHeader.h"
+
+#include "BFFMacros.h"
+#include "Tools/FBuild/FBuildCore/FLog.h"
+
+// Core
+#include "Core/Env/Assert.h"
+#include "Core/Strings/AString.h"
+
+// CONSTRUCTOR
+//------------------------------------------------------------------------------
+BFFMacros::BFFMacros()
+: m_Tokens(8, true)
+{
+}
+
+// DESTRUCTOR
+//------------------------------------------------------------------------------
+BFFMacros::~BFFMacros()
+{
+}
+
+// IsDefined
+//------------------------------------------------------------------------------
+bool BFFMacros::IsDefined(const AString& token) const
+{
+    // user defined tokens :
+    AString *const defined = m_Tokens.Find( token );
+    if ( defined != nullptr )
+    {
+        return true;
+    }
+
+    // fallbacking to predefined tokens :
+    if ( token == "__WINDOWS__" )
+    {
+        #if defined( __WINDOWS__ )
+        return true;
+        #endif
+    }
+    if ( token == "__LINUX__" )
+    {
+        #if defined( __LINUX__ )
+        return true;
+        #endif
+    }
+    if ( token == "__OSX__" )
+    {
+        #if defined( __OSX__ )
+        return true;
+        #endif
+    }
+
+    return false;
+}
+
+// Define
+//------------------------------------------------------------------------------
+bool BFFMacros::Define( const AString& token )
+{
+    if ( IsDefined( token ) )
+    {
+        // trying to overwrite an existing token :
+        return false;
+    }
+    else
+    {
+        m_Tokens.Append( token );
+        return true;
+    }
+}
+
+// Undefine
+//------------------------------------------------------------------------------
+bool BFFMacros::Undefine( const AString& token )
+{
+    AString *const defined = m_Tokens.Find( token );
+    if ( defined == nullptr )
+    {
+        // trying to remove an unexisting or predefined token :
+        return false;
+    }
+    else
+    {
+        m_Tokens.Erase( defined );
+        return true;
+    }
+}
+
+//------------------------------------------------------------------------------
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFMacros.h
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFMacros.h
@@ -1,0 +1,36 @@
+// BFFMacros - manages defined macros
+//------------------------------------------------------------------------------
+#pragma once
+#ifndef FBUILD_MACROS_H
+#define FBUILD_MACROS_H
+
+// Includes
+//------------------------------------------------------------------------------
+#include "Core/Containers/Array.h"
+#include "Core/Containers/Singleton.h"
+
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by ffulin** [comment]
This is a really nice change, thankyou!

I have a few minor comments:
* Invalid first line in BFFMacros.cpp - The first line is missing a /.  This will compile with VS because it magically ignores lines before the precompiled header include.  This will fail with GCC/Clang.
* Could you please use the Singleton template in Core, instead of rolling your own.  Not a big deal, since it's much the same as what you have, but it helps keep the code style unified.
* Avoid whitespace changes - There are many white-space only changes in this CL. Not a huge deal, but it makes inspecting the change more difficult (and potentially merging it).

**Comment by poppolopoppo** [comment]
No problem!
* Wow I did not know this behavior of VS, this is nasty. I'll fix that.
* My code was based on BFFStackFrame so I did not look for a generic singleton, will do though.
* Yeah sorry about that, my editor trimmed trailing whitespaces automatically on save. I disabled that.

**Comment by poppolopoppo** [comment]
Hi,

I updated the patch.
* Fixed invalid first line in BFFMacros.cpp.
* BFFMacros is a Singleton< BFFMacros >.
* Remove white space only changes.

https://github.com/fastbuild/fastbuild/pull/19.diff

**Comment by poppolopoppo** [comment]
Hi,

I updated the pull request with le following changes :
* Fixed parsing of macro names in `#if`, `#define` and `#undef` ([see commit for changes](https://github.com/poppolopoppo/fastbuild/commit/b342925b8d92c590809e5a85459e25becde67d97)) ;
* Added info log when defining or undefining a macro ;
* Added an alternative name `#ifdef` in addition to `#if` directive ... ;
* ... Because I wanted to add support for `#ifndef`, which seemed inconcistent without `#ifdef`.

**Comment by ffulin** [comment]
I've been giving some thought to whether I want #if and #ifdef to be subtly different as they are in C/C++.

The way I implemented #if originally, is like in C/C++, where it's performing a logical operation - i.e. __WINDOWS__, __LINUX__ and __OSX__ are always defined, and are defined to 0 or 1.

The idea i had was that we might later like to be able to do this:
```C++
#define THING 0
#if ( THING == 1 ) // false
````
which would be distinctly different from:
```C++
#define THING 0
#ifdef THING // true
```
and also diferent to:
```C++
#define THING 0
#if THING // false
```

With the implementation you have so far, #if has subtly changed so that it's essentially #ifdef.
I think we need to think very carefully about whether we want #if to work as before, and perhaps have #ifdef work like you have now.

It's important because once a version is out with your change integrated, it'll be impossible to change without breaking backwards compatibility.

What do you think?

poppolopoppo:
Yes, #ifdef is now completely equivalent to #if. 

I can add the support for :
````c
#define THING VALUE
````
And implement a behavior closer to C/C++ #if/#ifdef directives, like you described.

This raises some questions though :
* Is VALUE optional ?`
* Is VALUE restricted to integral value ? 
* If no, can VALUE be another macro ?

Personally I prefer the behavior of the C/C++ #if, because #if UNKNOWN will fail, while #ifdef UNKNOWN does not tell me if I forgot to define UNKNOWN or if this is intended. 

So I think you have a good point :)

**Comment by ffulin** [comment]
I kind of hate how C++ causes errors because of the subtlety of differences between #if and #endif. Would be good to avoid getting into that situation.

I am thinking that restricting defines to either exist or not, only only allowing a test whether they exist or not might be the way to go. Maybe we could do this:
```C++
#if THING
```
```C++
#if not THING
```
and not have #ifdef at all?  That would keep it simple. What do you think?

(If we add the ability to assign values to defines later, we could then make it an error to just use #if THING with a define that had a value)

ffulin:
Could just make it this to make parsing easier:
```C++
#ifnot THING
```

ffulin:
I re-read what I wrote, and realized maybe it's not clear what I meant :)

I propose that for now, we just have #if and #ifnot which both gracefully handle the define not existing.
That means we can do:
```C++
#if __WINDOWS__
   #define DOTHING
#endif
...
#if DOTHING
   ....
#endif
```
and this would work as expected on Linux and OS X where DOTHING would not be defined.

This also means the way \__WINDOWS__/\__LINUX__/\__OSX__ is handled when checking for defines can be removed as a special case, since it now behaves the same. We just auto-define the right one once at startup depending on the platform.

**#ifnot** naturally works in the reverse of #if.

This way we:
* Avoid the confusion of two different keywords
* Still allow for value assignment in the future if that ever becomes useful

So some hypothetical future **(which we totally ignore for now)** where we allow defines to have values could be managed like this
```C++
#define THING_LEVEL 7
#if THING_LEVEL // <-- Error - not allowed to check for existence only
```
we would require the user to be more explicit in that case:
```C++
#if ( THINGLEVEL == 7 ) // If define has an associated value, it must be explicitly compared
...
#if defined( THINGLEVEL )  // If you *really* want to check just for existence you have to do thi
```

poppolopoppo:
Ok, guess I can remove #ifdef, #ifndef & #undef. 
And also remove platform defines special handling (which was done that way to forbid #undef on them).

But #ifnot feels really weird, better not doing anything than doing that for now.

ffulin:
No problem with #undef btw.

ffulin:
Do you have a suggestion for an alternate #ifnot syntax?  I just wanted to avoid #if and #ifdef doing the same thing and I agree with what you said before that #if and #ifndef don't match.

poppolopoppo:
In order to release the feature I'll keep only `#if` and `#undef`.
Builtin macros should be stored apart to disable `#undef` on them.

I'll respect this behavior :
````c
#if __WINDOWS__
   #define DOTHING
#endif
...
#if DOTHING
   ....
#endif
````

Instead of `#ifnot`I propose to parse :
````c
#if ! DOTHING
````
This is less surprising for the new user and will be naturally integrated in a true boolean grammar.

Thoughts ?

ffulin:
That sounds good.  Let me know when you've updated the pull request.

poppolopoppo:
I removed the last commit which added support #ifndef and added support for '!'.

ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.87)

Additionally, I've made some further modifications which I've [integrated](https://github.com/fastbuild/fastbuild/commit/8dcafb6320b357846ea994a26763c2ea82a961ab) from the Perforce mainline into "dev":
 - Added documentation for new keywords
 - Added documentation for new errors (and some existing missing ones)
 - Added an explicit error when attempting to #undef a built-in token (rather than reporting that it doesn't exist)
 - Added unit tests for the new functionality

poppolopoppo:
Thanks for your work :)

ffulin:
My pleasure :)

PR #20, opened 2015-06-27
Support for optional variable assignment

poppolopoppo:
New assignment operator '?'.
Variable won't be assigned if it already exists.
Allows optional parameter handling.

Ex:
````
.ProjectName = 'MyProject1'
{
    .BinaryName ? "$ProjectName$.exe"
    Print( .BinaryPrefix ) ; prints "MyProject1.exe"
}
````
````
.ProjectName = 'MyProject2'
.BinaryName  = 'SpecialName.exe'
{
    .BinaryName ? "$ProjectName$.exe"
    Print( .BinaryPrefix ) ; prints SpecialName.exe"
}
````

### CODE REVIEW

**Comment by ffulin** [comment]
Hello :)

The implementation looks solid, and your example makes sense, but I'm trying to understand what a real-world use case for this would be.

If we add the ? operator, and make it behave this way, we can essentially never change it, so we need to be certain this works exactly the way that gives most value.

Could you please provide a more detailed example that could help me understand the use-case?

**Comment by poppolopoppo** [comment]
Hi :)

I am using FastBuild like this most of the time :
````
; MySolution.bff
#include "BFF/CompileConfigs.bff"
.ProjectIncludes = ''
.ProjectLibraries = ''
#include "Projects/MyProject.bff"
````
````
; MyProject.bff
#once
{
    .ProjectName = 'MyProject'
    .ProjectPath = 'Projects/MyProject'
    .ProjectLibraries +  'SomeExternalLib.lib'
#   include "BFF/Template/Executable.bff"
}
````
````
; BFF/Template/Executable.bff
; no #once
{
    Unity( '$ProjectName$-Unity' )
    {
        .UnityInputPath = .ProjectPath
    }
    ForEach( .CompileConfig in .AllCompileConfigs )
    {
        Using( .CompileConfig )
        Executable( '$ProjectName$-Lib-$Config$-$Platform$' )
        {
            .CompilerOptions + .ProjectIncludes + .ProjectLibraries
            .LinkerOutput = "Output/$Config$/$Platform$/$ProjectName$.exe"
        }
     }
     ; ....
}
````

Which is already nice, because in MyProject.bff I did not set .ProjectIncludes, even if Library.bff need it. This is because I declared at top level the variable with a default value. This is how I handle optional parameters of my "templates".

One downside though is that I am obliged to declare every default parameters in some parent scope. And that parent scope has no knowledge about my already defined "parameters" like .ProjectName.

Adding optional parameters allows to rewrite all like this :
````
; MySolution.bff
#include "BFF/CompileConfigs.bff"
#include "Projects/MyProject.bff"
````
````
; MyProject.bff
#once
{
    .ProjectName = 'MyProject'
    .ProjectBinaryName = 'MyGameName'
    .ProjectLibraries =  'SomeExternalLib.lib'
    .ProjectCompileConfigs = .OnlyDebugCompileConfigs
#   include "BFF/Template/Executable.bff"
}
````
````
; BFF/Template/Executable.bff
; no #once
{
    .ProjectLibraries ? ''

    ; can't express those without ? :
    .ProjectBinaryName ? .ProjectName
    .ProjectPath ? 'Projects/$ProjectName$'
    .ProjectCompileConfigs ? .AllCompileConfigs
    .ProjectIncludes ? '/I"$ProjectPath$/Headers"'

    Unity( '$ProjectName$-Unity' )
    {
        .UnityInputPath = .ProjectPath
    }
    ForEach( .CompileConfig in .ProjectCompileConfigs )
    {
        Using( .CompileConfig )
        Executable( '$ProjectName$-Lib-$Config$-$Platform$' )
        {
            .CompilerOptions + .ProjectIncludes + .ProjectLibraries
            .LinkerOutput = "Output/$Config$/$Platform$/$ProjectBinaryName$.exe"
        }
     }
     ; ....
}
````
With optional assignment I can keep all default values in Executable.bff, where I can deduce more parameters since I have access to callee scope. The reader also knows that he can override those values if needed.

**Comment by Agga** [comment]
I tried this proposal and it works good. Here are a couple of examples i tried.


    // case A
    .arrayA ? {}
    .arrayA + {
    	'data0',
    	'data1'
    }
    
    // case B
    .arrayB ? {} + {
    	'data0',
    	'data1'
    }
    
    // case C
    .arrayC ? .arrayB

Combined with another recent feature we can do interesting build definitions.

    // ensure existence
    ."$Platform$_Includes" ? {} 
    ."$BuildType$_Includes" ? {} 
    ."$Project$_Includes" ? {}
    
    // concat with possibly defined arrays
    .All_Includes ? {}
    	+ ."$Platform$_Includes"
    	+ ."$BuildType$_Includes"
    	+ ."$Project$_Includes"

**Comment by Manuzor** [comment]
I like the idea behind this PR and stumbled upon this myself several times already, but it feels like this is just a use-case of a more general problem.

I think we can get more value out of this by providing proper branching. In issue #82 there's a discussion about something similar to this with regards to importing environment variables. But a `#if`/`#else`/`#endif` construct was already part of the discussion there.

If we had general branching, we could add some syntactic sugar at the parsing stage that would translate `.Var ? Value` to something like this:

```
#if !defined(.Var)
  .Var = Value
#endif
```

This would eliminate the need for supporting optionals internally. I'm not too familiar with the internals of FASTBuild, but it feels like this would be the more general solution to this problem.

PR #26, opened 2015-07-05
Adding multi-input dependencies for Exec command

inbilla:
Previously only one file could be defined that would invalidate the Exec command's output by timestamps. With this change, a list of files can be defined that will invalidate the Exec command's outputs.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ExecNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ExecNode.cpp
@@ -25,19 +25,22 @@ ExecNode::ExecNode( const AString & dstFileName,
 						const AString & arguments,
 						const AString & workingDir,
 						int32_t expectedReturnCode,
-						const Dependencies & preBuildDependencies )
+						const Dependencies & preBuildDependencies,
+						const Dependencies & additionalDependencies )
 : FileNode( dstFileName, Node::FLAG_NONE )
 , m_SourceFile( sourceFile )
 , m_Executable( executable )
 , m_Arguments( arguments )
 , m_WorkingDir( workingDir )
 , m_ExpectedReturnCode( expectedReturnCode )
+, m_AdditionalDependencies( additionalDependencies )
 {
 	ASSERT( sourceFile );
 	ASSERT( executable );
 	m_StaticDependencies.SetCapacity( 2 );
 	m_StaticDependencies.Append( Dependency( sourceFile ) );
 	m_StaticDependencies.Append( Dependency( executable ) );
+	m_StaticDependencies.Append( additionalDependencies );
 	m_Type = EXEC_NODE;
 
 	m_PreBuildDependencies = preBuildDependencies;
@@ -113,6 +116,7 @@ ExecNode::~ExecNode()
 	NODE_LOAD( AStackString<>,	workingDir );
 	NODE_LOAD( int32_t,			expectedReturnCode );
 	NODE_LOAD_DEPS( 0,			preBuildDependencies );
+	NODE_LOAD_DEPS( 0,			additionalDependencies );
 
 	NodeGraph & ng = FBuild::Get().GetDependencyGraph();
 	Node * srcNode = ng.FindNode( sourceFile );
@@ -127,7 +131,8 @@ ExecNode::~ExecNode()
 								  arguments,
 								  workingDir,
 								  expectedReturnCode,
-								  preBuildDependencies );
+								  preBuildDependencies,
+								  additionalDependencies );
 	ASSERT( n );
 
 	return n;
@@ -144,6 +149,7 @@ ExecNode::~ExecNode()
 	NODE_SAVE( m_WorkingDir );
 	NODE_SAVE( m_ExpectedReturnCode );
 	NODE_SAVE_DEPS( m_PreBuildDependencies );
+	NODE_SAVE_DEPS( m_AdditionalDependencies );
 }
 
 // EmitCompilationMessage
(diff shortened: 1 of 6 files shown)


ffulin:
It seems like this change as implemented simply duplicates the functionality of .PreBuildDependencies - PreBuildDependencies.

I imagine if you wanted to support multiple inputs via ExecInput, you'd generally want to pass those to the command being executed (i.e. expanding %1 multiple times).

If you wanted to do that, you'd need to:
a) Get one or more nodes in FunctionExec::Commit:
```C++
if ( !GetNodeList( funcStartIter, ".ExecInput", deps ) )
```
b) Expand %1 (or "%1") multiple times at build time
See CSNode::GetInputFiles for an example

NOTE: This would still support providing multiple input files, but only passing one of them to your executable - you'd do that in the BFF by specifying the file manually in your .ExecArguments (instead of using %1).

inbilla:
So, the key difference between PreBuildDependencies and this new variable, is that this new variable is added directly to "StaticDependencies" in the constructor of the ExecNode.
I did a search around, this means it should have the same treatment as "SourceFile" with regard to the dependency graph (as far as I can see). That means that when one of the "AdditionalDependencies" changes, then it'll re-trigger the command to be executed.

you said this:
>I imagine if you wanted to support multiple inputs via ExecInput, you'd generally want to pass 
>those to the command being executed (i.e. expanding %1 multiple times).
That's actually not the case here.
The command only needs to execute _once_, and during it's execution, it might take the contents of many files, and let them drive the generation of one (or many) outputs. But there is no mapping of "input file" -> "output file" in the case of custom commands. It's (0-N inputs) -> (0-N outputs) with a single command execution.

This means that we don't want to execute the command multiple times (one for each input).
And there is no input expansion that would make sense in all cases. Instead the most general way to be able to specify this is to be able to define a list of files to "depend" on, but not use them directly in the command (not that FASTBuild explicitly knows about anyway).

In the situation where we need to call the command multiple times for each input, I believe that's the realm of one of the compilation commands... as they support the output filename conversion using patterns already too. It seems like a duplication of functionality.

I'm happy to make changes - just the suggested ones don't fit the current use-case yet :)
infact, thinking about it... the variable name probably should be "ExecExtraInputs" or something like that, rather than "AdditionalDependencies"... 
Let me know what you think anyway, Thanks for your time!

ffulin:
**> That's actually not the case here. The command only needs to execute once...**
I didn't mean that you would run the executable more than once (that's what ObjectList is for), but rather that all the files would be passed on one command line.
So this:
```C++
Exec( 'Test' )
{
    .ExecInput = { 'a.cpp', 'b.cpp', 'c.cpp' )
    .ExecExecutable = 'thing.exe'
    .ExecArguments = '%1'
```
would invoke thing.exe once as follows:
```C++
thing.exe a.cpp b.cpp c.cpp
```
Also, to clarity my NOTE: at the end; If you wanted to only pass one of your files to the executable, you could do this:
```C++
Exec( 'Test' )
{
    .ExecInput = { 'a.cpp', 'b.cpp', 'c.cpp' )
    .ExecExecutable = 'thing.exe'
    .ExecArguments = 'a.cpp'
```
which would result in:
```C++
thing.exe a.cpp
```
Does that make sense?

The two changes I proposed:
a) make the ExecInput accept multiple files
b) expand the %1 token multiple times like in CSAssembly
would give the above behaviour.

inbilla:
Thanks Franta,
That all sounds fine, I'll make the changes as soon as possible (probably in the next two days).

inbilla:
Updated... seems to be working.
Let me know what you think. I drew code from CSNode to help with the expansion...
Is there somewhere that code could go as a utility instead of being on the individual classes?

Code/Tools/FBuild/FBuildCore/Graph/ExecNode.cpp
```diff
@@ -20,23 +20,22 @@
...
 , m_Executable( executable )
 , m_Arguments( arguments )
 , m_WorkingDir( workingDir )
 , m_ExpectedReturnCode( expectedReturnCode )
 {
-	ASSERT( sourceFile );
 	ASSERT( executable );
 	m_StaticDependencies.SetCapacity( 2 );
-	m_StaticDependencies.Append( Dependency( sourceFile ) );
+	m_StaticDependencies.Append(m_InputFiles);
```
ffulin: Minor change:
m_StaticDependencies.SetCapacity( m_InputFiles.GetSize() + 1 );
would avoid some memory allocations

ffulin:
Looks good to me.  

It's a shame about the duplicate of the arg expansion as you pointed out, but I wouldn't worry about ift for now.  The bigger function is subtly different in every case and calls the other function which is different in most cases. If it starts to look like a maintencance problem in the future I can try to refactor it then.

inbilla:
Good catch, Sorted out the preallocation.

Again, would you like me to update the documentation + unit tests?
I have to admit, I'm not tooo comfortable with the CSNode code there since I'm not using it, I've not tested it enough for me to be happy with.

ffulin:
I think this change is good to go.

If you want to have a crack at adding a test/updating the docs, that would be great.  Let me know if you have any trouble with it.

Same info as per the other ticket:
```
UnitTests and documentation would be awesome. They all should be passing, but some changes
might be needed to get them working locally:
 - paths to compilers (like the main compilation of FASTBuild)
 - Clang for Windows to run the clang-specific tests (and the path to it)

If you take a look in testcommon.bff, everything should be in there.
```

inbilla:
Hi Franta,

I'm just looking into getting this pull request ready again.
Thinking more on it, this one actually breaks backwards compatibility.
The ExecInput argument is changed from a string to an array.
Do you have a preferred way to handle this?

Let me know how you would like to proceed :)

Josh

inbilla:
Closing for now, since it isn't currently ready

ffulin:
Changing String to ArrayOfStrings is fine for backwards compat.  We can just convert a single string into an ArrayOfStrings with one thing in it.

In your change you are calling GetNodeList which already handles that, so on this front at least, there is no problem.

PR #27, opened 2015-07-05
Adding ExecUseStdOutAsOutput to ExecCommand

inbilla:
Exec command was generating a warning when defining a dummy output.

I doubt this is the best way to do it, but it removed the warning for me at least.
If you have any other ideas, let me know.


ffulin:
Silencing this warning is hiding two potential problems:
* Anything in the graph that depends on this node is going to fail because no output is generation. If your Exec is the final output this may not be a problem
* Because there is no output, your Exec step will be run in every build.  This is almost never what you want.

What situation do you have where you are calling Exec, but it doesn't create anything?

inbilla:
I agree, it could hide potential problems.
The use case is with post build steps that do not generate outputs. This is quite common in most build systems. In CMake I'm seeing a common occurrence of Post-Build steps defined in CMake not having Outputs. Whilst I agree it takes away the fast path from the user, without supporting "no output" custom commands, then there is a class of functionality being denied to users (build time reporting, metric reporting, pdb submission, source indexing etc). Basically - end result - CMake doesn't require an output to be defined for a custom command.

So, what alternative could I add?
Easiest option I think would be adding a variable ".ExecGeneratesOutput" and my code can set it to "false". (But still supply a dummy output filename). Then we add a function to check if the output exists (ExecNode can override this function and check the generates output variable). Then the warning code can call this check function instead, before outputting the warning.

Thoughts? 
If you're happy with that, I'll put together another pull request with that in it...

ffulin:
I guess there are three classes of "post-build step":
a) Generates an output (like a report)
b) Modifies the executable/dll in-place (like stripping debug info or patching in a version string)
c) Writes to or runs some external system(like source indexing/pdb storage).

**a)** This can be handled with the current setup (setting the ExecOutput of an Exec step to the generated result file). 
**b)** This can be handled in most cases by the .LinkerStampExe option, which is essentially a "run a post build step in-place".  (You might think this could also handle c), but I think there is a better way)

To do **c)** in a safe/nice (and not too difficult way), that also doesn't have to run every time, I think the right way to do this is to extend Exec() with a new option as follows:
```C++
Exec( 'sourceindexing' )
{
    .ExecUseStdOutAsOutput = true
    .ExecOutput = 'tmp/sourceindexing.txt'
```
If this new option is set, we'd write the (already captured) stdout from the process to the output file specified in the existing ExecOutput param.  In this way, the normal datestamps checks would cause this step to only be run when needed (i.e. something source indexing would be run only when the executable is re-linked, rather than every time).

Additionally, it can be targeted like any other real node on the command line.

What do you think?

inbilla:
OK, so,  I'm not sure how feasible it is to use the LinkerStampExe option from cmake... It certainly would complicate things... So not at this point anyway.

Your suggestion for using stdout is possible. I'm not sure it would cover all use cases... For example... I'm not sure how to express something that doesn't make an output.. But must always be run (for some reason). This is possible in cmake, though im not 100 percent confident of the distinction between the two (if there is any)... And so how to detect that case and adjust the output would be difficult. I would imagine dependencies would still be defined in cmake, but mostly to enforce ordering... But that's not clear given the information made available during generation. They would just look like inputs.

In the meantime though, I'm happy to make these changes and try them? 
I'm just not sure they'd facilitate all the use cases I need.

To be honest, always running the command, whilst not optimal, is bound to be desired behaviour at some point. So facilitating it explicitly, and calling it out as not recommended would be my preference....

inbilla:
Further reading of this page http://www.cmake.org/cmake/help/v3.2/command/add_custom_command.html
Suggests that the only way to not have an output is with pre and post build steps. And those are only run if the target is built. So let's try your suggestion and hopefully that covers all the bases.

ffulin:
Even if the executable you run doesn't emit anything, I think this will work - it will capture the empty stdout and write a 0 byte file.

If we want to force an Exec step to run, I wouldn't be opposed to adding a new option:
```C++
    .ExecAlwaysRun = true
```
that way by default you have the optimal behaviour, but you can opt in to the slower case if that's useful.

inbilla:
Sounds good

inbilla:
Ok, This appears to be working.
Let me know if there is anything you would like changed.

Would you like me to sort out some UnitTests and update the documentation too?
I'll have to figure out how your testing works.. I ran some earlier on the version of Dev that I have (a few commits old), but it appeared a number were currently failing. 
Anyway, let me know

ffulin:
Looks good to me.  The DB version will need to be bumped, but I'll take care of that when doing the release.

UnitTests and documentation would be awesome.  They all should be passing, but some changes might be needed to get them working locally:
 - paths to compilers (like the main compilation of FASTBuild)
 - Clang for Windows to run the clang-specific tests (and the path to it)

If you take a look in testcommon.bff, everything *should* be in there.

I think your change is pretty safe, so if you have problems with the other tests, let me know. If you just get your new test passing, that would probably be enough.

Thanks!

inbilla:
Hi Franta, sorry I haven't been working on this much lately!
I've now updated the unit tests... 
What are your thoughts?

I'm wondering if this will need to change again when exec node supports multiple outputs.
In that case, maybe we want to actually define the output file for the stdout data....

What blockers do you see to merging this in now?

ffulin:
The code looks good.  I'll have to integrate and run it to verify all the tests, but they look of at a glance also. 

One thing I did notice though is that you are missing the REGISTER_TESTGROUP macro in TestMain.cpp which is needed to activate the test.  I'll add that though - no big deal.

Looking at the test, my guess is that you developed/tested it on Windows only, right?

PS - Updating test and docs is much appreciated!

inbilla:
Yes, only developed in Windows.

You're right about the Register_TestGroup macro, I must have missed it in
the commits because I actually heavily changed that file :) (Disabled
everything but my test at one point).

No worries about the test and docs.
I'll move onto the other one once this pull request has gone in, it
modified some of the same lines, and will need to use these new test files
too.

Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
```diff
@@ -587,7 +587,8 @@ void JobQueue::FinishedProcessingJob( Job * job, bool success, bool wasARemoteJo
 			ASSERT( ( result == Node::NODE_RESULT_OK ) || ( result == Node::NODE_RESULT_OK_CACHE ) );
 
 			// (don't check existence of input files)
-			if ( node->GetType() != Node::FILE_NODE )
+			if ( node->GetType() != Node::FILE_NODE &&
+				node->GetType() != Node::EXEC_NODE)
```
ffulin: While reviewing the change for integration into the mainline, I noticed this.
We should always verify the output of a node. This whole change is in order to satisfy this constraint. Without doing this, we could have some very hard to debug build problems.

(The only exception is a FileNode since that is a leaf input to the build which we rely on the dependent node (the compiler) to report (as a missing include typically))

I'll revert this in the mainline integration.

Is this an error or was there some edge case this was covering?

Code/Tools/FBuild/Documentation/docs/functions/exec.html
```diff
@@ -21,15 +21,17 @@
...
-  .ExecInput            ; Input file to pass to executable
-  .ExecOutput           ; Output file generated by executable
-  .ExecArguments        ; (optional) Arguments to pass to executable
-  .ExecWorkingDir       ; (optional) Working dir to set for executable
+  .ExecExecutable         ; Executable to run
+  .ExecInput              ; Input file to pass to executable
+  .ExecOutput             ; Output file generated by executable
+  .ExecArguments          ; (optional) Arguments to pass to executable
+  .ExecWorkingDir         ; (optional) Working dir to set for executable
+  .ExecReturnCode         ; (optional) Expected return code from executable (default 0)
```
ffulin: Thanks for adding this missing ExecReturnCode doc as well :)

ffulin:
I've integrated this into the Perforce mainline. I made some small tweaks:
 - simplify test exec.bff and make it closer to platform agnostic
 - activate test (Windows only)

This change will be in the next release (v0.87)

inbilla:
Thanks Franta!
Regarding the output check, that must have been left behind while I was
moving commits around in Git.
Reverting the change should be fine.

Thanks!

Issue #24, opened 2015-07-04
/ShowIncludes and /P /Fi cl.exe flags


poppolopoppo:
Hi,

I need to call the preprocessor in my build chain. For that I use an ObjectList() with my compiler and just add '/P /Fi "%2"' to _.CompilerOptions_, which causes cl.exe to [write preprocessor output](https://msdn.microsoft.com/en-us/library/ee207218.aspx) to %2.

This fits in nicely in theory, but I got an access violation when running it. With a debugger I found the crash in ObjectNode::ProcessIncludesMSCL(), dereferencing output which was null. This is only asserted and not handled :
````cpp
// FBuild\FBuildCore\Graph\ObjectNode.cpp(437)
// ProcessIncludesMSCL
//------------------------------------------------------------------------------
bool ObjectNode::ProcessIncludesMSCL( const char * output, uint32_t outputSize )
{
	Timer t;

	{
		ASSERT( output && outputSize );

		CIncludeParser parser;
		bool result = parser.ParseMSCL_Output( output, outputSize );
		if ( result == false )
		{
			FLOG_ERROR( "Failed to process includes for '%s'", GetName().Get() );
			return false;
		}
````
Called from here :
````cpp
    // FBuild\FBuildCore\Graph\ObjectNode.cpp(247)
	// compiled ok, try to extract includes
	if ( ProcessIncludesMSCL( ch.GetOut().Get(), ch.GetOutSize() ) == false )
	{
		return NODE_RESULT_FAILED; // ProcessIncludesMSCL will have emitted an error
	}
````

The problem seems to come from the /ShowIncludes flag of cl.exe. [On the MSDN](https://msdn.microsoft.com/en-us/library/hdkef6tk.aspx) we can read :
> The /showIncludes option emits to **stderr**, not **stdout**.

For some obscure reason this is not the case when calling cl.exe with "/c %1 /Fo %2" or even by adding "/E" or "/EP", in those cases -and in fact most of the time- the output is written on **stdout**.
But when I add "/P /Fi" then /ShowIncludes will write on **stderr**, stdout will be empty and ch.GetOutput() will return nullptr.

ffulin:
As you pointed out, /showIncludes seems to go to stdout.  I'm very surprised to see the documentation say otherwise!

We could fix this problem in two ways:

**Option A - Emulate /Fi**
 - detect and note the /Fi flag
[...]

PR #34, opened 2015-07-11
Support for /Fi flag of cl.exe

poppolopoppo:
Avoid a crash when using cl.exe /Fi flag, which causes /showIncludes flag to be dumped on stderr instead of stdout.

* Detects the /Fi flag ;
* Disable caching/distribution ;
* Go through the same non-cache/non-dist path like when we do /showIncludes already ;
* Pass the stderr /showIncludes output to the MSVC function because we noted the /Fi before.

Fixes the issue described in https://github.com/fastbuild/fastbuild/issues/24.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -244,8 +244,24 @@ ObjectNode::~ObjectNode()
 		return NODE_RESULT_FAILED; // SpawnCompiler has logged error
 	}
 
+	const char *output = nullptr;
+	uint32_t outputSize = 0;
+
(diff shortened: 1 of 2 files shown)


ffulin:
I have integrated this change into the P4 mainline and this will be part of the next release (v0.84).

I had to change the behaviour slightly to make it work correctly in all use cases.  TL;DR: It checks for /P instead of /Fi.

This means you can now use it two ways:
* /P and use .CompilerOutputPath and .CompilerOutputExtension to tell FASTBuilld where cl.exe will put the .i file (next to the cpp file by default)
* /Fi"%2" : Tell cl.exe where to put the output file to it matches the default location expected by FASTBuild

I added a Test case to cover this bug fix as well.

PR #36, opened 2015-07-12
Fixing a linker error in some project configurations where the implib destination dir does not exist

inbilla:
Fixing a linker error in some project configurations where the implib is stored in a directory that does not yet exist. The DLL command will now make sure that the implib destination directory will exist before executing the linker.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/DLLNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/DLLNode.cpp
@@ -37,6 +37,21 @@ DLLNode::~DLLNode()
 {
 }
 
+// DoBuild
+//------------------------------------------------------------------------------
+/*virtual*/ Node::BuildResult DLLNode::DoBuild( Job * job )
+{
+	// Make sure the implib output directory exists
+	if ( m_ImportLibName.IsEmpty() == false)
+	{
+		AStackString<> cleanPath;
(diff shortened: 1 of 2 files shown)


Code/Tools/FBuild/FBuildCore/Graph/DLLNode.cpp
```diff
@@ -37,6 +37,21 @@ DLLNode::~DLLNode()
...
+// DoBuild
+//------------------------------------------------------------------------------
+/*virtual*/ Node::BuildResult DLLNode::DoBuild( Job * job )
+{
+	// Make sure the implib output directory exists
+	if ( m_ImportLibName.IsEmpty() == false)
+	{
+		AStackString<> cleanPath;
+		NodeGraph::CleanPath( m_ImportLibName, cleanPath );
+		EnsurePathExistsForFile( cleanPath );
```
ffulin: Need to handle an error here:
	if ( EnsurePathExistsForFile( cleanPath ) == false )
	{
		return NODE_RESULT_FAILED; // EnsurePathExistsForFile will have emitted error
	}

ffulin:
Apart from the missing error handling, this looks good.

inbilla:
Excellent, thanks for catching that.
All sorted now.

ffulin:
This change has been integrated into the Perforce mainline and will be in the v0.84 release.

Thanks for the fix!

Issue #33, opened 2015-07-09
CIncludeParser does not handle localization


poppolopoppo:
Hi again,

So, after my pull request about unity time stamp "fix" (https://github.com/fastbuild/fastbuild/pull/18) I  continued to investigate my false-nothing-to-build-state while using unities. 

I am French and my compiler does not output :
````
Note: including file: X:\A\B\C
````
But :
````
Remarque : inclusion du fichier : X:\A\B\C
````
Which is not recognized by the parser, resulting in ObjectNodes without dynamic dependencies.
The problem is coming from [CIncludeParser::ParseMSCL_Preprocessed()](https://github.com/fastbuild/fastbuild/blob/master/Code/Tools/FBuild/FBuildCore/Helpers/CIncludeParser.cpp#L54), which does not handle  compiler message localization :

In the end It was not just the unities, but all ObjectNodes that were missing their includes.
Your code is handling all the dependencies nicely, It just needs a better include parser ;)

ffulin:
There are two different include extraction paths for the MS compiler. The other one uses the preprocessor and checks for #line directives in the preprocessed output, so at least that one should be ok (since #line is a keyword and not localized).  (This path is used when compiling in distributed mode or using the cache).

The "Show Includes" path is used when we don't need to do 2-pass compilation as it's a little bit faster. Even a distributed build can have some nodes go through this path. It's unfortunate that these messages are localized, but not that surprising.

I guess the fix for this will have to be to change the parser to search for the pattern:
.*:.*: path

poppolopoppo:
Yes the other method is not a problem. For now I just use a simple fix which greps "Remarque : inclusion du fichier :" in addition to "Note: including file:". And everything behind is now
working like a charm, I am so glad to have fix this very annoying dependency problem. 

I'll surely do a pull request with a more robust solution. Parsing `.:.: path` seems ok, but this parsing rule is pretty weak and it will be necessary to check if parsed file names exist to guard against false positives (the current parser has a strong enough parsing rule to bypass that).

ffulin:
I think checking for the existence of the files will be too slow. Note that this is only parsing in the /showIncludes case, where there is typically very little other output. I think the chance of false positives is very low.

poppolopoppo:
Ok, I added a [pull request](https://github.com/fastbuild/fastbuild/pull/37) with a fix (which is not checking files for existence).

ffulin:
[...]

PR #37, opened 2015-07-13
Fix support of /showIncludes with non-english Visual Studio

poppolopoppo:
CInludeParser::ParseMSCL_Output() now supports localized compiler messages.

The new parsing rule is `.+:.+:\s+[^<>"|?*]+`.
Previous rule `\nNote: including file: ` assumed English language.

Parsed included files are assumed to be existing.

Relates to the issue https://github.com/fastbuild/fastbuild/issues/33


ffulin:
This has been integrated into the Perforce mainline and will be part of the next release. (v0.84)

Thanks for the fix!

PR #46, opened 2015-08-23
New function Eval()

poppolopoppo:
Eval( .var ) performs variable substitutions on the given String, ArrayOfStrings, Struct or ArrayOfStructs :

``` javascript
.Foobar = "foo bar"
.Barfoo = "^$Foobar^$ 42" // '^$' is already parsed as '$' on master thanks to escaping
Print( .Barfoo ) // "$Foobar$ 42" 

Eval( .Barfoo )
Print( .Barfoo ) // "foo bar 42"

.Barfoos = { "^$Foobar^$", "^$Barfoo^$" }
Print( .Barfoos ) // { "$Foobar$", "$Barfoo$" } 

Eval( .Barfoos )
Print( .Barfoos ) // { "foo bar", "foo bar 42" }

.Foofoo = 
[
    .Foo = [ .a =  '^$Foobar^' ]
    .Bar = [ .b =  { "^$Foobar^$", "^$Barfoo^$" } ]
    .All = { .Foo, .Bar }
]
Eval( .Foofoo )
Print( .Foofoo ) // [ 
                 // .Foo = [ .a = "foo bar" ] .
                 // .Bar = [ .b = { "foo bar", "foo bar 42" } ] 
                 // .All = { [ .a = "foo bar" ], [ .b = { "foo bar", "foo bar 42" } ] }
                 // ]
```

Allows the following BFF configuration :

``` javascript
.BuildA =
[
    .Platform   = "Win32"
    .Config     = "Debug"
]
.BuildB =
[
    .Platform   = "Win32"
    .Config     = "Release"
]

.AllBuilds = { .BuildA, .BuildB }

.ProjectOutput = "MyExe-^$Platform^$-^$Config^$.exe"
.Libraries = 'MyLib-^$Platform^$-^$Config^$'
.Libpaths = { 'Bin/^$Platform^$/^$Config^$', 'External/^$Platform^$/^$Config^$' }

ForEach( .Build in .AllBuilds )
{
    Using( .Build )

    Eval( .ProjectOutput )
    Print( .ProjectOutput ) // "MyExe-Win32-Debug.exe"
                            // "MyExe-Win32-Release.exe"

    Eval( .Libraries )
    Print( .Libraries ) // "MyLib-Win32-Debug"
                        // "MyLib-Win32-Release"

    Eval( .Libpaths )
    Print( .Libpaths )  // { 'Bin/Win32/Debug', 'External/Win32/Debug' }
                        // { 'Bin/Win32/Release', 'External/Win32/Release' }
}
```

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
@@ -10,6 +10,7 @@
 #include "FunctionCopyDir.h"
 #include "FunctionCSAssembly.h"
 #include "FunctionDLL.h"
+#include "FunctionEval.h"
 #include "FunctionError.h"
 #include "FunctionExec.h"
 #include "FunctionExecutable.h"
@@ -98,6 +99,7 @@ Function::~Function() = default;
     g_Functions.Append( FNEW( FunctionCSAssembly ) );
     g_Functions.Append( FNEW( FunctionDLL ) );
     g_Functions.Append( FNEW( FunctionError ) );
+    g_Functions.Append( FNEW( FunctionEval ) );
     g_Functions.Append( FNEW( FunctionExec ) );
     g_Functions.Append( FNEW( FunctionExecutable ));
     g_Functions.Append( FNEW( FunctionForEach ) );
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionEval.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionEval.cpp
@@ -1,0 +1,205 @@
+// FunctionEval
+//------------------------------------------------------------------------------
+
+// Includes
+//------------------------------------------------------------------------------
+#include "FunctionEval.h"
+#include "Tools/FBuild/FBuildCore/BFF/BFFParser.h"
+#include "Tools/FBuild/FBuildCore/BFF/BFFStackFrame.h"
+#include "Tools/FBuild/FBuildCore/BFF/Tokenizer/BFFTokenRange.h"
+#include "Tools/FBuild/FBuildCore/FLog.h"
+#include "Tools/FBuild/FBuildCore/Graph/NodeGraph.h"
+
+// CONSTRUCTOR
+//------------------------------------------------------------------------------
+FunctionEval::FunctionEval()
+: Function( "Eval" )
+{
+}
+
+// AcceptsHeader
+//------------------------------------------------------------------------------
+/*virtual*/ bool FunctionEval::AcceptsHeader() const
+{
+    return true;
+}
+
+// NeedsHeader
+//------------------------------------------------------------------------------
+/*virtual*/ bool FunctionEval::NeedsHeader() const
+{
+    return true;
+}
+
+// NeedsBody
+//------------------------------------------------------------------------------
+/*virtual*/ bool FunctionEval::NeedsBody() const
+{
+    return false;
+}
+
+//------------------------------------------------------------------------------
+/*virtual*/ bool FunctionEval::ParseFunction(   NodeGraph & nodeGraph,
+                                                BFFParser & parser,
+                                                const BFFToken * functionNameStart,
+                                                const BFFTokenRange & headerRange,
+                                                const BFFTokenRange & bodyRange ) const
+{
+    (void)nodeGraph;
+    (void)parser;
+    (void)functionNameStart;
+    (void)headerRange;
+    (void)bodyRange;
+
+    // we want 1 frame above this function
+    BFFStackFrame * parentFrame = BFFStackFrame::GetCurrent()->GetParent();
+    ASSERT( parentFrame );
+
+    // parse it all out
+    BFFTokenRange headerIter = headerRange;
+    while ( headerIter.IsAtEnd() == false )
+    {
+        // Expect an existing source variable
+        {
+            const BFFToken * localVarToken = headerIter.GetCurrent();
+            headerIter++;
+            if ( localVarToken->IsVariable() == false )
+            {
+                Error::Error_1200_ExpectedVar( localVarToken, this );
+                return false;
+            }
+
+            // Resolve the name of the local variable
+            AStackString< BFFParser::MAX_VARIABLE_NAME_LENGTH > localName;
+            bool localParentScope = false;
+            if ( BFFParser::ParseVariableName( localVarToken, localName, localParentScope ) == false )
+            {
+                return false;
+            }
+
+            // find variable
+            const BFFVariable * v = nullptr;
+            BFFStackFrame * const varFrame = ( localParentScope )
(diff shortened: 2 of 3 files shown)


Agga:
What is the status of this pull request? I think this feature is quite handy.

Here is another usecase:

    .IncludePaths = { 'libA/inc/' }
    .IncludeFormatter = ' /I"^$Include^$"'
    ForEach( .Include in .IncludePaths )
    {
	    Eval( .IncludeFormatter )
	    ^CompilerOptions + .IncludeFormatter
    }


All shown examples utilize lazy string evaluation.
I am not sure if this use case is intended but it
might be useable as some sort of pseudo functions.

     // theory
     // lazily evaluate structs
     .someStruct = [
	     .someData = "^$someData^$"
     ]
     
     Using( .someStruct )
     Print( .someData ) //^$someData^$
     
     .someData = 'A'
     Eval( .someStruct )
     Using( .someStruct )
     Print( .someData ) //A


Going back on my example above we are able to
do somehting like this?

    .formatIncludes = [
	    .IncludePaths = "^$IncludePaths^$" // lazy evaluation
	    .CompilerOptions = ^$CompilerOptions^$" // lazy evaluation
	    .IncludeFormatter = ' /I"^$Include^$"'
    
	    ForEach( .Include in .IncludePaths )
	    {
		    Eval( .IncludeFormatter )
		    ^CompilerOptions 
	    }
    ]
    
    .IncludePaths = { 'pathA', 'pathB' }
    Eval( .formatIncludes )
    Using( .formatIncludes )

ffulin:
The hesitation in integrating this change is that (as with all syntax changes), we need to be sure it's exactly what we want in the long term. We essentially can't change or remove the syntax in the future since we need to maintain backwards compatibility.

This pull request serves an important need, but I'm not super happy about the syntax. In particular the need to escape things feels a bit messy.

I've been thinking for quite some time how to solve this problem (and a few others). I think the answer is a proper function/macro/template system. This would let you define any valid set of bff, but with delayed expansion.

So imagine you declare a helper to expand includes:
```C++
#declare AddIncludePaths( .Includes )
{
    ForEach( .Include in .Includes )
    {
        ^CompilerOptions + ' /I"$Include$"'
    }
}
```
You could then utilize it later on:
```C++
.IncludePaths = { 'libA/inc/' }
#use AddIncludePaths( .IncludePaths )
```
There are still some things to think about, for example:
* The exact syntax
* Should arg types be explicit?

I think this could be very powerful, solving the problems addressed by this pull request, as well as allowing entire chunks of config to be factored out.

What do you think?

Agga:
I like your idea of custom functions with lazy evaluation, but i think the 'preprocessor' like syntax isn't flexible enough. Have you considered something like functors? 

// declaration
```
// declare a function with 0 arguments
.Function0 = ( /*empty*/ ){
	Print( "No Arguments" )
}

// declare a function with 1 argument
.Function1 = ( .Argument ){
	Print( .Argument )
}
```

// calling
```
// call a function
.foo = 'bar'
Call( .Function0 )
Call( .Function1, .foo )
Call( /*NAME*/, 0-N ARGUMENTS )
```

// benefits of functors
```
// assignment
.Function = .Function0
Call( .Function )

// assignment with dynamic construction
.Number = '0'
.Function = ."Function$Number$"
Call( ."Function$Number$" )
Call( .Function )

// assignment by "scope inheritance"
.Foo = [
	.Bar = (){ 
		Print( "Foobar" ) 
	}
]

Using( .Foo )
Call( .Bar )

// experimental
// needs new conversion rule for bool -> string 
.UsePCH = false
.UsePCH_true = ( /*empty*/ ){} // handle pch usage
.UsePCH_false = ( /*empty*/ ){} // default case
Call( ."UsePCH_$UsePCH$" )
```

Now for the fun stuff; how are arguments passed to a function?I can think of two ways, one being a positional approach and one being a name based one ( coherent to built-ins ), or a mix of both?
What do you think about this argument matching?
```
.Func = ( .Foo, .Bar ){
	Print( .Foo )
	Print( .Bar )
}

.Foo = 'foo'
.Bar = 'bar'

a) Call( .Func, .Foo, .Bar ) // foo bar
b) Call( .Func, .Bar, .Foo ) // bar foo
c) Call( .Func, .Bar ) // bar bar
d) Call( .Func ) // foo bar
```

Are functions able to return something?  I think this topic is related to scopes aswell.
A functor invocation could introduce a new scope for the duration of it's execution. At the end of the scope all changes are lost unless assigned to variables in the outer scope.
```
.Foo = 'foo'
.Function0 = (){
	^Foo = 'bar'
}
.Function1 = (){
	^Bar = 'bar'
}

Print( .Foo ) // foo
Call( .Function0 )
Print( .Foo ) // bar

a) 	Call( .Function1 ) // error, ^Bar is unknown
b) 	Call( .Function1 )
	Print( .Bar ) // bar
```
I'd like to hear your opinion on this.

poppolopoppo:
I like the callback approach, you could even pass functions as arguments to another function.

This also kind of introduces javascript-like prototype programming :
````
.MyPrototype = 
[
     .MyMember = 0
     .MyArray = {'1','2','3'}

     .Build = () {
         // fun stuff goes here :
         ^MyMember = 42
     }

    .DoThings = (.Context) {
        // this can solve my lazy evaluation problem :
        Using( .Context ) 
        // callee may not know that .Context contains those :
        Print( "$SomeVars$ $Defined$ $InContext$" )
    }

    .EachItem = (.Functor) {
        // my first algorithm :
        ForEach( .it in ^MyArray ) {
            Call( .Functor, .it )
        }
    }
]

.MyOtherPrototype = 
[
    Using( .MyOtherPrototype )
]
````
This is evil :) 

You don't have a member access operator to do `Call(.MyPrototype->Build)` for now though, and this may be way too overkill for a language that has no if conditions (well tbf you can do `Using( ."Branch_$Test$" )` now :p) but definitely more powerful than the `#declare` approach.

**Comment by kylawl** [comment]
Hey gang, I just realized looking at this PR that its' been like a decade since we chatted about the parent scope variable push thing.

I've been thinking about this eval proposal a lot lately since I've been refactoring the fbuild files in our engine and doing some new work. I think I've got an alternative solution that maybe you might find agreeable. I've supplied a simple example, let me know what you think.

```
.VarA = 'MrMcGoo'

// Double dollar variables are deferred to the next substitution use
.DeferredSubstituionString = 'Foo-$VarA$-$$DeferredB$$' // Result: 'Foo-MrMcGoo-$DeferredB$'

.DeferredB = 'MsMcFoo'

// Substitution resolves $DeferredSubstituionString$
.CompleteString = 'Hello-$DeferredSubstituionString$' // Result: 'Hello-Foo-MrMcGoo-MsMcFoo'

```
I like this because 
- It looks like variable substitution but different enough to see that it's special
- It doesn't change the way escaped strings evaluate
- It only works 1 level deep so it's not terribly complex to implement
- Variable must exist when the deferred substitution resolves so missing variable errors behave the same

PR #49, opened 2015-08-26
MSVC can use the lib command OR the link /LIB command (lib apparently…

inbilla:
Hi Franta,

Here is a small change to fix a discrepancy that I noticed whilst implementing the "LinkerType" parameter. This flag only appears to affect response file usage in this command, but it probably should be fixed anyway? It looks like Cmake only ever uses the link.exe command style for libraries.

Thanks


ffulin:
That's really strange, especially since this isn't documented as an option on the MSDN documentation!  No problem taking the fix though.

ffulin:
This fix has been integrated into the Perforce mainline and will be in the next version (v0.87)

PR #52, opened 2015-08-30
Fix Visual Studio project GUID generation

arttupeltonen:
This commit fixes GUID generation so that identical project GUIDs are generated for .sln and .vcxproj files. If the full path to a project is "my/path/projectName.vcxproj", the GUID is generated based ond the string "projectName". 

Root of the problem was that 
- VCXProjectNode.cpp generated the GUID from the string "projectName.vcxproj"
- SLNGenerator.cpp generated the GUID from the string "\\projectName"

Updated TestProjectGeneration/fbuild.bff to include a generated solution. Ideally the unit test would verify that GUIDs between the solution and project match, but couldn't find a way to do this quickly.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
@@ -140,27 +140,29 @@ void SLNGenerator::WriteProjectListings(    const AString& solutionBasePath,
         // check if this project is the master project
         const bool projectIsActive = ( solutionBuildProject.CompareI( (*it)->GetName() ) == 0 );
 
-        // project relative path
         AStackString<> projectPath( (*it)->GetName() );
-        projectPath.Replace( solutionBasePath.Get(), "" );
 
         // get project base name only
-        const char * p1 = projectPath.FindLast( NATIVE_SLASH );
-        const char * p2 = projectPath.FindLast( '.' );
-        AStackString<> projectName( p1 ? p1 + 1 : projectPath.Get(),
-                                    p2 ? p2 : projectPath.GetEnd() );
+        const char * lastSlash  = projectPath.FindLast( NATIVE_SLASH );
+        const char * lastPeriod = projectPath.FindLast( '.' );
+        AStackString<> projectName( lastSlash  ? lastSlash + 1  : projectPath.Get(),
+									lastPeriod ? lastPeriod		: projectPath.GetEnd() );
 
         // retrieve projectGuid
         AStackString<> projectGuid;
         if ( (*it)->GetProjectGuid().GetLength() == 0 )
         {
-            AStackString<> projectNameForGuid( p1 ? p1 : projectName.Get() );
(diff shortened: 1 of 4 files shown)


ffulin:
The code as it stands should be working correctly.  In the project/solution generation for FASTBuild itself, the guids match and the sln is not dirtied on load.  It should use "/projectName.vcxproj" in both cases (which is admittedly a bit arbitrary).

From what I can see, there might be a mismatch if the paths to the slashes are mismatched somehow, i.e. perhaps it's relying on native slashes which are not fixed up in one of the two cases. This would be the bug to fix.

Unfortunately, we can't simply change the generation in both places as you've done, because this will break for existing users who have created their own solution files by hand.

ffulin:
Assuming the bug is related to slashes, we could fix the problem as follows:
 - make both places pass just the projectName part of "/projectName.vcxproj"
 - have the GenerateGuid function format that back into a string which always matches the old behaviour (with the slash and extension)

This way, the code would be more sensible at each location (passing the project name) while centralizing the arbitrary logic that we don't want to break because of backwards compatibility.  The GenerateGuid function could contain a comment about what that arbitrary formatting is there.

arttupeltonen:
Thanks for the response, I didn't really think about backwards compatibility earlier. I looked into this further, and the root of the problem is actually this line in SLNGenerator.cpp:

    projectPath.Replace( solutionBasePath.Get(), "" );

For FASTBuild itself, the solution and projects output directories are different. Because `solutionBasePath` includes a trailing slash, this results in "dir/project.vcxproj" if the output directory is different and "project.vcxproj" if it's the same. This would cause the GUID to be generated from "/project.vcxproj" from the VCXProject side and "project.vcxproj" from the VSSolution side, specifically in the case where the .sln and the .vcxproj files are output into the same directory.

The solution of moving the logic into `FormatDeterministicProjectGUID` has a problem regarding backwards compatibility that it would cause also the solution folders to get a GUID based on "/FolderName.vcxproj" which is probably not desirable. I would propose just adding the missing `NATIVE_SLASH` inside `SLNGenerator::WriteProjectListings` when necessary, with a comment explaining the behavior. What do you think?

arttupeltonen:
There seems to be a separate bug with the ".vcxproj" suffix sometimes being left out, but that looks like a simple logic error within `SLNGenerator::WriteProjectListings` and can be fixed easily at the same time.

ffulin:
That all sounds reasonable.

One other thing I noticed is that FormatDeterministicProjectGuid() really needs to normalize the slashes to Windows format, so projects/solutions generated on non-Windows machines are also consistent (and don't cause re-loads) when then used on a Windows machine.

arttupeltonen:
OK, there's now a new fix. It keeps the preceding slash and .vcxproj extension in SLNGenerator.cpp, leaving VCXProjectNode.cpp unchanged. I also replace `NATIVE_SLASH` with `BACK_SLASH` in FormatDeterministicProjectGUID like you suggested.

ffulin:
This fix has been integrated into the Perforce mainline and will be in the next release (v0.87).

Thanks again!

PR #58, opened 2015-09-17
Make implib detection case insensitive

inbilla:


ffulin:
I've integrated this fix into the Perforce mainline. It will be in the next release (v0.87).

Thanks for the fix!

PR #60, opened 2015-09-17
Adding LinkerType parameter to Executable and DLL nodes.

inbilla:
This allows the linker type to be explicitly specified. (Useful in cases when the linker does not have an executable name that matches one of the expected profiles) i.e. when using scripts or CMake.


ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.88).

My git-fu was not strong enough to resolve the conflicts via the command line (and GitHubs suggestions just simply don't work) so I have manually integrated the resolved changes back into "dev".

ffulin:
Closing pull request (manually integrated).

PR #61, opened 2015-09-22
Adds -isolateallfiles CL option

i-reuts:
Adds a command line option to perform non-blob compilation with all files isolated.

This option is used to check for missing include statements or forward declarations that can be undetected when using unity files.UnityInputIsolateWritableFiles is not always enough as there are still ways to introduce missing includes/forward declarations.

### DIFF
--- a/Code/Tools/FBuild/FBuildApp/Main.cpp
+++ b/Code/Tools/FBuild/FBuildApp/Main.cpp
@@ -107,6 +107,7 @@ int Main(int argc, char * argv[])
 	bool fixupErrorPaths = false;
 	bool waitMode = false;
 	bool noStopOnError = false;
+	bool isolateAllFiles = false;
 	int32_t numWorkers = -1;
 	WrapperMode wrapperMode( WRAPPER_MODE_NONE );
 	AStackString<> args;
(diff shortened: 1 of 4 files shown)


ffulin:
Hello

I have a pending changelist (that similar to yours) but has not been submitted for the same reason I can't integrate this one. It contains a bug:

If you alternate:
 - normal build
 - isolateallfiles build
 - normal build
 - isolateallfiles build

On the second nounity pass, because nothing needs recompiling, things will not be correctly re-linked. I haven't come up with a clean way to manage this issue (that doesn't cause unnecessary rebuilds), so I've put this feature on the back-burner.

You can safely use it locally as long as you are aware of the issue, or if you use -clean when switching back and forth.

There are also a couple of other subtle issues with your specific implementation:
 - if .UnityInputIsolateWriteableFiles is not set, the command line does nothing
 - if .UnityInputIsolateWriteableFilesLimit is set and -isolateallfiles is used, it will only isolate files upto the limit

I'd like to add this feature: it's useful for the reasons you mentioned, but for the moment this can't be integrated. I'll see if I can find a way to cleanly fix the issue in my change. In that case I'll also take the changelist i've already prepared.

I'm sorry for your wasted effort! :(

If it's cool with you, let's leave this pull request open until the issue is resolved.

i-reuts:
Hello,

Thank you for your feedback. We'll continue to use the feature locally until an official fix is released.

I'm totally fine with using this pull request to track the issue.

Thanks.

ffulin:
I'm marking this issue as blocked because it's still blocked by me doing some internal refactoring.  We'll still have this feature eventually!

ffulin:
A version containing this new option has now been released (v0.96).

PR #62, opened 2015-09-27
Added templates support

gdaow:
Allowing the declaration of "templates" variables with the following syntax :
<pre><code> .TemplateName =
(
 	Stuff here
)
</code></pre>
Wich can be further instanciated :
<pre><code> $TemplateName </code></pre>
This will evaluate the content of the template in a new scope. This allow config factorisation like this :
<pre><code>.Executable =
(
    ForEach( .CompileConfig in .ProjectCompileConfigs )
    {
        Using( .CompileConfig )
        Executable( '$ProjectName$-$Config$-$Platform$' )
        {
            .LinkerOutput = "Output/$Config$/$Platform$/$ProjectBinaryName$.exe"
            ; ...
        }
     }
     ; ....
)
.ProjectName = 'MyProject'
.ProjectBinaryName = 'MyBinaryName'
$Executable

.ProjectName = 'MyProject2'
.ProjectBinaryName = 'MyBinaryName2'
$Executable
</code></pre>
So very much like including files at the template's instanciation places, except using templates doesn't imply knowing their definition place, and it allow more flexible config repartition in .bff files.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFStackFrame.h
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFStackFrame.h
@@ -41,6 +41,10 @@ public:
 									  const Array< const BFFVariable * > & structs,
 									  BFFStackFrame * frame );
 
+	static void SetVarTemplate( const AString & name,
+								const BFFTemplate & tpl,
+								BFFStackFrame * frame );
+
 	// set from an existing variable
 	static void SetVar( const BFFVariable * var, BFFStackFrame * frame );
 
@@ -84,6 +88,9 @@ private:
 
 	// variables at current scope
 	Array< BFFVariable * > m_Variables;
+
+	// Template being instanciated
+	BFFTemplate* m_Template;;
 
 	// pointer to parent scope
 	BFFStackFrame * m_Next;
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFTemplate.h
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFTemplate.h
@@ -1,0 +1,36 @@
+// BFFTemplate - Lazily evalues inner scope
+//------------------------------------------------------------------------------
+#pragma once
+#ifndef FBUILD_BFFTEMPLATE_H
+#define FBUILD_BFFTEMPLATE_H
+
+// Includes
+//------------------------------------------------------------------------------
+#include "Core/Containers/Array.h"
+#include "BFFIterator.h"
+
+// Forward Declarations
+//------------------------------------------------------------------------------
+class AString;
+
+// BFFTemplate
+//------------------------------------------------------------------------------
+class BFFTemplate
+{
+public:
+	BFFTemplate();
+	BFFTemplate(const BFFTemplate& rhs);
+	explicit BFFTemplate(const BFFIterator& defStart, const BFFIterator& defEnd);
+	~BFFTemplate();
+
+	const BFFIterator & DefStart() const { return m_DefStart; }
+	const BFFIterator & DefEnd() const { return m_DefEnd; }
+
+private:
+	BFFIterator m_DefStart;
+	BFFIterator m_DefEnd;
+};
+
+//------------------------------------------------------------------------------
+#endif // FBUILD_BFFTEMPLATE_H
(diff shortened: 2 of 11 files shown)


gdaow:
It's missing unit tests & documentation, but I wanted to get some have your feedback about it before going further.

ffulin:
FASTBuild needs a template/macros system as you've noted. We need to very carefully consider how such a feature would work, since we'll be stuck with the syntax for some time.

Making templates another type of variable allows for some interesting possibilities, but I think it might be better to create a specific template feature. i.e. one where you declare a template using some explicit syntax, something along the lines of:
```C++
#template MyTemplate( .Param1, .Param2 )
(
   /// BFF config here
)
```
and then use:
```C++
#invoke MyTemplate( "thing", "stuff" )
```
I have been favoring this approach because:
 - It can have more explicit error messaging
 - It won't complicate the already complex variant variable system/code
 - it can have explicit params, rather than relying on previously defined variables which can be error prone

Treating it as a variable might allow for some interesting things, like procedurally generated or self-modifying templates, but I'm not sure that's such a good idea :)

Let me know your thoughts.

Code/Tools/FBuild/FBuildCore/BFF/BFFStackFrame.h
```diff
@@ -85,6 +89,9 @@ class BFFStackFrame
 	// variables at current scope
 	Array< BFFVariable * > m_Variables;
 
+	// Template being instanciated
+	BFFTemplate* m_Template;;
```
ffulin: double ;

Code/Tools/FBuild/FBuildCore/BFF/BFFTemplate.h
```diff
@@ -0,0 +1,36 @@
...
+	explicit BFFTemplate(const BFFIterator& defStart, const BFFIterator& defEnd);
+	~BFFTemplate();
+
+	const BFFIterator & DefStart() const { return m_DefStart; }
+	const BFFIterator & DefEnd() const { return m_DefEnd; }
+
+private:
+	BFFIterator m_DefStart;
+	BFFIterator m_DefEnd;
+};
```
ffulin: This is not safe. A variable (and therefore a template) can be declared in an included file, and used after that file has been closed and has fallen out of scope. We'd need to take a copy of the BFF config segement to avoid dereferencing freed memory.

Code/Tools/FBuild/FBuildCore/Error.cpp
```diff
@@ -356,6 +356,21 @@
 	FormatError( iter, 1040u, nullptr, "#undef of built-in token not allowed." );
 }
 
+// Error_1041_TemplateNotFound
+//------------------------------------------------------------------------------
+void Error::Error_1041_TemplateNotFound( const BFFIterator & iter, const AString& variableName)
+{
+
+	FormatError( iter, 1040u, nullptr, "Template %s was not declared.", variableName.Get());
```
ffulin: mismatched error number

Code/Tools/FBuild/FBuildCore/Error.cpp
```diff
@@ -356,6 +356,21 @@
...
+}
+
+// Error_1042_VariableIsNotATemplate
+//------------------------------------------------------------------------------
+void Error::Error_1042_VariableIsNotATemplate( const BFFIterator & iter, const AString& variableName)
+{
+	FormatError( iter, 1040u, nullptr, "%s is not a template", variableName.Get());
```
ffulin: mismatched error number

gdaow:
Totally agree with that. I indeed thought templates as variables would allow funky things, but funky things are sometimes not that a good idea :) I think your syntax is a lot more clear and intuitive, so I have no problem with you closing this pull request, I'll wait for the proper way to do :)

ffulin:
Sounds good. I'll close this as agreed.

Nobody is actively working on a template/macro feature, but it's nearing the top of the list in terms of features that would bring value to FASTBuild, so it's likely something I'd be starting on soon. Given the importance of the feature, and the subtleties likely to be involved, it's probably something I should implement. I'd be happy to discuss the implementation through an issue to lock down a good design though.

PR #63, opened 2015-09-27
Fixs Posix

gdaow:
That's some fixs / implementations for the Posix platforms (SetThreadName, the CachePlugin code, SharedMemory). I tested under Linux, it should work on Mac modulo the build configuration, as I checked the functions I used were Posix-compliant. I'll try to set up a MacOS virtual machine to fix the build on Mac, and I guess I'll have some more fix soon. I push this already so the work isn't made two times :)

### DIFF
--- a/Code/Core/CoreTest/Tests/TestSharedMemory.cpp
+++ b/Code/Core/CoreTest/Tests/TestSharedMemory.cpp
@@ -1,0 +1,94 @@
+// TestSharedMemory.cpp
+//------------------------------------------------------------------------------
+
+// Includes
+//------------------------------------------------------------------------------
+#include "TestFramework/UnitTest.h"
+
+// Core
+#include <Core/Env/Assert.h>
+#include <Core/Process/SharedMemory.h>
+#include <Core/Process/SystemMutex.h>
+#include <Core/Process/Thread.h>
+
+
+#if defined(__LINUX__) || defined(__APPLE__)
+	#include <sys/types.h>
+	#include <wait.h>
+	#include <unistd.h>
+#endif
+
+// TestSharedMemory
+//------------------------------------------------------------------------------
+class TestSharedMemory : public UnitTest
+{
+private:
+	DECLARE_TESTS
+
+	void CreateAccessDestroy() const;
+};
+
+// Register Tests
+//------------------------------------------------------------------------------
+REGISTER_TESTS_BEGIN( TestSharedMemory )
+	REGISTER_TEST( CreateAccessDestroy )
+REGISTER_TESTS_END
+
+// CreateAccessDestroy
+//------------------------------------------------------------------------------
+void TestSharedMemory::CreateAccessDestroy() const
+{
+
+#if defined(__WINDOWS__)
+	// TODO:WINDOWS Test SharedMemory (without fork, so).
+#elif defined(__LINUX__) || defined(__APPLE__)
+	SystemMutex mutex("fbuild_shared_memory_test");
+	bool locked = mutex.TryLock();
+
+	TEST_ASSERT(locked);
+
+	int pid = fork();
+
+	if(pid == 0)
+	{
+		while(!mutex.TryLock())
+		{
+			Thread::Sleep(10);
+		}
+		SharedMemory shm;
+		shm.Open("fbuild_test_shared_memory", sizeof(unsigned int));
+		unsigned int* magic = static_cast<unsigned int*>(shm.GetPtr());
+
+		// Asserts raise an exception when running unit tests : forked process
+		// will not exit cleanly and it will be ASSERTed in the parent process.
+		TEST_ASSERT(magic != nullptr);
+		TEST_ASSERT(*magic == 0xBEEFBEEF);
+		*magic = 0xB0AFB0AF;
+		_exit(0);
+	}
+	else
+	{
+		SharedMemory shm;
+		shm.Create("fbuild_test_shared_memory", sizeof(unsigned int));
+		unsigned int* magic = static_cast<unsigned int*>(shm.GetPtr());
+
+		if(magic != nullptr)
+		{
+			*magic = 0xBEEFBEEF;
+		}
+
+		mutex.Unlock();
+
+		int status;
+
+		TEST_ASSERT(-1 != wait(&status));
+		TEST_ASSERT(WIFEXITED(status) && WEXITSTATUS(status) == 0);
+		TEST_ASSERT(magic != nullptr);
+		TEST_ASSERT(*magic == 0xB0AFB0AF);
+	}
+#else
+	#error Unknown Platform
+#endif
+}
+
+//------------------------------------------------------------------------------
--- a/Code/Core/Process/SharedMemory.cpp
+++ b/Code/Core/Process/SharedMemory.cpp
@@ -6,17 +6,67 @@
 #include "Core/PrecompiledHeader.h"
 
 #include "SharedMemory.h"
+#include "Core/Env/Assert.h"
+#include "Core/Strings/AString.h"
 
 #if defined( __WINDOWS__ )
     #include <windows.h>
+#elif defined(__LINUX__) || defined(__APPLE__)
+	#include <fcntl.h>
+	#include <unistd.h>
+	#include <sys/mman.h>
 #endif
 
+
+#if defined(__LINUX__) || defined(__APPLE__)
+namespace
+{
+void PosixMapMemory(const char* name,
+                    size_t length,
+                    bool create,
+                    int * mapFile,
+                    void ** memory,
+                    AString &portableName)
+{
+		ASSERT(*mapFile == -1);
+		ASSERT(*memory == nullptr);
+
+		// From man shm_open :
+		// For portable use, a shared memory object should be identified by a
+		// name of the form /somename; that is, a null-terminated string of up
+		// to NAME_MAX (i.e., 255) characters consisting  of an initial slash,
+		// followed by one or more characters, none of which are slashes.
+		portableName = "/";
+ 		portableName += name;
+		ASSERT(portableName.FindLast('/') == portableName.Get());
+		ASSERT(portableName.GetLength() <= 255);
+
+		*mapFile = shm_open(portableName.Get(),
+		                    O_RDWR | (create ? O_CREAT : 0),
+		                    S_IWUSR | S_IRUSR);
+		ASSERT(*mapFile != -1);
+
+		if(create)
+		{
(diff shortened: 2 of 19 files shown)


Code/Core/CoreTest/Tests/TestSharedMemory.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+
+#if defined(__WINDOWS__)
+	// TODO:WINDOWS Test SharedMemory (without fork, so).
+#elif defined(__LINUX__) || defined(__APPLE__)
+	SystemMutex mutex("fbuild_shared_memory_test");
+	bool locked = mutex.TryLock();
+
+	(void)locked;
+	ASSERT(locked);
```
ffulin: Use TEST_ASSERT. This is not stripped in debug and will remove the need to use (void)x to avoid unused variable wanings.

Code/Core/CoreTest/Tests/TestSharedMemory.cpp
```diff
@@ -0,0 +1,96 @@
...
+REGISTER_TESTS_BEGIN( TestSharedMemory )
+	REGISTER_TEST( CreateAccessDestroy )
+REGISTER_TESTS_END
+
+// CreateAccessDestroy
+//------------------------------------------------------------------------------
+void TestSharedMemory::CreateAccessDestroy() const
+{
+
+#if defined(__WINDOWS__)
```
ffulin: Ideally we'd unify the tests across platforms by having the test compile and run an executable. Probably ok to leave as is and I can re-factor that later.

Code/Core/Process/SharedMemory.cpp
```diff
@@ -6,17 +6,67 @@
...
     #include <windows.h>
+#elif defined(__LINUX__) || defined(__APPLE__)
+	#include <fcntl.h>
+	#include <unistd.h>
+	#include <sys/mman.h>
 #endif
 
+
+#if defined(__LINUX__) || defined(__APPLE__)
+namespace
```
ffulin: Note that SharedMemory might not be required for OSX and Linux.
It's currently only used by one feature on Windows, which is to pass info between child processes when FASTBuild is run from Visual Studio. We need two processes because of the way VS terminates compilation.

Code/Core/Process/SystemMutex.cpp
```diff
@@ -58,15 +58,16 @@ bool SystemMutex::TryLock()
...
     #elif defined( __LINUX__ ) || defined( __APPLE__ )
         AStackString<> tempFileName;
-        tempFileName.Format( "/var/run/%s.lock", m_Name.Get() );
+        tempFileName.Format( "/tmp/%s.lock", m_Name.Get());
         m_Handle = open( tempFileName.Get(), O_CREAT | O_RDWR, 0666 );
         int rc = flock( m_Handle, LOCK_EX | LOCK_NB );
         if ( rc )
         {
-            if ( errno == EWOULDBLOCK )
+            if ( errno == EAGAIN )
```
ffulin: When I wrote this code, I verified that the return value was EWOULDBLOCK. Did you get EAGAIN when running a second FASTBuild? Should we be checking for both?

Code/Tools/FBuild/FBuildCore/Cache/CachePlugin.cpp
```diff
@@ -64,20 +74,18 @@
...
-void * CachePlugin::GetFunction( const char * friendlyName, const char * mangledName ) const
+void * CachePlugin::GetFunction( const char * name) const
 {
     #if defined( __WINDOWS__ )
         ASSERT( m_DLL );
-        void * func = ::GetProcAddress( (HMODULE)m_DLL, mangledName );
+        void * func = ::GetProcAddress( (HMODULE)m_DLL, name );
         if ( !func )
         {
             FLOG_WARN( "Missing CachePluginDLL function '%s' (Mangled: %s)", friendlyName, mangledName );
```
ffulin: This won't compile anymore on Windows - you deleted friendlyName and mangledName

ffulin: This won't compile anymore on Windows, since it references friendlyName and mangledName

Code/Tools/FBuild/FBuildCore/Cache/CachePlugin.cpp
```diff
@@ -42,18 +47,23 @@
             return;
         }
 
-        m_InitFunc		= (CacheInitFunc)		GetFunction( "CacheInit",		"?CacheInit@@YA_NPEBD@Z" );
-        m_ShutdownFunc	= (CacheShutdownFunc)	GetFunction( "CacheShutdown",	"?CacheShutdown@@YAXXZ"  );
```
ffulin: These mangled names on Windows ensure we get the correct function signature.

Code/Tools/FBuild/FBuildCore/Cache/CachePluginInterface.h
```diff
@@ -4,12 +4,21 @@
...
+	#error Unknown Platform
+#endif
+
 // DLL Export
 //------------------------------------------------------------------------------
 // Externally compiled DLL should define this appropriately
 // e.g.
 //#define CACHEPLUGIN_DLL_EXPORT __declspec(dllexport)
 
+extern "C" {
```
ffulin: This breaks compatibility of the interface for existing cache plugin users.

Code/fbuild.bff
```diff
@@ -660,6 +660,7 @@ Alias( 'All+Tests' )
 	.Targets		= { 'All', 'Tests' }
 }
 
+#if __WINDOWS__
```
ffulin: This shouldn't be necessary. The latest version (v0.87) supports VS project and solution generation on all platforms.

ffulin:
Thanks for the pull request - one couple of steps closer to feature parity!

I added some notes to various files.  It's mostly good to go, but there are a couple of issues we'd need to resolve before integrating this.

Code/Core/Process/SystemMutex.cpp
gdaow: You're rigth, I misread some documentation. The good way seems to be checking both :
From http://www.gnu.org/savannah-checkouts/gnu/libc/manual/html_node/Error-Codes.html : 
Portability Note: In many older Unix systems, this condition was indicated by EWOULDBLOCK, which was a distinct error code different from EAGAIN. To make your program portable, you should check for both codes and treat them the same.

gdaow:
Nice, thanks a lot for the review, I'll get all that fixed. About the mangled names, how should it be handled ? As the mangling is compiler-specific, it seems dangerous to me to hard-code it. (Although I think GCC and Clang mangle the same way).

ffulin:
I looked into it a little bit, and it seems switching these functions to extern "C" is probably the only way to go long-term. As you pointed out, the mangling is compiler-specific, and in-fact, it's not even consistent between x86 and x64 (there is already a bug in that the plugin dll only work for x64 even on Windows).

So if we change the functions to be extern C as you proposed, it fixes the problem going forward, but leaves us with the backwards compat issue. I think we can solve that by making the code try the mangled names on x64 Windows only first, and if that fails, fall back to the unmangled names. I'll need to try that to make sure it's going to work, but I think it should be ok.

For your change, perhaps just implement the OS X/Linux part (using the friendlyName and ignoring the mangledName) and leave the Windows as-is. Once it's integrated I'll have a crack at sorting out the Windows bit afterwards.

gdaow:
I fixed the things you pointed out. For the mangled name, it seems a good idea to me, I added a default parameter for mangledName, ignored it on Posix, and leaved the Windows code unchanged. Note that backward compatibility will be broken if recompiling plugins, as 'extern "C"' should be added in the plugins code as well. Let me know if there is still things to improve.

Code/Tools/FBuild/FBuildCore/Cache/CachePlugin.cpp
```diff
@@ -41,19 +46,29 @@
...
-        ASSERT( false ); // TODO:LINUX Implement CachePlugin
+
+    #elif defined( __APPLE__ ) || defined( __LINUX__ )
+        m_DLL = dlopen(dllName.Get(), RTLD_NOW);
+        if ( !m_DLL )
+        {
+            FLOG_WARN( "Cache plugin '%s' load failed (0x%x).", dllName.Get(), dlerror() );
+            return;
+        }
+        dlerror();
```
ffulin: this dlerror() seems left over from some debugging?

Code/Tools/FBuild/FBuildCore/Cache/CachePlugin.cpp
```diff
@@ -64,20 +79,18 @@
 
 // GetFunction
 //------------------------------------------------------------------------------
-void * CachePlugin::GetFunction( const char * friendlyName, const char * mangledName ) const
+void * CachePlugin::GetFunction( const char * friendlyName, const char* mangledName) const
 {
     #if defined( __WINDOWS__ )
         ASSERT( m_DLL );
-        void * func = ::GetProcAddress( (HMODULE)m_DLL, mangledName );
+        void * func = ::GetProcAddress( (HMODULE)m_DLL, name );
```
ffulin: This should read mangledName for now.

poppolopoppo: Build is broken on dev because of this :P
````
Code\Tools\FBuild\FBuildCore\Cache\CachePlugin.cpp(86) : error C2065: 'name' : undeclared identifier
````

ffulin: I'm in the process of fixing it - I'll integrate a new working version soon.

poppolopoppo: Ok thanks

ffulin: Should be fixed now.

ffulin:
I've integrated this into the Perforce mainline. These changes will be in the next release (v0.88).

I've submitted a follow-up fix to "dev" which resolved some issues with this change on Windows.

In a future revision, I'll look into breaking the dependency on mangled names in Windows.

ffulin:
As a follow up to this, I've intergrated some work-in-progress changes from the Perforce mainline related to completing Linux/OS X support:
 - distributed compilation tests fixed/enabled
 - deleted unused code we don't need to port
 - made the FBuildWorker function as a command line application on OS X/Linux

PR #65, opened 2015-10-15
Warnings emitted from Visual Studio's Compiler "cl.exe" are not…

matlec:
…dumped (unless they are treated as an error)

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -244,6 +244,12 @@ ObjectNode::~ObjectNode()
 		return NODE_RESULT_FAILED; // SpawnCompiler has logged error
 	}
 
+	// output any warnings
+	if( ContainsWarningsMSCL( ch.GetOut().Get(), ch.GetOutSize() ) )
+	{
+		DumpOutput( job, ch.GetOut().Get(), ch.GetOutSize(), GetName() );
+	}
+
 	const char *output = nullptr;
 	uint32_t outputSize = 0;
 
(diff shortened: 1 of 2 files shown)


ffulin:
HI there.  Thanks for the pull request.

Looking at the code, it seems like when not using the cache or distributed compilation (where FASTBuild uses /showIncludes to extract includes) the output which would be dumped if there is a warning will contain all the /showInclude info.

matlec:
Hey there!

Thanks for the review of my pull request. I just verified that ObjectNode::DumpOutput() filters the output for "#line" information, so only lines that actually contain warnings are dumped to the output. So the attached code should work as described.

ffulin:
In the /showincludes case, it's actually "note: including file:" that needs replacing, which that function also does. This means the code is ok as you've submitted it, but sadly only for English locales. (MSVC localizes the "note: including file" bit depending on the user locale :( )

I'll look into making that filtering more reliable, so I'll hold off on integrating this for the moment.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -821,6 +827,44 @@ const char * ObjectNode::GetObjExtension() const
...
+			if( ( *lineEnd == '\r' ) || ( *lineEnd == '\n' ) )
+			{
+				break;
+			}
+			lineEnd++;
+		}
+
+		if( ( lineEnd - lineStart ) >= sizeof( warningString ) - 1 )
+		{
+			containsWarnings = containsWarnings || ( strstr( lineStart, warningString ) != nullptr );
```
ffulin: We could early out here which would also remove the need for the containsWarnings variable.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -821,6 +827,44 @@ const char * ObjectNode::GetObjExtension() const
...
+{
+	ASSERT( IsMSVC() );
+	bool containsWarnings = false;
+	static const char warningString[] = "warning";
+
+	if( ( data == nullptr ) || ( dataSize == 0 ) )
+	{
+		return containsWarnings;
+	}
+
```
ffulin: We could take advantage of the fact that the buffer returned by Process is null terminated so we could replace most of this logic with:
```C++
return ( strstr(data, "\nwarning") != nullptr );
```
I need to do some tests to see if this string is localized though.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -821,6 +827,44 @@ const char * ObjectNode::GetObjExtension() const
 	return m_ObjExtensionOverride.Get();
 }
 
+// ContainsWarningsMSCL
+//------------------------------------------------------------------------------
+bool ObjectNode::ContainsWarningsMSCL(const char * data, uint32_t dataSize) const
+{
```
ffulin: If we extend DetermineFlags to check for /WX (warnings as errors) we could early out of this function:
```C++
// If /WX is enabled, compilation would have failed, so we can never have warnings
if ( GetFlag( FLAG_MSVC_WARNINGS_AS_ERRORS ) )
{
   return false; // cannot contain errors
}
```

ffulin:
I've been thinking about how to reduce to overhead of this check. I've added a couple of notes about that to the pull request. Would you mind taking a look please and let me know what you think?

matlec:
I agree with your suggestions, they can further reduce the overhead introduced by this bugfix. Will you incorporate your suggestions?

ffulin:
I will take your pull request and make those additional changes.

matlec:
Great, thank you!

ffulin:
I integrated this, but then realized the pull request was against the main branch, so I've had to back it out. Since this is a relatively small change, I'll manually re-create the changes against the "dev" branch.

ffulin:
I've integrated this change into the Perforce mainline. It will be in the next release (v0.90). I also made the changes/optimizations we talked about and added unit tests.

Thanks for your change and sorry it took so long to integrate.

matlec:
Thank you again! And sorry for picking the wrong branch, I will do better next time :-)

PR #69, opened 2015-11-01
Fuzzy string matching for unknown targets

poppolopoppo:
When a target is unknown FBuild will use a simple Levenshtein distance to find closely matching targets. Nothing will be executed, but a list of suggestions will be displayed.
Ex:
````
$ FBuild.exe FbuldApp-x64
Unknown build target 'FbuldApp-x64'
Did you mean 'FBuildApp-X64' ?
        - or 'FBuildApp-X86' ?
````

I tuned the filter to display at max 5 suggestions with a distance to user input inferior or equal to 3.
Levenshtein distance was implemented in String::EditDistanceI(), the code was taken from [wikibooks](https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Levenshtein_distance#C).
This behavior was inspired by git command line tool.

I can extend it to unknown options before merge if you like it.

### CODE REVIEW

File: Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp [hunk 1]
```diff
@@ -1436,6 +1436,77 @@ Node * NodeGraph::FindNodeInternal( const AString & fullPath ) const
...
+	ASSERT( capacity <= 8 );
+	uint32_t nodesDistance[8] = {0};
+
+	size_t count = 0;
+	for ( size_t i = 0 ; i < NODEMAP_TABLE_SIZE ; i++ )
+	{
+		if ( nullptr == m_NodeMap[i] )
+			continue;
+
+		Node * const n = m_NodeMap[i];
```
ffulin: The m_NodeMap is a table of linked lists (nodes with colliding hashes are chained together). You need to iterate the linked list to make sure you get all the nodes.

Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -1436,6 +1436,77 @@ Node * NodeGraph::FindNodeInternal( const AString & fullPath ) const
...
+				}
+			}
+
+			nodes[pos] = n;
+			nodesDistance[pos] = d;
+
+			if ( count < capacity )
+				count++;
+
+			if ( count < 1 )
```
ffulin: seems like this if/else should read:
```C++
ASSERT( count >= 1 );
worstMinDistance = nodesDistance[count - 1];
````

File: Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp [hunk 3]
```diff
@@ -1436,6 +1436,77 @@ Node * NodeGraph::FindNodeInternal( const AString & fullPath ) const
 	return nullptr;
 }
 
+// FindNearestNodesInternal
+//------------------------------------------------------------------------------
+size_t NodeGraph::FindNearestNodesInternal( const AString & fullPath, const Node **nodes, uint32_t capacity ) const
```
ffulin: Question: Is there a reason you don't use an Array< Node * > for this? It would track capacity and size for you as well as providing assertions for indexing.

Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -1436,6 +1436,77 @@ Node * NodeGraph::FindNodeInternal( const AString & fullPath ) const
...
+	ASSERT( nodes );
+	ASSERT( capacity );
+
+	if ( fullPath.IsEmpty() )
+		return 0;
+
+	const uint32_t maxDistance = 5;
+	uint32_t worstMinDistance = fullPath.GetLength() + 1;
+
+	ASSERT( capacity <= 8 );
```
ffulin: You could also use an Array<> here, presized to the capacity of the input vector.

ffulin:
This change is very nice!  I've made some comments (one bug and a few suggestions/questions) on the change itself, but also I wonder if you might be able to make one or two high level changes?

**Make a new .h/,cpp file**
These functions are very specific, so I don't think they should go in the base AString class. Could you please move these functions into a new Strings/LevenshteinDistance.cpp/.h file. Something like this maybe:
```C++
class LevenshteinDistance
{
public:
     static uint32_t Calculate( const char * str );
     static uint32_t Calculate( const AString & str );
```

**Unit Tests**
It would be great if the core Levenshtein Distance stuff could have a unit created in Core/CoreTest/Tests/LevenshteinDistance.cpp. Mostly to catch the edge cases. I can help you get the tests working locally if you run into any problems - the CoreTests shouldn't be problematic since they don't depend on anything (like the way the FBuildTest depends on compilers).

poppolopoppo:
I fixed the bug and changed the code according to your remarks.
I added unit tests for the new LevenshteinDistance class. CoreTests worked without problem as you predicted :)

I also made a slight change of behavior : it will now display the 5 closest targets, without a maximum distance, but the target must share at least one character with user input. I found it more useful like this, but tell me if you have a better idea.

Ex:
````
$ FBuild.exe FbuldApp-x64
Unknown build target 'FbuldApp-x64'
Did you mean one of these ?
    FBuildApp-X64
    FBuildApp-x64OSX
    FBuildApp-x64Linux
    FBuildApp-x64-Debug
    FBuildApp-x64-Profile
````
````
$ FBuild.exe FbuldApp-x86
Unknown build target 'FbuldApp-x86'
Did you mean one of these ?
    FBuildApp-X86
    FBuildApp-X86Clang
    FBuildApp-x86-Debug
    FBuildApp-x86-Profile
    FBuildApp-x86-Release
````

ffulin:
I took a look at the updated pull request and it all looks good. Thanks for making those changes and thanks for adding the test! I'll try to get that integrated soon.

ffulin:
I have integrated this into the Perforce mainline. This will be in the next release (v0.89)

PR #79, opened 2015-12-19
added -targets flag to list all available build targets

cyberkm:
Show list of available build targets when -targets flag added

### DIFF
--- a/Code/Tools/FBuild/FBuildApp/Main.cpp
+++ b/Code/Tools/FBuild/FBuildApp/Main.cpp
@@ -6,6 +6,8 @@
 #include "Tools/FBuild/FBuildCore/FBuild.h"
 #include "Tools/FBuild/FBuildCore/FBuildVersion.h"
 #include "Tools/FBuild/FBuildCore/FLog.h"
+#include "Tools/FBuild/FBuildCore/Graph/NodeGraph.h"
+#include "Tools/FBuild/FBuildCore/Graph/Node.h"
 
 #include "Core/Containers/Array.h"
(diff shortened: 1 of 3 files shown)


Code/Tools/FBuild/FBuildApp/Main.cpp
```diff
@@ -604,3 +620,20 @@ int WrapperIntermediateProcess( const AString & args, const FBuildOptions & opti
 }
 
 //------------------------------------------------------------------------------
+// DisplayTargetsList
```
ffulin: Each entry in the hash table is a linked list. You'll need to walk the chain (m_Next) or you won't find all the nodes when there are collisions.
You'll likely need to move this function into the NodeGraph in order to be able to access the m_Next member of Node.

Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -523,6 +523,13 @@ Node * NodeGraph::GetNodeByIndex( uint32_t index ) const
 	return n;
 }
```
ffulin: If you move the DisplayTargetList function into NodeGraph, you won't need to expose this.

ffulin:
This change looks good. I'll integrate it soon.

ffulin:
I've integrated this change into the P4 mainline. It will be in the next release (v0.89).

I made small change during integration: The cmdline option is now "-showtargets" instead of "-targets".

cyberkm:
Fine by me, thanx :)

Issue #80, opened 2015-12-24
Add the ability to add #define from command line to control the build


shooshx:
Currently the only useful #defines are __WINDOWS__, __LINUX__, __OSX__ and anything that derives directly from those. It is not possible to add additional custom conditional compilation.
some use cases:
- "RELEASE_BUILD" - if I don't want to create separate structs for debug and release to avoid complexity 
- "USE_SIMD_INSTRUCTIONS"  - enables just a single compilation switch
- "WITH_GLES2_CG_SUPPORT"  - add a "-DGLES2_CG_SUPPORT" flag to all compilation
The only way to do these things today is to add more and more configurations in an ever expanding matrix of options
I propose adding a command line option "-define NAME" that would add an implicit #define to the .bff parsing.
Attached is a patch that does this.

[define_from_cmd.diff.txt](https://github.com/fastbuild/fastbuild/files/71805/define_from_cmd.diff.txt)

ffulin:
Hello

When I designed FASTBuild, I wanted to make sure that you could build multiple configurations in parallel. It's really nice to be able to build all configs (Debug & Release for example) in parallel before submitting changes (with "fbuild all" for example). A make-like command line support for defines is counter to this goal.

I think you really want to declare your permanent configs in the bff, but I can see how if you have many transient options creating a full matrix is pretty painful.

With that in mind, I see two options:

**1) In the bff**
 - declare the permanent configs (like release and debug)
 - group your temporary settings (those that aren't automatically derived from the above) together and declare them with #define and enable disable them as required:
```C++
//#define USE_SIMD_INSTRUCTIONS
#define WITH_GLES_CG_SUPPORT
```
**Pro:** Switching between Debug and Release does't cause unnecessary rebuilding
**Pro:** No changes required to FASTBuild
**Con:** Local edits to bff may be undesireable

**2) Add command line -define support**
Your patch as it stands isn't safe, because it doesn't cause the bff config to be re-evaluated when you change the defines. This means config based on those defines isn't updated and things are not rebuilt when needed. So switching between build configs won't correctly rebuild for example.
This can be fixed by extending your change to record the name and values of the defines and re-parsing if those change between builds. FASTBuild already does this for [#import ](http://fastbuild.org/docs/syntaxguide.html#import) directive, so that code could be extended to track these defines in the same way.
**Pro:** Less complicated bff / aligns with your expectations
**Con:** Rebuilds whenever you change anything
**Con:** Can't parallelize multiple configs
[...]

PR #83, opened 2016-01-24
add -define command line to control building and saved given defines to database

shooshx:
see discussion in https://github.com/fastbuild/fastbuild/issues/80
add a command line option -define
save and load the given defines to the serialized database
if there's a change (addition or omission), cause a reparsing of the .bff file, similar to what happens when an imported environment variable is changed

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
@@ -9,6 +9,7 @@
 
 #include "Tools/FBuild/FBuildCore/BFF/BFFParser.h"
 #include "Tools/FBuild/FBuildCore/BFF/Functions/FunctionSettings.h"
+#include "Tools/FBuild/FBuildCore/BFF/BFFMacros.h"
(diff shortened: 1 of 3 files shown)


Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -423,6 +461,18 @@ void NodeGraph::Save( IOStream & stream ) const
...
 
+		// externally defined variables
+		const Array< AString > & definedStrings = BFFMacros::Get().Tokens();
+		const uint32_t definedStringsSize = static_cast<uint32_t>( definedStrings.GetSize() );
+		ASSERT( definedStringsSize == definedStrings.GetSize() );
+		stream.Write( definedStringsSize );
+		for ( uint32_t i = 0; i < definedStringsSize; ++i )
+		{
+			const AString & def = definedStrings[i];
+			stream.Write(def);
```
ffulin: It seems like this will cause all tokens from the previous run to be saved, including those that were defined in the config file itself. This will cause a permanent re-parsing situation.

PR #84, opened 2016-02-10
Adds parameter .CompilerInputFilesRoot

SergeyMakeev:
We use custom build system to generate FastBuild .bff files.
Our build system generates list of files to compile using .CompilerInputFiles option.

If we include multiple files with the same name but in different directories using .CompilerInputFiles, we got the following error:

Conflicting objects found:
 File A: a.cpp
 File B: test\a.cpp

This error does not occur if we adding files using .CompilerInputPath

To fix this error new parameter .CompilerInputFilesRoot was added to Library function.
This parameter must be used with .CompilerInputFiles to generate the correct .obj file names if we have multiple files with the same name but in different directories.
I generate the subPath string using same way as subPath is generated when dirNode is known.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
@@ -470,15 +470,8 @@ bool FunctionObjectList::GetInputs( const BFFIterator & iter, Dependencies & inp
 		inputs.Append( dirNodes );
 	}
 
-	const char * rootDirectory = nullptr;
-	const BFFVariable * compilerInputFilesRoot = nullptr;
-	if ( GetString( iter, compilerInputFilesRoot, ".CompilerInputFilesRoot", true ) )
-	{
-		rootDirectory = compilerInputFilesRoot->GetString().Get();
-	}
-
 	// do we want to build a specific list of files?
-	if ( !GetNodeList( iter, ".CompilerInputFiles", inputs, false, rootDirectory ) )
+	if ( !GetNodeList( iter, ".CompilerInputFiles", inputs, false ) )
 	{
 		// helper will emit error
 		return false;
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
@@ -470,8 +470,15 @@ bool FunctionObjectList::GetInputs( const BFFIterator & iter, Dependencies & inp
 		inputs.Append( dirNodes );
 	}
 
+	const char * rootDirectory = nullptr;
+	const BFFVariable * compilerInputFilesRoot = nullptr;
+	if ( GetString( iter, compilerInputFilesRoot, ".CompilerInputFilesRoot", true ) )
+	{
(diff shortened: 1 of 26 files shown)


ffulin:
Hello. Thanks for the pull request. 

Unfortunately, there are a few problems with this changelist as-is:
 ** 1) Doesn't work if file used in multiple projects** Since the FileNode is created with the basePath set for the very first use, it can only be used correctly in one project.  i.e. if Folder/File.cpp is used with a base path of Folder/ there is no possibility to have this file with an alternate basePath if used in multiple projects.
 ** 2) Adds a non-trivial overhead to all file nodes** The additional AString member in FileNode adds a nont-trivial memory, creation and serialization cost
 ** 3) BasePath is not checked** The basePath is used to generate the sub-path without verifying if the file is within that path. This could result in malformed paths or crashes. You have some asserts in the code in relation to this, but an end user will never see this as asserts are stripped in Release builds. (Asserts are used to check/enforce code logic rather than user inputs)
 ** 4) Missing error check on new parameter** When you retrieve the .CompilerInputFilesRoot property you don't correctly handle errors (if the user tries to set it to the wrong type for example).

I think the right way to solve this problem would be for the ObjectListNode to store the base path instead of every FileNode and apply the base path logic you have to loose files only when they begin with the path.  This would resolve issues 1), 2) and 3) as well as greatly reducing the number of files to modify in this pull request.  (Issue 4 is just a minor bug that can be resolved by returning false if GetString returns false in FunctionObjectList)

I hope that makes sense - let me know! :)

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
```diff
@@ -470,8 +470,15 @@ bool FunctionObjectList::GetInputs( const BFFIterator & iter, Dependencies & inp
 		inputs.Append( dirNodes );
 	}
 
+	const char * rootDirectory = nullptr;
+	const BFFVariable * compilerInputFilesRoot = nullptr;
+	if ( GetString( iter, compilerInputFilesRoot, ".CompilerInputFilesRoot", true ) )
+	{
```
ffulin: We need to return false here if GetString returns false as that indicates a parsing error has occurred.

Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
```diff
@@ -353,7 +353,20 @@ bool ObjectListNode::CreateDynamicObjectNode( Node * inputFile, const DirectoryL
...
-    }
+	}
+	else
+	{
+		// if source comes from a specific list of files, use path relative to file base directory
+		if ( inputFile->GetType() == Node::FILE_NODE )
+		{
+			FileNode * fileNode = inputFile->CastTo< FileNode >();
+			const AString & baseDir = fileNode->GetBaseDirectory();
+			if (baseDir.GetLength() > 0)
```
ffulin: This needs to check if fileName begins with the baseDir (using PathUtils::PathBeginsWith) otherwise you could end up with a corrupt filename, or worse, a crash (if baseDir was longer than fileName)

SergeyMakeev:
Thanks for the your comments.
The proposed method using ObjectListNode is much better.
I will rewrote code in such way and sent new pull request to you.

PR #85, opened 2016-02-11
Adds .CompilerInputFilesRoot parameter to Library and ObjectList functions

SergeyMakeev:
.CompilerInputFilesRoot implementation was rewritten

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
@@ -212,7 +212,11 @@ FunctionObjectList::FunctionObjectList()
 		return false;
 	}
 
-	const char * baseDirectory = GetBaseDirectory( funcStartIter );
+	AStackString<> baseDirectory;
+	if ( !GetBaseDirectory( funcStartIter, baseDirectory ) )
+	{
+		return false; // GetBaseDirectory will have emitted error
+	}
 
 	// Create library node which depends on the single file or list
 	ObjectListNode * o = ng.CreateObjectListNode( m_AliasForFunction,
@@ -486,23 +490,23 @@ bool FunctionObjectList::GetInputs( const BFFIterator & iter, Dependencies & inp
 
 // GetBaseDirectory
 //------------------------------------------------------------------------------
-const char * FunctionObjectList::GetBaseDirectory( const BFFIterator & iter ) const
+bool FunctionObjectList::GetBaseDirectory( const BFFIterator & iter, AStackString<> & baseDirectory) const
 {
-	const char * baseDirectory = nullptr;
-	const BFFVariable * compilerInputFilesRoot = nullptr;
-	if ( !GetString( iter, compilerInputFilesRoot, ".CompilerInputFilesRoot", true ) )
+	AStackString<> baseDir;
+	if ( !GetString( iter, baseDir, ".CompilerInputFilesRoot", false ) ) // false = optional
 	{
-		return nullptr;
+		return false; // GetString will have emitted error
+	}
+	if ( !baseDir.IsEmpty() )
+	{
+		NodeGraph::CleanPath( baseDir, baseDirectory );
 	}
 	else
 	{
-		if ( compilerInputFilesRoot != nullptr )
-		{
-			return compilerInputFilesRoot->GetString().Get();
-		}
+		baseDirectory.Clear();
 	}
 
-	return "";
+	return true;
 }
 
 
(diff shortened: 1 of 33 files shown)


Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
```diff
@@ -353,7 +360,15 @@ bool ObjectListNode::CreateDynamicObjectNode( Node * inputFile, const DirectoryL
             // ... use everything after that
             subPath.Assign(fileName.Get() + baseDir.GetLength(), lastSlash ); // includes last slash
         }
-    }
+	}
+	else
+	{
+		if ( !m_BaseDirectory.IsEmpty() && fileName.BeginsWith( m_BaseDirectory ) )
```
ffulin: Need to do this:
```C++
if ( !m_BaseDirectory.IsEmpty() && PathUtils::PathBeginsWith( fileName, m_BaseDirectory ) )
```
to correctly handle case-sensitivity on all platforms. (Currently this is incorrectly case sensitive on Windows and OS X)

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
```diff
@@ -212,6 +212,8 @@ FunctionObjectList::FunctionObjectList()
 		return false;
 	}
 
+	const char * baseDirectory = GetBaseDirectory( funcStartIter );
```
ffulin: There are a few problems here:
**a) Not "cleaning" the path.** This will work currently only if you provide exactly the right full path. Cleaning the path takes care of canonicalizing the path.
**b) Not checking for errors.** If the bff has a syntax error, we need to return instead of continuing parsing.
**c) .CompilerInputFiles is not optional.** We want this to be opt-in, since it's only when listing files explicitly that this is needed.
We need something like this:
```C++
AStackString<> baseDir, baseDirClean;
if ( !GetString( funcStartIter, baseDir, ".CompilerInputFilesRoot", false ) ) // false = optional
{
    return false; // GetString will have emitted error
}
if ( !baseDir.IsEmpty() )
{
    NodeGraph::CleanPath( baseDir, baseDirClean );
}
```
Then you can always pass "baseDirClean" through to CreateLibraryNode.

ffulin:
Thanks for the updated pull request.  We're super-close now.  I added a couple of comments to the pull request for some small changes that are needed to make this robust.

PS - Sorry that this is so finicky. I'm working on simplifying that by converting all params to the reflection system, but haven't gotten to ObjectList() yet. For converted classes, adding an option like this is one line:
```C++
	REFLECT( m_BaseDir,				"CompilerInputFilesRoot",		MetaOptional() + MetaPath() )
```
which will take care of everything (error handling, path canonicalization etc).

SergeyMakeev:
Thank you for review my push request.
I was fixed all issues according to your comments.

It will be great to use the new syntax provided by reflection system. :)

SergeyMakeev:
I found another issue with the source files having the same name.
If I use a distributed compilation and two source files have the same name, for example:

```
Test1\a.cpp
Test2\a.cpp
```
In this case, the remote worker creates a temporary file to be compiled, using the following path:

`C:\Users\UserName\AppData\Local\Temp\.fbuild.tmp\0x00000000\core_1001\a.cpp`

Both "Test1\a.cpp" and "Test2\a.cpp" will be placed in temporary file with same name.
This is usually is not a problem, but not in the case if both files uses anonymous namespaces.

For example, if a both files contains next lines:

```
namespace
{
  void TestFunc()
  {
  }
}
```
We will get linker error while linking "Test1\a.obj" and "Test2\a.obj"

The error arises from the fact that Visual Studio compiler generates a unique identifier for an anonymous namespace using the full path to the source file. And in our case, the both source files has same path, which leads to the generation of the same identifier for anonymous namespace.
As a result namespaces in different translation units generates the same identifier and in fact are no longer anonymous. Linker sees two identical functions in different .obj files that is causing linking error.

Also the both files have a same path to source file inside PDB file that are bad for the debugging purposes.

I rewrote the code for generating a temporary file on the remote worker.
Comments are welcome :)

ffulin:
**>  Visual Studio compiler generates a unique identifier for an anonymous namespace using the full path to the source file.**

That's a little troubling. I will try to reproduce this. Could you please extract the additional changes you've made into a separate pull request? This will make things much easier to manage. I'll add some comments to that new pull request about the details of that change.

**> Also the both files have a same path to source file inside PDB file that are bad for the debugging purposes.**

On the remote end, we compile preprocessed output which includes #line directives that control the debug information paths, so this should not be a problem.

SergeyMakeev:
> Could you please extract the additional changes you've made into a separate pull request?

No problem, I was revert my last changes from this push request.

ffulin:
I've integrated this change into the Perforce mainline and updated the documentation for the new option.  This will be in the next release (v0.90).  Thanks!

PR #89, opened 2016-02-20
Fixed linker issue with anonymous namespaces and distributed compilation

SergeyMakeev:
Move this commit from the other push request.

I found another issue with the source files having the same name.
If I use a distributed compilation and two source files have the same name, for example:

```
Test1\a.cpp
Test2\a.cpp
```
In this case, the remote worker creates a temporary file to be compiled, using the following path:

`C:\Users\UserName\AppData\Local\Temp\.fbuild.tmp\0x00000000\core_1001\a.cpp`

Both "Test1\a.cpp" and "Test2\a.cpp" will be placed in temporary file with same name.
This is usually is not a problem, but not in the case if both files uses anonymous namespaces.

For example, if a both files contains next lines:

```
namespace
{
  void TestFunc()
  {
  }
}
```
We will get linker error while linking "Test1\a.obj" and "Test2\a.obj"

The error arises from the fact that Visual Studio compiler generates a unique identifier for an anonymous namespace using the full path to the source file. And in our case, the both source files has same path, which leads to the generation of the same identifier for anonymous namespace.
As a result namespaces in different translation units generates the same identifier and in fact are no longer anonymous. Linker sees two identical functions in different .obj files that is causing linking error.

Also the both files have a same path to source file inside PDB file that are bad for the debugging purposes.

I rewrote the code for generating a temporary file on the remote worker.
Comments are welcome :)

### DIFF
--- a/Code/Core/Strings/AString.cpp
+++ b/Code/Core/Strings/AString.cpp
@@ -646,6 +646,25 @@ const char * AString::FindI( const char * subString ) const
 	return nullptr;
 }
 
+
+// FindFirst
+//------------------------------------------------------------------------------
+const char * AString::FindFirst( char c ) const
+{
+	const char * pos = m_Contents;
+	const char * end = pos + m_Length - 1;
+	while ( pos <= end )
+	{
+		if ( *pos == c )
+		{
+			return pos;
+		}
+		pos++;
+	}
+	return nullptr;
+}
+
+
 // FindLast
 //------------------------------------------------------------------------------
 const char * AString::FindLast( char c ) const
--- a/Code/Core/Strings/AString.cpp
+++ b/Code/Core/Strings/AString.cpp
@@ -646,25 +646,6 @@ const char * AString::FindI( const char * subString ) const
 	return nullptr;
 }
 
-
-// FindFirst
-//------------------------------------------------------------------------------
-const char * AString::FindFirst( char c ) const
-{
-	const char * pos = m_Contents;
-	const char * end = pos + m_Length - 1;
-	while ( pos <= end )
-	{
-		if ( *pos == c )
-		{
-			return pos;
-		}
-		pos++;
-	}
-	return nullptr;
-}
-
-
 // FindLast
 //------------------------------------------------------------------------------
 const char * AString::FindLast( char c ) const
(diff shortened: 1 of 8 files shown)


ffulin:
I've reproduced this locally. I never noticed this before since I encourage all users to use Unity files which mostly forces you to uniquify your conflicting anonymous functions anyway. I actually think it feels like this is a compiler bug, since I would have expected it to use the #line directives in the function name generation as it does for the debug info. Oh well :)

Thinking about the specifics of your fix, I wonder if instead of re-creating the partial directory structure, we should hash the path instead and use that as part of the file name. i.e. something like:
```C++
FileName_0x56F73452.obj 
```
or alternatively, if we want the obj name to be "clean", just make one directory like:
```C++
0x56F73452/FileName.obj 
```

This would be faster and would also avoid path length limit problems if the file is really deep in a hierarchy of folder. What do you think?

Code/Core/Strings/AString.cpp
```diff
@@ -646,6 +646,25 @@ const char * AString::FindI( const char * subString ) const
 	return nullptr;
 }
 
+
+// FindFirst
+//------------------------------------------------------------------------------
```
ffulin: AString already has a a Find() function, so you don't need this.

ffulin:
Thanks for the additional fix btw!

SergeyMakeev:
> What do you think?
It is a good idea, I will rewrite code and will send update.

ffulin:
Nice.  I think the "0x56F73452/FileName.obj" is probably the safest way to go (so it'll create/cleanup one dir each time, which should be fairly low overhead, and the obj name is kept nice).

Thanks for sorting this out :)

SergeyMakeev:
I fixed the code according to your comments. Now my changes are much more elegant and local :)

kelleyma49:
Thank you for submitting a fix for this, @SergeyMakeev .  I'm seeing the same issue.  It looks like Microsoft labels this as "by design". [See this issue for more info](https://connect.microsoft.com/VisualStudio/feedback/details/1260338).

kelleyma49:
FYI, changing the TEMP directory to a unique location on the remote builders seems to workaround this problem.

ffulin:
I've integrated this into the Perforce mainline. This fix will be in the next release (v0.90). I've also created a regression test.

Thanks again for the fix!

Issue #55, opened 2015-09-04
How to prevent from rebuilding everything?


psallandre:
Hi,
I have a main fbuild.bff that includes other ones:
#include "lib1.bff"
#include "lib2.bff"
...
#include "exe.bff"
All is good but there are linker errors, so I modify linker options in "exe.bff" but it tells me "BFF file 'exe.bff' has changed (reparsing will occur)" and rebuilds everything including all libs.
I use '-cache' on the commandline but it doesn't seem to have any impact.
Thx

ffulin:
Hello

Modifying any .bff currently causes a full rebuild. The expectation is that once your bff is setup, it rarely needs to change. However, as you've found, when setting up for the first time this can be annoying.

In the long term, I plan to change this behaviour, so only affected items will rebuild when a bff is modified. This requires some non-trivial re-factoring which is already underway, but won't be complete for some time.

In the short term, as you've tried, the -cache option can be used to try to reduce the rebuild time to lessen the impact of this problem. The main reason this wouldn't work for you is if you are using an incompatible compiler or compiler option. Please take a look at the caching restrictions documented here: http://www.fastbuild.org/docs/features/caching.html

If you still can't get caching working, let me know.

psallandre:
Thanks for the explanation. I'm looking forward for this functionality! 

As a starting point, maybe a "per file" modify check would be a good improvement. In my example above, a change in "exe.bff" would only trigger "exe.bff" to be rebuilt as it is the last included bff file. "lib_n.bff" would trigger "lib_n+1.bff" and all following bff files.

Anyway, I'm beginning to experiment with caching and distributing. I tried with fastbuild itself. I just modified the fuild.bff to compile with VS2015 under Windows 10 ( cf my fork to see the modifications). The only files that are cached are the LZ4 files, all others are marked uncacheable in the report. Is this the expected behavior with MSVC?
Thx

ffulin:
It's very common that configuration be split across many files.  To reliably know which file affects which target is unfortunately the same as determining which bit of the file relates to which target, so what you propose sadly requires the same solution that's already being worked on.

For the caching:
FASTBuild supports generating Unity files to accelerate compilation.
FASTBuild also supports isolating locally modified files out of the Unity to improve iteration times. It does this by checking the read-only status. The idea is that if you use something like Perforce for source control, your files are usually read-only, and are writeable only when you "open for edit" those files and start modifying them.
[...]

PR #91, opened 2016-03-03
Check hash of BFF file if its timestamp differs from one saved in the node DB

dummyunit:
This commit adds hash-based check in addition to time stamp check to determine if .bff is actually changed before deciding that re-parsing is needed. It should prevent unnecessary rebuilds caused by saving .bff without changing it or by git modifying file time stamp when switching branches back and forth, etc.
This should help mitigate "FASTbuild wants to rebuild everything" situations while re-factoring (mentioned in #55) that should bring more fine grained rebuilding rules is under way.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
@@ -209,8 +209,8 @@ bool NodeGraph::Load( IOStream & stream, bool & needReparsing )
 		const uint64_t dataHash = xxHash::Calc64( mem.Get(), size );
 		if ( dataHash == m_UsedFiles[ i ].m_DataHash )
 		{
-			// set file modification time to the one saved in the DB to save time on the next run
-			FileIO::SetFileLastWriteTime( fileName, timeStamp );
+			// file didn't change, update stored timestamp to save time on the next run
+			m_UsedFiles[ i ].m_TimeStamp = timeStamp;
 			continue;
 		}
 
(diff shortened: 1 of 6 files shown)


Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -187,12 +188,35 @@ bool NodeGraph::Load( IOStream & stream, bool & needReparsing )
...
+		if ( fs.Read( mem.Get(), size ) != size )
+		{
+			return false; // error reading
+		}
+
+		const uint64_t dataHash = xxHash::Calc64( mem.Get(), size );
+		if ( dataHash == m_UsedFiles[ i ].m_DataHash )
+		{
+			// set file modification time to the one saved in the DB to save time on the next run
+			FileIO::SetFileLastWriteTime( fileName, timeStamp );
```
ffulin: Instead of modifying the file stamp, perhaps we should modify m_TimeStamp to match that of the file. This has a couple of benefits:
 - Won't cause dirtying of build for other bff hierarchies using the same bff file
 - Won't impact source control that uses modification time to reflect something (i.e. perforce "modtime" option)

ffulin:
Hello.  Thanks for the pull request. Your change looks good.  I've added minor not about the file time management for your consideration. Let me know what you think. :)

Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
dummyunit: Sure, I didn't think of these situations. I will update this pull request.

ffulin:
I've integrated this into the Perforce mainline. This change will be in the next release (v0.90).
Thanks for your contribution!

PR #96, opened 2016-05-05
Uniquify database using root bff filename

jelmansouri:
Instead of invalidating the cache when building using different bff file I enforce one fdb by bff by adding .fdb to the bff file.
Since flushing the fdb file would not be necessary, It would be more efficient this way and it covers our prod use cases.

Tested on Rainbow6 with the following scenario:
 * folder containing fbuild.vs2012.bff & fbuild.vs2015.bff
 * switching between the two files and making sure that the fdb file gets properly created and updated
 * checking that file time-stamp comparison on the list of files on the fdb is still working

Unfortunately I could not run the FBuiltTests since I hacked very minimally the FBuild root bff to get a non dist build work with VS2012.

### DIFF
--- a/Code/Tools/FBuild/FBuildApp/Main.cpp
+++ b/Code/Tools/FBuild/FBuildApp/Main.cpp
@@ -430,7 +430,7 @@ int Main(int argc, char * argv[])
 	}
 
 	// load the dependency graph if available
-	if ( !fBuild.Initialize( FBuild::GetDependencyGraphFileName() ) )
+	if ( !fBuild.Initialize() )
 	{
 		if ( sharedData )
 		{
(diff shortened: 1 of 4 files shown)


ffulin:
Hello.  Unfortunately I had to revert this change because it introduces a stack corruption which causes the tests to fail.

The change seems reasonable in principal, but this crash will need to be investigated and resolved before I can re-merge.

ffulin:
Github doesn't seem to let me re-open this pull request, so we'll need to create a new pull request once the issue is resolved.

ffulin:
I've re-integrated this change and resolved the issue with the tests (https://github.com/fastbuild/fastbuild/commit/43f3e5dc201055d1a3ad7a6e5225ed1e93ad66ad)

The issue was caused by the test relying on the existing naming behaviour of the fdb, which changed.  I've robustified the test and also modified the pull request to retain the existing behaviour so that fdb file names will not change for existing users. (i.e. fbuild.bff -> fbuild.fdb instead of fbuild.bff.fdb)

Thanks for the fix!

PR #97, opened 2016-05-05
Some behavior enhancements that make things easier for .bff generators

Tengato:
While working on the GENie fastbuild generator, I encountered a few difficulties:

1. cl.exe will not proceed when passed a /Fd or /Fa option if the dir doesn't exist.  (/Fd and /Fa specify a custom dir path for .pdb and .asm output files).  It seems like MSBuild creates the dirs beforehand, so I figured FBuild should handle that too.

2. CompilerInputFiles specified separately have their compiler output objs placed into a flat target dir.  When two source files have the same file name, a conflict occurs.

These changes are my attempt to address those issues.   For the second issue, I just have the node generator try to use the CompilerInputPath bff property as a base dir for the file path.  If it doesn't match, the obj just goes to the target folder like before - to do this I removed an assert that didn't seem necessary, but maybe there is a case I'm not aware of.  Since I don't want to have the CompilerInputPath add filesfor this purpose, I just block that feature by adding an always false CompilerInputPattern as well.


h0tw1r3:
Very nice feature. Genie's experimental fastbuild support works quite well with this.

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionLibrary.cpp
```diff
@@ -72,7 +74,8 @@ FunctionLibrary::FunctionLibrary()
 		return false;
 	}
 
-    PathUtils::FixupFolderPath( compilerOutputPath );
+	PathUtils::FixupFolderPath( compilerInputPath );
```
ffulin: (I'll fix this as part of the intergation)

This actually needs to be like this:
```C++
	if ( !compilerInputPath.IsEmpty() )
	{
		AStackString<> tmp;
		NodeGraph::CleanPath( compilerInputPath, tmp );
		compilerInputPath = tmp;
	}
```
Without the IsEmpty() check, compilerInputPath would not be empty by default (FixupFolderPath would always add a trailing slash).  Also making this a CleanPath here removes the need for cleaning the path at build time.

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
```diff
@@ -51,20 +51,23 @@ FunctionObjectList::FunctionObjectList()
...
+		 !GetString( funcStartIter, compilerInputPath, ".CompilerInputPath", false ) ||
 		 !GetString( funcStartIter, compilerOutputPath, ".CompilerOutputPath", true ) ||
 		 !GetString( funcStartIter, compilerOutputPrefix, ".CompilerOutputPrefix", false ) ||
 		 !GetString( funcStartIter, compilerOutputExtension, ".CompilerOutputExtension", false ) )
 	{
 		return false;
 	}
 
-    PathUtils::FixupFolderPath( compilerOutputPath );
+	PathUtils::FixupFolderPath( compilerInputPath );
```
ffulin: I'll make the same change as FunctionLibrary.cpp here.

Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
```diff
@@ -349,14 +351,14 @@ bool ObjectListNode::CreateDynamicObjectNode( Node * inputFile, const DirectoryL
...
     // to replicate the folder hierearchy in the output
     AStackString<> subPath;
-    if ( dirNode )
+    if ( baseDir.GetLength() > 0 )
     {
-        const AString & baseDir = dirNode->GetPath();
-        ASSERT( PathUtils::PathBeginsWith( fileName, baseDir ) );
-        if ( PathUtils::PathBeginsWith( fileName, baseDir ) )
+		AString fullBaseDir;
+		NodeGraph::CleanPath(baseDir, fullBaseDir);
```
ffulin: As noted, by doing the CleanPath during parsing, we avoid doing it at build time.  I'll take care of this as part of the integration.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -1407,7 +1407,64 @@ bool ObjectNode::BuildArgs( const Job * job, Args & fullArgs, Pass pass, bool us
                 continue;
             }
         }
-				
+
```
ffulin: This is a good fix, but it's a little unfortunate that:
a) we perform this logic redundantly for every obj file (potentially hundreds or thousands of times in a build).  Of particular note is that EnsurePathExists is a call into the kernel.
b) we do it in BuildArgs, which isn't really a function that should have side-effects

Ideally, we'd have this logic performed once by the ObjectList or Library that owns the ObjectNodes, so we'd only have to do it once.  Sadly, I don't think there's a good way to do that right now.  I'll need to give it some more thought.

I think I can integrate this as-is though, and look into options for reducing the overhead.

Tengato: I agree 4/4.  Let me know if you need help with anything - thanks!

ffulin:
I've integrated this change into the Perforce mainline.  This will be in the next release (v0.90).

I made a few fixes/improvements which I've also pulled into the github dev branch (https://github.com/fastbuild/fastbuild/commit/0ae921186c44235877ac51224f38e1be36a96e7d).
Specifically:
 - pdb and asm paths are detected and stored during bff parsing instead of build time
 - we check that these dirs exist only once per Library/ObjectList
 - we copy with /Fa and /Fd with no folder (i.e. filename only) (there was null deref crash in the pull request)
 - I added tests for /Fa and /Fd set to folders other than normal compilation output

ffulin:
Further to this, for the .CompilerInputPaths related part of this change, another user created an alternative option .CompilerInputFilesRoot (https://github.com/fastbuild/fastbuild/pull/85) which I've integrated into dev already (and will be in v0.90).

This option is preferable because:
a) The new property is more explicit, so you no longer need to set a dummy wildcard pattern
b) It fixes a bug which was introduced by your change - it broke support for multiple .CompilerInputFiles (i.e. it can actually be a list of paths)

With that in mind, I'm going to remove the base path part of this change. Will it be ok for you to update GENie to use the .CompilerInputFilesRoot option instead?

(The /Fa and /Fd part of this change is not impacted)

Tengato:
Of course - that should be a straightforward switch.  Sorry for clunking
around in the code like that!

On a related note, We stopped using /Fa and /Fd in our projects because I
switched to the C7 symbol format in order to take advantage of caching.  It
seems like a few of our unit tests are failing for cached builds though, so
I'm currently investigating that.

Thanks!

ffulin:
Interestingly, I've found object files compiled with Z7 debug info typically compile faster too (presumably from reduced contention on the PDB), so you might see an improvement there as well.

Caching support for Zi/PDB debug info is something I hope to support in the future as well.

PR #103, opened 2016-06-08
Added workaround which allows caching of items consuming WinRT

ClxS:
This is a small workaround which allows some items consuming the windows runtime to be cached. It works just by stripping the /ZW flag and appending the define used inside /ZW included files on the preprocessor pass. My guess for why it fails without this workaround is that /ZW force includes certain files which does not play nice with two-step compilation like caching/distribution uses.
Distribution still does not work properly with this workaround yet.

This was mainly tested on UE4 on console platforms, so may not work on all projects depending on layout.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -815,18 +815,18 @@ bool ObjectNode::ProcessIncludesWithPreProcessor( Job * job )
 
 		// 1) clr code cannot be distributed due to a compiler bug where the preprocessed using
 		// statements are truncated
-		// 2) code consuming the windows runtime cannot be cached due to preprocessing weirdness
+		// 2) code consuming the windows runtime cannot be distributed due to preprocessing weirdness
 		// 3) pch files are machine specific
 		// 4) user only wants preprocessor step executed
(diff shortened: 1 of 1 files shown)


ffulin:
Hi.  This change seems pretty reasonable.  The only (small) concern is that this *could* be a breaking change if there are some cases where it still doesn't work.  Given that this feels like an overall improvement though, my thinking is that:
a) We can call that out in the release notes
b) People can retain the previous behaviour by explicitly opting any problem ObjectList/Library out of caching with .AllowCaching = false

In my experience (for Xbox One) it was non-trivially beneficial to compile times to greatly limit the amount of things compiled with /ZW.  Out of interest, does Unreal allow you to do that?  Or does it force you to have /ZW on everything?

ClxS:
They used to only use it (/ZW) on objects that actually needed it prior to v4.5, but since then it's been forced onto the entire library.
It would be possible to alter the UnrealBuildTool to only use /ZW on object which actually need it though I'm not sure how much work it'd be.

Maybe it could be best to set AllowCaching to false by default on things using WinRT, that way it's an opt-in that people would be able to use if it works as expected for their project?

ffulin:
I've integrated this into the Perforce mainline and added unit tests.  This will be in the next release (v0.90)

I ended up leaving it enabled by default.  There will be a note about the behaviour change in the release notes.

Thanks for the pull request!

PR #110, opened 2016-07-02
Allow concatenation/assigning of empty arrays to other arrays

dummyunit:
Following statements are valid when .X is a string but not when it's a
struct:
```
.E = {}
.A = { .X }
.T1 = {} + .X
.T2 = {} + { .X }
.T3 = {} + .A
.T4 = {} + { .A }
.T5 = { .X } + {}
.T6 = { .X } + .E
.T7 = { .X } + { .E }
```
This is a side effect of empty arrays being stored as ArrayOfStrings.

This change allows treatment of empty ArrayOfStrings as empty
ArrayOfStructs and vice versa, making those statements valid when .X is
a struct as well.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
@@ -1274,7 +1274,15 @@ bool BFFParser::StoreVariableArray( const AString & name,
 				return false;
     		}
 
-			if ( var->IsString() || var->IsArrayOfStrings() )
+			if ( var->IsArrayOfStrings() && var->GetArrayOfStrings().IsEmpty() )
+			{
+				// appending empty array, nothing to do
+			}
+			else if ( var->IsArrayOfStructs() && var->GetArrayOfStructs().IsEmpty() )
+			{
+				// appending empty array, nothing to do
+			}
+			else if ( var->IsString() || var->IsArrayOfStrings() )
 			{
 				// dest is consistent?
 				if ( structValues.IsEmpty() == false )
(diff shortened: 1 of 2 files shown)


ffulin:
Hello.  Thanks for doing this - it will greatly enhance the utility of the ^ operator in loops of structures.  I also really appreciate the updates to the tests.

If I interpret the changes correctly, this change seems to allow:
a) The **destination** variable to change types if is empty
b) The **destination** variable to change types if the source is empty
c) The concatenation of mismatched types to be silent if the **source** is empty

The general rule for variables in the bff is that once they declared, they cannot change type.. I think (a) is ok as an exception as we can consider empty arrays to be "typeless".  However I think (b) and (c) are problematic: I think they should not be allowed as they can hide configuration problems - i.e. I think they should continue to be errors.

What do you think?

dummyunit:
This change allows destination to change its type only between ArrayOfStrings and ArrayOfStructs and only when:
   1. Destination variable is an **empty array** before operation. This is to allow adding struct members (`.X = {} + .Struct`) and appending contents of existing ArrayOfStructs (`.X = {} + .ArrayOfStructs`) to empty array.
   2. **Empty array** is explicitly assigned to destination variable (e.g. `.X = {}` and not `.X = .EmptyArray`). This is a result of inconsistency I have discovered in #113, this will go away (`.X` would retain its type but become empty) automatically when type checking on explicit value assignment will be made stricter.

So to address your points:
   - In (b) type will only change when empty array is assigned explicitly: `.X = {}`. When other variable, that hold an empty array, is assigned type remains the same, here is an example:
```
$ cat test.bff 
.EmptyArray = {}
.Struct = [ .Value = 42 ]
.X = { .Struct }
.X = {}
.Y = { .Struct }
.Y = .EmptyArray
$ fbuild -verbose -config test.bff 
Registered <ArrayOfStrings> variable '.EmptyArray' with 0 elements
Registered <int> variable '.Value' with value '42'
Registered <struct> variable '.Struct' with 1 members
Registered <ArrayOfStructs> variable '.X' with 1 elements
Registered <ArrayOfStrings> variable '.X' with 0 elements
Registered <ArrayOfStructs> variable '.Y' with 1 elements
Registered <ArrayOfStructs> variable '.Y' with 0 elements
```
   I don't think this will hide any configuration errors, because type is changed only when user is explicitly clearing an array.
   - In (c), yes, it will hide this kind of error:
```
.Bar = { .StructA .StructB }
.Baz = {} // Declared empty and then populated in ForEach, but happened to be empty (and that is correct)
.Foo = { .StructC } + .Baz // Oops, wanted to add .Bar but misspelled, and now because .Baz is empty FASTBuild won't complain
```
   But what if I really wanted to add `.Baz`? FASTBuild can't possibly know that and without these changes it will complain. So I think it is better to allow some shooting in the foot in exchange for more flexible syntax.

Also here is summary on what works now and what doesn't:
   - `.X = {} + { .S }`, `.X = {} + { .ArrayOfS }` and `.X = { .S } + {}` works when `.S` is a string, and when `.S` is struct but only because of #113 (type of existing variable isn't being checked).
   - `.X = {} + .S`, `.X = {} + .ArrayOfS`, `.X = { .S } + .EmptyArray` and `.X = { .S } + { .EmptyArray }` works only when `.S` is a string, this change makes them allowed when `.S` is a struct.

One possibly unwanted result of this change is that it will allow conversion of non-empty ArrayOfStrings variable into non-empty ArrayOfStructs by assigning empty array to it first:
```
.Struct = [ .Y = 1 ]
.X = { "one" "two" }
.X = {}
.X + .Struct
```
However, because of #113, it is already possible with this syntax:
```
.Struct = [ .Y = 1 ]
.X = { "one" "two" }
.X = {}
.X + { .Struct }
```

That being said, there are other ways to improve support of empty ArrayOfStructs:
   1. Invent new syntax to declare an empty ArrayOfStructs instead of `.X = {}`. This will allow to ban any conversion between ArrayOfStrings and ArrayOfStructs (even between empty), but it will break (after fix for #113) peoples configs in a bad way: new configs wouldn't work with old version because of new syntax, and old configs wouldn't work with new version because `.X = {} + { .Struct }` (and others similar) would be banned.
   2. Introduce new internal type of `BFFVariable` — `VAR_ARRAY_EMPTY`. Only declarations of new variables will produce this type (e.g. `.VarNotPresentInCurrentFrame = {}`), and only conversions from `VAR_ARRAY_EMPTY` to `VAR_ARRAY_OF_STRINGS` and from `VAR_ARRAY_EMPTY` to `VAR_ARRAY_OF_STRUCTS` will be allowed. This will require inspection of all code that uses `BFFVariable`  to ensure that `VAR_ARRAY_EMPTY` is handled properly.

ffulin:
I think you're right - the change makes sense as-is. If an empty array is "typeless" until something is put in it, then clearing an array should restore it to the "typeless" state, which your change essentially does.

ffulin:
I've integrated this change into the Perforce mainline. It will be part of the next release (v0.91).
Thanks again! :)

PR #111, opened 2016-07-02
Add missing guard for windows-only test

dummyunit:
This fixes test building on Linux and Mac.


ffulin:
Hello.  This fix is good, but I just fixed this with tHe latest integration of the v0.90 release.  With that in mind, I won't integrate this pull request. Thanks :)

PR #112, opened 2016-07-02
Remove unwanted files and ban them in gitignore

dummyunit:
Those generated files (FBuild binary and couple of profile.json) probably shouldn't be stored in repository. :)


ffulin:
Nice one - thankyou :)

Issue #113, opened 2016-07-02
Inconsistency in variable assignment from other variables and explicit values


dummyunit:
While working on #110 I have stumbled upon this inconsistent behaviour:
```
.TheAnswer = 'fourty two'
.X = 42
.X = 'fourty two'
.Y = 42
.Y = .TheAnswer
```
Here second assignment to .X works fine, changing its type to integer. But second assignment to .Y (which, as one could expect, should produce the same value as in .X) fails with this error:
```
/home/dummyunit/fastbuild/Code/test.bff(5,6): FASTBuild Error #1034 - Operation not supported: 'Int' = 'String'.
.Y = .TheAnswer
     ^
     \--here
```
Is this inconsistency by design? If not, which behaviour is correct: changing variable type or producing an error?

ffulin:
Hello.  The rule so far has been "a variable cannot change type" once created.  I think your work on #110 is ok to consider as exception if we consider empty arrays to be "typeless" until assigned.

Being able to change the type of .X is a bug.  It should fail with the same error as the .Y case.

ffulin:
With the changes you made (which are in v0.91), I believe this issue can now be resolved. Your snippet now results in:
```C++
BFF file 'C:\p4\depot\Code\fbuild.bff' has changed (reparsing will occur).
C:\p4\depot\Code\fbuild.bff(3,6): FASTBuild Error #1034 - Operation not supported: 'Int' = 'String'.
.X = 'fourty two'
     ^
     \--here
```
What do you think?

dummyunit:
Yes, this looks good now.

PR #117, opened 2016-07-04
Improve variable handling logic

dummyunit:
This series of commits bans most ways for a variable to change its type, which were discovered in #113, by adding checks that ensure that variable to be modified has same or compatible type.

It also fixes some others bugs/inconsistencies discovered along the way:
   - Subtraction operator wasn't handled properly in many cases. In operations where right operand was variable or explicit structure it worked like assignment. With this changes:
      1. Subtractions of structs are banned.
      2. Subtractions like `.String1 - .String2` work like `.String1 - "explicit string"`, removing occurrences of `.String2` from `.String1`.
      3. Subtractions like `.ArrayOfStrings - .String` work like `.ArrayOfStrings - "explicit string"`, removing elements that are equal to `.String` from `.ArrayOfStrings`.
      4. Added subtractions for integers (that was easier than adding error message :) ).
      5. Other subtractions are banned (bool and arrays).
   - Integer addition only worked when right operand was a variable. With this changes it will work with explicit values as well.

I have also changed bool value parsing to be more clear and not use cryptic constructions like `iter.ParseToNext( 'e' )`.

I must also mention that with these fixes code structured like that (which probably used by many people) now produces an error:
```
$ cat test.bff 
.EnabledX = { "Foo" "Bar" }
.Array = {}
ForEach( .X in .EnabledX )
{
        .Struct =
        [
                .A = .X
                .B = "-option $X$"
        ]
        ^Array + { .Struct }
}
$ fbuild -verbose -config test.bff
Registered <ArrayOfStrings> variable '.EnabledX' with 2 elements
Registered <ArrayOfStrings> variable '.Array' with 0 elements
Function call 'ForEach'
Freezing loop array '.EnabledX' of type <ArrayOfStrings>
Registered <string> variable '.A' with value 'Foo'
Registered <string> variable '.B' with value '-option Foo'
Registered <struct> variable '.Struct' with 2 members
/home/dummyunit/fastbuild/Code/test.bff(10,13): FASTBuild Error #1034 - Operation not supported: 'ArrayOfStrings' + 'Struct'.
        ^Array + { .Struct }
                   ^
                   \--here
```
So, if this will be merged, something must done to support empty arrays — either updated version of #110 or different internal type for fresh empty arrays (like I described in comment 3 in #110), or something else.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFIterator.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFIterator.cpp
@@ -209,6 +209,35 @@ bool BFFIterator::ParseToMatchingBrace( char openBrace, char closeBrace )
 	return false;
 }
 
+// ParseExactString
+//------------------------------------------------------------------------------
+bool BFFIterator::ParseExactString( const char * string )
+{
+	const char * const originalPos = m_Pos;
+
+	const char * stringPos = string; // stores pointer to the next character to be checked for presence in our sequence
+	while ( !IsAtEnd() && *stringPos != '\000' )
+	{
+		if ( *m_Pos != *stringPos )
+			break;
+		m_Pos++;
+		stringPos++;
+	}
+
+	if ( *stringPos == '\000' )
+	{
+		// we are at the end of string, that means all characters have successfuly matched
+		return true;
+	}
+	else
+	{
+		// we are not at the end of string, that means either our sequence ended, or there is non matching character
+		// restore our pos before returning
+		m_Pos = originalPos;
+		return false;
+	}
+}
+
 // IsAtValidVariableNameCharacter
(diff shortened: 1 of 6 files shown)


ffulin:
This change seems sensible to me.  Could you please resolve it against your previous commit?
Thanks again for you work on this!

dummyunit:
Done. Rebased on top of dev and adjusted some commits (mostly the first 2).

Also added one new commit (the last one) that fixes one rare case I didn't account for in #110: addition/subtraction/assignment of explicit string to empty ArrayOfStructs wasn't working, while the same with string _variable_ and empty ArrayOfStructs was.

ffulin:
I've integrated this into the Perforce mainline.  It will be in the next release (v0.91).  Thanks!

PR #119, opened 2016-07-10
Increase maximum variable length

poppolopoppo:
From 64 to 256.
The limit on MSVC/ICC in 2048, and GCC has no limit.

I hit this problem with generated variable names and had to rise the limit.


ffulin:
The original limit was just as a sanity check, so this increase is totally reasonable.

I've integrated the change into the Perforce mainline, and this will be in the next release (v0.91)

Thanks :)

PR #120, opened 2016-07-10
VSSolution() function will merge all solution folders with the same path

poppolopoppo:
This allow a more user friendly usage :
```
.SolutionFolderA =
[
    .Path = 'Test'
    .Projects = { "ProjectA" }
]

.SolutionFolderB =
[
    .Path = 'Test'
    .Projects = { "ProjectB" }
]

VSSolution()
{
    // [...]
    .SolutionFolders = { .SolutionFolderA, .SolutionFolderB }
}
```
In this example `ProjectA` and `ProjectB` will be both in the same folder `Test`.
Moreover, an empty path will be considered as the root solution folder.

Declarations can now be split thanks to those changes.


ffulin:
This change seems pretty reasonable to me. I'll get that integrated soon.  Sorry for the slow turn-around.

ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.92).
Thanks for your contribution!

PR #121, opened 2016-07-10
integers and booleans can be formated into strings

poppolopoppo:
ex:
````
.int = 42
Print( 'format int = $int$' )
=> 'format int = 43'
````
````
.bool = false
Print( 'format int = $bool$' )
=> 'format int = false'
````


ffulin:
I've integrated this into the Perforce mainline.  This will be in the next release (v0.91).  Thanks for the changes! :)

PR #130, opened 2016-08-20
Add Exclude Patterns to ObjectList-Based Functions

fitzymj:
This change adds ".CompilerExcludePattern" to ObjectList-based Functions like Library, Exe, ObjectList, etc.  It includes:
- A DB version bump
- Tests for ObjectList and Unity exclusion patterns
- Whitespace (spaces->tabs) fixes around the adjacent file exclusion list properties
- A forced call to `PostTest()` after a test failure so callback outputs clean up and `Tracing::AddCallbackOutput` doesn't hit an assert on the following test

Also I am barely functional with GitHub, so let me know if I need to change anything about this process or pull request.  Thanks!

### DIFF
--- a/Code/TestFramework/UnitTestManager.cpp
+++ b/Code/TestFramework/UnitTestManager.cpp
@@ -130,6 +130,7 @@ bool UnitTestManager::RunTests( const char * testGroup )
 		catch (...)
 		{
 			OUTPUT( " - Test '%s' *** FAILED ***\n", s_TestInfos[ s_NumTests - 1 ].m_TestName );
+			s_TestInfos[ s_NumTests - 1 ].m_TestGroup->PostTest();
 		}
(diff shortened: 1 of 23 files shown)


fitzymj:
I'm just realizing that the old filtering at `UnityNode.cpp:397` is probably now redundant, since it's operating on the DirectoryListNode files that have now _already_ been filtered.

ffulin:
I see the change wasn't quite as straightforward as I claimed - sorry about that!  Your change looks good though.  I'll look at integrating that soon.

I really appreciate the extra effort in making the feature consistent (in CSAssembly for example) and the update of the  tests and documentation!

ffulin:
I've integrated this into the Perforce mainline.  This will be in the next release (v0.92)
Thanks for your contribution!

PR #131, opened 2016-08-21
Add documentation for the Test function

Manuzor:
I've written some documentation for the `Test()` function. If anything is missing that you'd like to see here, like some details I didn't mention, please let me know and I'll add it.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/test.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/test.html
@@ -55,7 +55,7 @@ Defines the execution of an executable as a test.</p>
       <p><b>.TestWorkingDir</b> - String - (Optional)</p>
       <p>Working dir for test execution.</p>
       <hr>
-      <p><b>.TestTimeOut</b> - Number - (Optional)</p>
+      <p><b>.TestTimeOut</b> - Integer - (Optional)</p>
       <p>The amount of milliseconds to wait for a test to finish execution.</p>
       <p>The default is 0, which means there is no timeout and FASTBuild will wait until the executable terminates on its own.</p>
     </div>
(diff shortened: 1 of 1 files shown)


Code/Tools/FBuild/Documentation/docs/functions/test.html
```diff
@@ -12,18 +12,64 @@
...
+      <p>When executing tests, FASTBuild will capture standard output channels of the executable and write them to this file when done.</p>
+      <hr>
+      <p><b>.TestArguments</b> - String - (Optional)</p>
+      <p>Arguments to pass to test executable.</p>
+      <hr>
+      <p><b>.TestWorkingDir</b> - String - (Optional)</p>
+      <p>Working dir for test execution.</p>
+      <hr>
+      <p><b>.TestTimeOut</b> - Integer - (Optional)</p>
+      <p>The amount of milliseconds to wait for a test to finish execution.</p>
```
ffulin: The timeout is in seconds.  I'll fix this up after integrating.

(Confusingly, until I fixed it in v0.91, the timeout was documented in seconds, but actually implemented in ms. I have fixed the implementation to be in seconds, which was what I originally intended)

ffulin:
Thanks for your help in improving the docs.  I'll get this integrated soon.

ffulin:
I've integrated this change into the Perforce mainline. It will be part for the next release (v0.92).
Thanks for your contribution!

Manuzor:
You're most welcome!

Code/Tools/FBuild/Documentation/docs/functions/test.html
Manuzor: I've actually looked at the implementation of v.90 and saw that it was milliseconds, so I wrote that down in the documentation. I'm glad you're going for seconds instead because I think it makes more sense from a user's perspective! :+1:

PR #140, opened 2016-09-09
Add diagnostics to worker brokerage initialisation

samhocevar:
The “No workers available - Distributed compilation disabled” message
could be more helpful; this change adds more explanations for the user.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
+++ b/Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
@@ -9,6 +9,7 @@
 
 // FBuild
 #include "Tools/FBuild/FBuildCore/Protocol/Protocol.h"
+#include "Tools/FBuild/FBuildCore/FLog.h"
 
 // Core
 #include "Core/Env/Env.h"
@@ -83,6 +84,7 @@ void WorkerBrokerage::FindWorkers( Array< AString > & workerList )
 
     if ( m_BrokerageRoot.IsEmpty() )
     {
+        FLOG_WARN( "No brokerage root; did you set FASTBUILD_BROKERAGE_PATH?" );
         return;
     }
 
@@ -92,6 +94,7 @@ void WorkerBrokerage::FindWorkers( Array< AString > & workerList )
                             false,
                             &results ) )
     {
+        FLOG_WARN( "No workers found in %s", m_BrokerageRoot.Get() );
         return; // no files found
     }
 


Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
```diff
@@ -92,6 +94,7 @@ void WorkerBrokerage::FindWorkers( Array< AString > & workerList )
                             false,
                             &results ) )
     {
+        FLOG_WARN( "No workers found in %s", m_BrokerageRoot.Get() );
```
ffulin: FASTBuild already prints:
```C++
No workers available - Distributed compilation disabled
```
What case does this help disambiguate?

Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
```diff
@@ -83,6 +84,7 @@ void WorkerBrokerage::FindWorkers( Array< AString > & workerList )
 
     if ( m_BrokerageRoot.IsEmpty() )
     {
+        FLOG_WARN( "No brokerage root; did you set FASTBUILD_BROKERAGE_PATH?" );
```
ffulin: A pretty common use-case is to have project generation setup to always pass -dist on the command line, but to selectively define the FASTBUILD_BROKERAGE_PATH only for some users.  A great example of this is for a company with  remote workers - you might define the path for local users, but not for remote workers.  This change would cause those users to always get a warning.

samhocevar:
I understand that use case, but I think the current behaviour breaks the principle of least astonishment. If the user passes `-dist`, it probably means they want distribution, but most likely they want to know *why* there is no distribution.

If the user deliberately unsets `FASTBUILD_BROKERAGE_PATH` (or sets it to an invalid location), they should not be surprised to get a (harmless) warning message.

ffulin:
On reflection, I think you're right about the "No brokerage root" warning.  We can enable that and it should not really cause any problems.

For the other error (No workers found in %s), do you think it helps over and above the warning you already get? (No workers available - etc)

samhocevar:
I think the `%s` part of the other diagnostic message is helpful, too. We have build tools that can dynamically set `FASTBUILD_BROKERAGE_PATH` depending on which machine is building (some of them are in a remote datacenter but can still talk to each other). A few times I could debug issues by looking at that variable’s actual value in build logs.

samhocevar:
I should add that the above scenario was actually with `distcc`, but now I would be very happy to have equivalent error reporting when using FASTBuild.

ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.92).  Thanks for your contribution!

PR #144, opened 2016-09-13
Add bash completion script

dummyunit:
Offered completions:
   1. All fbuild options.
   2. Files after -config.
   3. Targets (parsed from -showtargets output), when config is valid.

Tested to work on Linux, in msysGit bash (old (<2.0.0) Git for Windows)
and in MSYS2 bash (new (>=2.0.0) Git for Windows).

To install fbuild.bash-completion, either:
   1. (locally) Source it from your ~/.bashrc
   2. (system-wide) Copy it to /etc/bash_completion.d/ (requires modern
       bash-completion).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,9 @@
 # Auto detect text files and perform LF normalization
 * text=auto
 
+# Bash scripts have to have LF line endings, otherwise they work only on Windows
+*.bash-completion eol=lf
+
 # Custom for Visual Studio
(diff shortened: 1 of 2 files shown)


ffulin:
Looks good - thanks for this.  I'll get that integrated soon and I'll include that in the available downloads.

ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.92) Thankyou once again for your contribution!

PR #148, opened 2016-09-19
Basic support for .prx (Orbis .dll)

jairbubbles:
This is a quick fix I've made to make Orbis prx work on my project. This is not something used much but still I'm sure there is room for improvement.

### DIFF
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -187,11 +187,11 @@
 //------------------------------------------------------------------------------
 /*static*/ void PathUtils::StripFileExtension( AString & filePath )
 {
-	const char * lastDot = filePath.FindLast('.');
-	if (lastDot)
-	{
-		filePath.SetLength(lastDot - filePath.Get());
-	}
+    const char * lastDot = filePath.FindLast('.');
+    if (lastDot)
+    {
+        filePath.SetLength(lastDot - filePath.Get());
+    }
 }
 
 //------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ffulin:
Is it the case that when creating a dll with the Orbis compiler, that a side-by-side file is created which has this "weak" naming extension by default?
I've never *created* dlls when compiling for Orbis, and I don't currently have access to the Orbis toolchain to check the behaviour.

Assuming this is the case, this seems like a reasonable improvement to the default behaviour.

jairbubbles:
Yes that's the idea. you can also have "not weak" stub libraries but it's not supported yet.

I feel that the command line detection is very simplist as something like ""--oformat = prx" (with spaces) will not be detected.

Also, I duplicated the code `// get name minus extension (handle no extension gracefully)` that could be transformed as a helper function somewhere. As I don't know the code source well I didn't know where to put it so I duplicated the code like an a pig :-p

ffulin:
Does clang accept spaces between the arguments like that? It would of course be ideal to have it robust in that case.

As for the code duplication, I don't consider it particularly egregious :)  There is always a threshold with these things and this isn't clearly over that in my mind. Having said that, one easy way to reduce the duplication would be to add a function like this to the PathUtils.cpp:
```C++
/*static*/ void PathUtils::StripFileExtension( ASTring & filePath )
{
    const char * lastDot = filePath.FindLast( '.' );
    if ( lastDot )
    {
        filePath.SetLength( lastDot - filePath.Get() );
    }
}
```

ffulin:
I have integrated this fix into the Perforce mainline. It will be in the next release (v0.92).  Thanks for the fix!

PR #149, opened 2016-09-22
Drop the requirement that Exec node inputs must be files.

samhocevar:
There is no particular reason to require Exec inputs to be files only. There
are situations where one would want to operate on object files, for instance,
or on Executable nodes for symbol stripping. Potentially, any rule output may
be used as input for an Exec node.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionExec.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionExec.cpp
@@ -78,22 +78,6 @@ FunctionExec::FunctionExec()
     {
         return false; // GetNodeList will have emitted an error
     }
-    else
-    {
-        // Make sure all nodes are files
-        const Dependency * const end = inputNodes.End();
-        for (const Dependency * it = inputNodes.Begin();
-            it != end;
-            ++it)
-        {
-            Node * node = it->GetNode();
-            if (node->IsAFile() == false)
-            {
-                Error::Error_1103_NotAFile(funcStartIter, this, "ExecInput", node->GetName(), node->GetType());
-                return false;
-            }
-        }
-    }
 
     // optional args
     const AString & arguments(  argsV ?         argsV->GetString()      : AString::GetEmpty() );


ffulin:
Hello.  The examples you gave (Object Files and Executables) are files, so they should be ok with this check. Using Exec to strip debug info as an example is something that already works.

The logic you've removed is designed to protect against cases that won't work reliably - i.e. something that doesn't output a file and therefore can't be correctly tracked for detecting when builds should happen.

Can you provide some more information about the use-case you have that you were having problems with that (presumably) was triggering this "Is Not A File" error when you expected it should work?

samhocevar:
Thanks for the comments. The issue is as explained in this patch for Unreal_FastBuild: https://github.com/liamkf/Unreal_FASTBuild/pull/8

The problematic rule combination is roughly:

    ObjectList('Action_0')
    {
        .Compiler = 'UE4Compiler'
        [...]
        .CompilerOutputExtension = '.obj'
    }

    Exec('Action_1')
    {
        .ExecExecutable = 'c:\Windows\system32\cmd.exe'
        .ExecArguments = '/c "type D:\build\file > D:\build\file.tmp"'
        .ExecInput = {
            'D:\build\file',
            'Action_0',
        }
        .ExecOutput = 'D:\build\file.tmp'
    }

It fails with the following message:

    5>  fbuild.bff(20,1): FASTBuild Error #1103 - Exec() - 'ExecInput' ('Action_0') is a 'ObjectList', which does not describe a file.
    5>  Exec('Action_1')
    5>  ^
    5>  \--here

samhocevar:
As a side note, I think that this is one of several situations where FASTBuild tries to be smarter than the user by actively preventing possibly dangerous behaviour, but also preventing otherwise desirable mechanisms. I believe a more liberal stance could be adopted, in this case by just issuing warnings such as `Warning: node <nodename> does not always output files, dependency tracking for <othernodename> may be unreliable`.

ffulin:
Thanks for the extra info.  I spent some time going over the Exec code again in the context of this discussion. The TL;DR: is that the change you've made works for your specific use-case, but isn't safe in the general case.  I'll provide some background to explain this.

The ExecInput property is currently expanded at bff parsing time, which means that it doesn't support expansion to things which might change at build time.  ObjectList is an example of this, since you can add/remove source files with the change. (I was wrong when I originally indicated that this should work).

The "Is Not A File" check needs to be an error here because the user would expect the list of objects built by ObjectFile to be expanded in their Exec args at build time (as %1) but instead they would just get the name of the Alias (in your case "Action_0"). The choice here is between an explicit error to indicate a limitation of FASTBuild versus unreliable/broken behaviour the user would have to diagnose. I strongly prefer the former.

The removal of this check is safe in your case only because you happen to not use %1 as an expansion in your arguments, which is not the common use-case for Exec.  You should be able to achieve your desired behaviour by using .PreBuildDependencies instead of .ExecInput to specify these files, and this would require no code changes to FASTBuild. (Since .PreBuildDependencies don't need runtime expansion, they don't perform he problematic safety check).

Ideally the behaviour of Exec should be improved such that expansion of the ExecInput is performed at build time instead of bff parsing time, so it would also support ObjectList in ExecInput. It would then have worked more intuitively in your case as well as being safe in the case where %1 expansion is needed.  This change would involve deleting the safety check as you have done, but also modifying ExecNode::GetInputFiles to handle the expansion of ObjectList.

ffulin:
As noted in the comments, this doesn't safely allow use of ObjectList as an input to Exec, and removes a needed safety check. Safe support of ObjectList on Exec should be possible, but will require more complex changes.

PR #151, opened 2016-09-25
WIP : Appveyor + SonarQube

### LINKED ISSUE #150: Add CI and analysis to the GitHub depot
Hello, I think it will be very cool to have continuous builds with AppVeyor or TravisCI. I I haven't setup something like this but it doesn't seem too complex. I don't know if they offer all environments that are targeted by fastbuild. I can spend some time to investigate if you're interested.

Also, SonarSube analysis is now available for open source projects (https://about.sonarqube.com/news/2016/09/01/sonarqubedotcom-open-to-aynone.html).  In my last merge request (orbis prx) it would have been pretty cool as it should have detected my code duplication and forbade the pull request. That example show that behavior https://github.com/bellingard/multi-language-project/pull/5.

You can add tags on the project page too.

jairbubbles:
I finally managed to get compilation (https://ci.appveyor.com/project/jairbubbles/fastbuild) and analysis (https://sonarqube.com/overview?id=fastbuild) working.

I also put badges on the readme to look modern and cool ;-)

A few notes:
- Clang was failing and I didn't spend time to investigate. It's only x86 / x64 for now.
- Pull request with fast analysis should be working now.  See here for more infos http://docs.sonarqube.org/display/PLUG/GitHub+Plugin
- Switching SDKs is not convenient and I had to update main bff. On a different project I had the issue too and I ended up creating a fbuild_vs2012.bff which was not very pretty. What would be great is to be able to switch between sub scripts that only change SDK, includes, libs and compiler files. These sub scripts could be shared between projects for people who are starting with FB.
- PowerShell is a pain in the ass. There are a lot of weird things and I didn't find a simple tutorial so it was a lot of back and forth.
- AppVeyor is really great. You have a full windows image so you can do everything you want.

You need to setup some en var in the AppVeyor settings :
```
# This assumes that the 2 following variables are defined:
# - SONAR_HOST_URL => should point to the public URL of the SQ server (e.g. : https://sonarqube.com)
# - SONAR_TOKEN    => token of a user who has the "Execute Analysis" permission on the SQ server
# - GITHUB_TOKEN    => token for commenting pull requests in GitHub
```

Closes #150

### DIFF
--- a/SonarQubeAnalysisForAppVeyor.ps1
+++ b/SonarQubeAnalysisForAppVeyor.ps1
@@ -13,7 +13,7 @@ Param(
 if(Test-Path env:APPVEYOR_PULL_REQUEST_NUMBER) {
 	Write-Output 'Pull Request Analysis'
 	Write-Output $env:APPVEYOR_PULL_REQUEST_NUMBER
-	.\SonarQubeAnalysis.ps1 -h $env:SONAR_HOST_URL -l $env:SONAR_TOKEN -s $sources -n $env:APPVEYOR_PROJECT_NAME -k $env:APPVEYOR_PROJECT_SLUG -v $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand $buildWrapperCommand -gitHubPullRequest $env:APPVEYOR_PULL_REQUEST_NUMBER -gitHubOauth $GITHUB_TOKEN -gitHubRepository $env:APPVEYOR_REPO_NAME
+	.\SonarQubeAnalysis.ps1 -h $env:SONAR_HOST_URL -l $env:SONAR_TOKEN -s $sources -n $env:APPVEYOR_PROJECT_NAME -k $env:APPVEYOR_PROJECT_SLUG -v $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand $buildWrapperCommand -gitHubPullRequest $env:APPVEYOR_PULL_REQUEST_NUMBER -gitHubOauth $env:GITHUB_TOKEN -gitHubRepository $env:APPVEYOR_REPO_NAME
 }
 # Full analysis for master branch only
 ElseIf($env:APPVEYOR_REPO_BRANCH -eq 'master') {
--- a/SonarQubeAnalysisForAppVeyor.ps1
+++ b/SonarQubeAnalysisForAppVeyor.ps1
@@ -1,0 +1,23 @@
+# This assumes that the 2 following variables are defined:
+# - SONAR_HOST_URL => should point to the public URL of the SQ server (e.g. : https://sonarqube.com)
+# - SONAR_TOKEN    => token of a user who has the "Execute Analysis" permission on the SQ server
+# - GITHUB_TOKEN    => token for commenting pull requests in GitHub
+
+Param(
+	[parameter(Mandatory=$true)]
+	[string]$sources,
+	[string]$buildWrapperCommand	
+)
+
+# Special behavior for pull requests
+if(Test-Path env:APPVEYOR_PULL_REQUEST_NUMBER) {
+	Write-Output 'Pull Request Analysis'
+	Write-Output $env:APPVEYOR_PULL_REQUEST_NUMBER
+	.\SonarQubeAnalysis.ps1 -h $env:SONAR_HOST_URL -l $env:SONAR_TOKEN -s $sources -n $env:APPVEYOR_PROJECT_NAME -k $env:APPVEYOR_PROJECT_SLUG -v $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand $buildWrapperCommand -gitHubPullRequest $env:APPVEYOR_PULL_REQUEST_NUMBER -gitHubOauth $GITHUB_TOKEN -gitHubRepository $env:APPVEYOR_REPO_NAME
+}
+# Full analysis for master branch only
+ElseIf($env:APPVEYOR_REPO_BRANCH -eq 'master') {
+	Write-Output 'Full Analysis'
+	.\SonarQubeAnalysis.ps1 -h $env:SONAR_HOST_URL -l $env:SONAR_TOKEN -s $sources -n $env:APPVEYOR_PROJECT_NAME -k $env:APPVEYOR_PROJECT_SLUG -v $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand $buildWrapperCommand
+}
+else {Write-Output 'Skipping Analysis'}
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -2,12 +2,10 @@ before_build:
 - ps: (new-object net.webclient).DownloadFile('http://fastbuild.org/downloads/v0.91/FASTBuild-Windows-x64-v0.91.zip', 'FastBuild.zip')
 - ps: Add-Type -assembly “system.io.compression.filesystem”
 - ps: "[io.compression.zipfile]::ExtractToDirectory('FastBuild.zip', $env:APPVEYOR_BUILD_FOLDER)"
-- ps: Write-Output $env:APPVEYOR_BUILD_FOLDER
-- ps: dir
  
 build_script:
-- ps: .\FastBuild\fbuild.exe -config Code\fbuild.bff
+- ps: .\FBuild.exe -config Code\fbuild.bff
   
 after_build:
-- ps: .\FastBuild\fbuild.exe -config Code\fbuild.bff solution
+- ps: .\FBuild.exe -config Code\fbuild.bff solution
 - ps: .\SonarQubeAnalysis.ps1 -hostUrl 'https://sonarqube.com' -login  d5fcf62e7cba7c67b913e9edb4ecee2e9a697d52 -sources Code/ -projectName $env:APPVEYOR_PROJECT_NAME -projectKey $env:APPVEYOR_PROJECT_SLUG -projectVersion $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand "msbuild  tmp\VisualStudio\FASTBuild.sln  /t:rebuild"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,11 +1,11 @@
 before_build:
 - ps: (new-object net.webclient).DownloadFile('http://fastbuild.org/downloads/v0.91/FASTBuild-Windows-x64-v0.91.zip', 'FastBuild.zip')
 - ps: Add-Type -assembly “system.io.compression.filesystem”
-- ps: "[io.compression.zipfile]::ExtractToDirectory('FastBuild.zip', $env:APPVEYOR_BUILD_FOLDER +'/')"
+- ps: "[io.compression.zipfile]::ExtractToDirectory('FastBuild.zip', $env:APPVEYOR_BUILD_FOLDER)"
  
 build_script:
-- ps: fbuild.exe -config Code\fbuild.bff
+- ps: FastBuild\fbuild.exe -config Code\fbuild.bff
   
 after_build:
-- ps: fbuild.exe -config Code\fbuild.bff solution
+- ps: FastBuild\fbuild.exe -config Code\fbuild.bff solution
 - ps: .\SonarQubeAnalysis.ps1 -hostUrl 'https://sonarqube.com' -login  d5fcf62e7cba7c67b913e9edb4ecee2e9a697d52 -sources Code/ -projectName $env:APPVEYOR_PROJECT_NAME -projectKey $env:APPVEYOR_PROJECT_SLUG -projectVersion $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand "msbuild  tmp\VisualStudio\FASTBuild.sln  /t:rebuild"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -6,11 +6,7 @@ before_build:
 - ps: "[io.compression.zipfile]::ExtractToDirectory('FastBuild.zip', $env:APPVEYOR_BUILD_FOLDER)"
 - ps: .\FBuild.exe -config Code\fbuild.bff solution
 - ps: $env:Path = $env:Path + $env:APPVEYOR_BUILD_FOLDER
-- ps: dir 'C:\Program Files (x86)\Microsoft SDKs\Windows'
-- ps: dir 'C:\Program Files (x86)\Microsoft SDKs\Windows\v8.0'
-- ps: dir 'C:\Program Files (x86)\Microsoft SDKs\Windows\v8.0A'
-- ps: dir 'C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1'
-- ps: dir 'C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1A'
+- ps: dir 'C:\Program Files (x86)\Microsoft SDKs\Windows\v8.1A\bin'
  
 # Using FastBuild
 build_script:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -3,7 +3,7 @@ before_build:
 - ps: Add-Type -assembly “system.io.compression.filesystem”
 - ps: "[io.compression.zipfile]::ExtractToDirectory('FastBuild.zip', $env:APPVEYOR_BUILD_FOLDER +'/FastBuild')"
  
-build:
+build_script:
 - ps: fbuild.exe -config Code\fbuild.bff
   
 after_build:
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -24,4 +24,5 @@ build_script:
   
 after_build:
 #Launch analysis
+- ps: cd ..
 - ps: .\SonarQubeAnalysis.ps1 -hostUrl 'https://sonarqube.com' -login  d5fcf62e7cba7c67b913e9edb4ecee2e9a697d52 -sources Code/ -projectName $env:APPVEYOR_PROJECT_NAME -projectKey $env:APPVEYOR_PROJECT_SLUG -projectVersion $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand "msbuild  '../tmp/VisualStudio/FASTBuild.sln'  /t:rebuild"
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -10,7 +10,7 @@ before_build:
 # Generate VS projects
 - ps: .\FBuild.exe -config Code\fbuild.bff solution
 # Add current location to path
-- ps: $env:Path = $env:Path + $env:APPVEYOR_BUILD_FOLDER
+- ps: "$env:Path = $env:Path + $env:APPVEYOR_BUILD_FOLDER"
  
 # Using FastBuild
 build_script:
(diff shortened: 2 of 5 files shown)


ffulin:
Looks good. Thanks again for looking into this. I will look into making some changes to the bff syntax so that we can auto-detect the compiler location and then I should be able to integrate this.

jairbubbles:
No problem. I wanted to look into this for some time. It was a good excuse ;-)

About compiler discovery. Wil you be able to choose if you have different visual studio versions on your computer ?

jairbubbles:
I created a depot with the PowerShell scripts https://github.com/jairbubbles/sonar-powershell. It might be a good idea to make it a sub module for the fastbuild depot.

ffulin:
> About compiler discovery. Wil you be able to choose if you have different visual studio versions on your computer ?

I think so.  I'm thinking of adding a "#if fileexists" and #error" mechanism which would allow for explicitly vendorized versions (what I use), falling back to looking at whatever might be installed (which most people seem to use):

```C++
// Try to use vendored VS 2015
.CompilerLocation = 'External/SDK/VS2015.2/VC/bin/cl.exe'
#if fileexists( .CompilerLocation )
    #define USE_VS2015
#else
    // Try to use vendored VS 2013
    .CompilerLocation = 'External/SDK/VS13.4/VC/bin/cl.exe'
    #if fileexists( .CompilerLocation )
        #define USE_VS2013
    #else
        // Try to use installed 2015
        .CompilerLocation = 'C:/Program Files (x86)/Microsoft Visual Studio 14.0/VC/bin/cl.exe' 
        #if fileexists( .CompilerLocation )
            #define USE_VS2015
        #else
            // Try to use installed 2013
            .CompilerLocation = 'C:/Program Files (x86)/Microsoft Visual Studio 12.0/VC/bin/cl.exe' 
            #if fileexists( .CompilerLocation )
                #define USE_VS2013
            #else
                #error "Could not find VS 2013 or 2015 compiler!"
            #endif
        #endif
    #endif
#endif    
```

That would automatically find the "best" version to use.  With some more complicated #if logic, we could also support explicit definition of the compiler (if you had 2015 installed but wanted to force 2013 for example.

jairbubbles:
Why don't juste use %VCINSTALLDIR% and %WindowsSdkDir% by default ? There are a bunch of over env var set by vcvars64.bat and its variants.

You'll still have to make some if logic to select the compiler files but when investigating bugs with msfastbuild it seems highly complex to do without actual code (because of localized resource dll)..

For custom path you'll juste have to set the env vars before calling fbuild.exe.

ffulin:
> Why don't juste use %VCINSTALLDIR% and %WindowsSdkDir% by default ? 

I don't want the bff for FASTBuild itself to be dependent on the Visual Studio installation in any way, including a dependency on the environment variables. I vendorize the toolchain for FASTBuild (and all other projects I work on) as it avoids a lot of serious stability and maintenance issues. The logic above is a compromise to make it easier for users of FASTBuild to more easily compile it themselves as well as for us to have it on AppVeyor, where we can't easily vendorize the compiler due to licensing issues.

> You'll still have to make some if logic to select the compiler files but when investigating bugs with msfastbuild...

This doesn't need to impact msfastbuild.  This is purely to allow the compilation of FASTBuild itself to support the explicit vendorized compilers I use.  msfastbuild can generate the bff as it sees fit, including using/importing the env vars.

I would always recommend to users that they look to switch to a vendorized toolchain approach though as it avoids a lot of development pitfalls. I can go on at length about this if you want :)

> ....investigating bugs with msfastbuild it seems highly complex to do without actual code (because of localized resource dll)". 

I'm not quite sure what you mean by that. Can you elaborate please?

jairbubbles:
I've used both ways for SDKs so I understand pros and cons. Anyway I do think fastbuild should compile out of the box just to make it more accessible.

I understand you don't want to use VC env var but we could imagine giving the arguments though fbuild command line  in a generic way, something like :
fbuild.exe /p VSBasePath="C:\Program Files (x86)\Microsoft Visual Studio 12.0"
It opens up to a lot of scenarios.

For the resource dll issue you can look at this issue: https://github.com/liamkf/msfastbuild/issues/2
Bascially it didn't work for french installation as clui.dll was in 1036 sub directory and not 1033. 
After some experimentations I've found out that this library was loaded dynamically by cl.exe by parsing the subfolders so we needed to do something similar to make it work.

ffulin:
> Anyway I do think fastbuild should compile out of the box just to make it more accessible.
Definitely agreed. Reducing friction for users is more important than ideology in this case :)

> imagine giving the arguments though fbuild command line in a generic way, 
I think that makes it harder to use for the user than what I've proposed. It also leaves me with the problem of not knowing which version of MSVC that is, since it's not just a path that's needed in the bff, but rather a bunch of options changes to compiler and linker options (particularly with the change from VS2013 to 2015 as the C/C++ sdk has been completely reworked).

> For the resource dll issue you can look at this issue: liamkf/msfastbuild#2
I understand what you mean now (thanks for the link). The dll management of VS is indeed painful because of the dynamic loading, the various directories they scatter their dlls in and the use of the registry.  It's particularly painful since (without virtualizing the registry on the remote machine) there isn't a good way to get the correct behaviour in all cases.  Thankfully it's not too problematic in practice, since a studio will tend to all be on the same language, and most engineers want their compiler errors in English since it's the only useful one to search the internet with.
Using environment variables doesn't really help with this problem, since it's the remote machine that is problematic, and you can't use the remote environment at all since it's not consistent.

jairbubbles:
> and most engineers want their compiler errors in English since it's the only useful one to search the internet with.

It was indeed an error if my Visual installation was in French but I can"t speak for everyone :-D For instance I'm pretty sure Japanese like their Visual Studio in japanese!

I'm ok if the auto detection works well and if we can overload it easily (a log would be nice to say which version it is using). It's more a philosophical subject on what should the build system do or not do. Except for .net applications you can't dynamically change SDK / compiler with MsBuild. 

Do you plan to ship that section of the bff which auto detects SDK? I like the idea that it's working out of the box but you can easily configure it for your own needs as it's not in the binary executable.

fitzymj:
Not to butt in, but since you offered, I'd love to hear more about your vendorized toolchain approach.  Any best practices for doing it that way, instead of relying on local installation?  Since I'm aware this thread is probably not the best place, even some external links about it would be interesting.  Thanks!

appveyor.yml
```diff
@@ -0,0 +1,28 @@
...
+- ps: Add-Type -assembly "system.io.compression.filesystem"
+- ps: "[io.compression.zipfile]::ExtractToDirectory('C:/FastBuild.zip', 'C:/FastBuild')"
+# Add FastBuild to path
+- ps: "$env:Path = $env:Path + ';C:/FastBuild'"
+# Generate VS projects
+- ps: fbuild solution
+ 
+# Using FastBuild
+build_script:
+- ps: fbuild All-x86 All-x64
```
ffulin: as above, we should add the -noprogress option here to de-spam the logs a bit

appveyor.yml
```diff
@@ -0,0 +1,28 @@
...
+
+# Using MSbuild
+#build:
+#  project: C:\projects\tmp\VisualStudio\FASTBuild.sln
+#  verbosity: minimal
+  
+after_build:
+#Launch analysis
+- ps: cd ..
+- ps:  ./SonarQubeAnalysisForAppVeyor.ps1 -sources Code/ -buildWrapperCommand "msbuild  'tmp/VisualStudio/FASTBuild.sln'  /t:rebuild"
```
ffulin: Does SonarQube support invoking FASTBuild directly?

appveyor.yml
```diff
@@ -0,0 +1,28 @@
...
+- ps: cd Code
+- ps: "[Environment]::CurrentDirectory = $PWD"
+# Download & Unzip FastBuild
+- ps: (new-object net.webclient).DownloadFile('http://fastbuild.org/downloads/v0.91/FASTBuild-Windows-x64-v0.91.zip', 'C:/FastBuild.zip')
+- ps: Add-Type -assembly "system.io.compression.filesystem"
+- ps: "[io.compression.zipfile]::ExtractToDirectory('C:/FastBuild.zip', 'C:/FastBuild')"
+# Add FastBuild to path
+- ps: "$env:Path = $env:Path + ';C:/FastBuild'"
+# Generate VS projects
+- ps: fbuild solution
```
ffulin: we should add the -noprogress option here to de-spam the logs a bit

appveyor.yml
```diff
@@ -0,0 +1,28 @@
+os: Visual Studio 2013
+
+before_build:
+# Change working directoty and current location
```
ffulin: typo in comment "directoty"

SonarQubeAnalysisForAppVeyor.ps1
```diff
@@ -0,0 +1,23 @@
...
+	[string]$buildWrapperCommand	
+)
+
+# Special behavior for pull requests
+if(Test-Path env:APPVEYOR_PULL_REQUEST_NUMBER) {
+	Write-Output 'Pull Request Analysis'
+	Write-Output $env:APPVEYOR_PULL_REQUEST_NUMBER
+	.\SonarQubeAnalysis.ps1 -h $env:SONAR_HOST_URL -l $env:SONAR_TOKEN -s $sources -n $env:APPVEYOR_PROJECT_NAME -k $env:APPVEYOR_PROJECT_SLUG -v $env:APPVEYOR_BUILD_NUMBER -buildWrapperCommand $buildWrapperCommand -gitHubPullRequest $env:APPVEYOR_PULL_REQUEST_NUMBER -gitHubOauth $GITHUB_TOKEN -gitHubRepository $env:APPVEYOR_REPO_NAME
+}
+# Full analysis for master branch only
```
ffulin: Seems like it might be more valuable to have the analysis on the dev branch?
Analysis on both would be ideal I guess.

ffulin:
@fitzymj : I think opening a new "issue" for FASTBuild would be good.  I can provide some details there in addition to how it relates to FASTBuild.  That could be nice for future users.

appveyor.yml
ffulin: I'll fix that in the integration

appveyor.yml
jairbubbles: Totally !

appveyor.yml
jairbubbles: Didn't test yet. I don't know what kind of output the build wrapper needs. But I guess with -IDE it should work.

SonarQubeAnalysisForAppVeyor.ps1
jairbubbles: On fastbuild depot it seems so yes. SonarQube branch management is not ideal for now so we're better with one. Anyway this a first step for an awesome build pipeline. A next step will be automatic release packaging through appveyor.

ffulin:
I decided to integrate this into dev as-is for the moment while I continue to work on the improvements to the bff syntax that I mentioned above.

ffulin:
> On fastbuild depot it seems so yes. SonarQube branch management is not ideal for now so we're better with one. 

What do we need to do to switch over to CI the dev branch instead?  Do you know if it's possible for us to cover both branches?  That would be really nice.

> A next step will be automatic release packaging through appveyor.

For the official release I don't want to do this. The official release is built and tested with a very specific compiler version (VS2013 Update 4 on Windows for example), so I'd like to keep those executable as the officially released ones.

For the dev branch on the other hand, it would be very nice to be able to get the "experimental" exe built automatically based on whatever the latest state of is.

jairbubbles:
> What do we need to do to switch over to CI the dev branch instead? Do you know if it's possible for us to cover both branches? That would be really nice.

You just need to update line 19 in SonarQubeAnalysisForAppVeyor.ps1:
```
# Full analysis for master branch only
ElseIf($env:APPVEYOR_REPO_BRANCH -eq 'master') {
```

There is no limitation to the branches you can analyse but right now all analysis will be uploaded on the same SonarQube project so the history will be fucked  up. I'll look more into the branch system on SonarQube but I've read it's not good and this is one of the next thing they're going to fix.

> 
For the official release I don't want to do this. The official release is built and tested with a very specific compiler version (VS2013 Update 4 on Windows for example), so I'd like to keep those executable as the officially released ones.

I'm sure you have obscure reasons for this :-p. AppVeyor allows to use "pre" version of images but I don't think you'll have an old updated VS versions on it.

> For the dev branch on the other hand, it would be very nice to be able to get the "experimental" exe built automatically based on whatever the latest state of is.

I can look into it. We can upload zip on GitHub, add a Nuget or chocolatey package. But it seems more suited for official releases. 

For instance for cmake:
https://chocolatey.org/packages/cmake
https://www.nuget.org/packages/CMake/
https://github.com/Kitware/CMake/releases

jairbubbles:
The main problem with SonarQube branch:
	
> Manage SCM branches. (avertissement) Two branches of the same project are considered to be different projects in SonarQube. As a consequence issues found in a project A in a branch B1 are not linked to issues found for this project A in a branch B2. Currently, there is no way to resolve automatically issues of B2 when they are resolved in B1 as again A-B1 & A-B2 are considered as separated project.

So you might end up doing things in double on SonarQube interface. We might try to do a preview analysis on dev branch but we would need a way to display the results (GitHub plugin only comment pull requests for now),

jairbubbles:
(I think you should have squashed my commits when accepting my pull request)

ffulin:
> I'm sure you have obscure reasons for this :-p. AppVeyor allows to use "pre" version of images but I don't think you'll have an old updated VS versions on it.

The reason is not obscure. It's very simple.  I want to ship to my users the most robust version of the thing I'm working one.  That means I want to ship the thing I've actually tested.

You might say this is a theoretical concern, but literally right now another project I am involved with which is using VS2015, cannot use Update 3 because of a code gen bug specific to that version.  (The bug is not present in Update 2).  In this case the bug was severe enough that it causes an alignment related crash, so it was discovered really quickly.  But what about if the bug was more subtle?  I don't want to develop and test on one thing, but release something else to my users.

Another thing to note is that the GitHub source for FASTBuild is a mirror for the official mainline, which is kept in Perforce.  All released versions are built from Perforce. I'm open to having it possible to download the exes from more places to make it more convenient, but If you get v0.92 from two different places, it should be the binary identical executable, and thus I would push new versions from the Perforce build process rather than the GitHub process.

> (I think you should have squashed my commits when accepting my pull request)

I wasn't aware of a way to do this that would still function as accepting your pull request.  Is there a way to do this?

Sadly, I also didn't notice that your pull request was against the main branch (pull requests are only accepted to "dev") and there is no (good) way (that I know of) to undo that.  :(

> The main problem with SonarQube branch:...

Yeah, that's unfortunate, though I can see why that's a non-trivial problem for them to solve.  I'll use your instructions to try and switch the analysis to the "dev" branch as that's where we'll get the most value.

jairbubbles:
I wasn't aware that vs2013 update 5 had that kind of issues. Anyway I was just being ironic.
Do you have more info on the bug with VS2015 update 3 ? We've had an issue too but it was related to SIMD (@onduril can tell you more about it).
I'll contact AppVeyor to see if we can have more control on the toolchain.

I'm sure you can do something about the git commit but I don't know how to do it :-S (something with rebase and force commit ^^).
Sorry about the pull request on the bad branch. It seems like we are two Git noobs here.

Regarding squash and merge it's in the interface:
https://help.github.com/articles/merging-a-pull-request/
(no need to invoke some voodoo git command line)

Regarding tokens on AppVeyor and SonarQube did you set up yours ? The link on the readme is still on the jairbubbles one.

ffulin:
> Do you have more info on the bug with VS2015 update 3 ?

I believe we raised an issue with MS about it.  I'll try to get you the link.  If I recall it was related to unsafe mixing of SIMD and non-SIMD instructions on an unaligned type.  It seems to be related to the [new optimizing back-end](https://blogs.msdn.microsoft.com/vcblog/2016/05/04/new-code-optimizer/) they added in Update 3 (but isn't fixed by disabling that with their "-d2SSAOptimizer-" flag).

> It seems like we are two Git noobs here.
Lol.  I certainly feel like a noob a lot of the time :)

> Regarding tokens on AppVeyor and SonarQube did you set up yours ? The link on the readme is still on the jairbubbles one.
I'm cleaning that stuff up now.
I am migrating the config files into the "dev" branch and I'm working to switch it to use the AppVeyor project i setup (https://ci.appveyor.com/project/ffulin/fastbuild)

ffulin:
Here's the link to the VS2015 Update 3 code-gen bug.  Seems it's fixed in VS2016 at least.

https://connect.microsoft.com/VisualStudio/feedback/details/3041794/visual-c-codegen-problem-movaps-instruction-to-unaligned-temporary

Onduril:
FYI:

https://connect.microsoft.com/VisualStudio/feedback/details/3001148/wrong-code-generation-for-mm-insert-ps-x64-optimized

jairbubbles:
I've played a little bit with artifacts and deployment in AppVeyor.

In after_build you just need to create zip files:
```
- ps: 7z a FASTBuild-Windows-x64.zip .\Code\LICENSE.TXT .\tmp\x64-Release\Tools\FBuild\FBuildApp\FBuild.exe .\tmp\x64-Release\Tools\FBuild\FBuildWorker\FBuildWorker.exe
- ps: 7z a FASTBuild-Windows-x86.zip .\Code\LICENSE.TXT .\tmp\x86-Release\Tools\FBuild\FBuildApp\FBuild.exe .\tmp\x86-Release\Tools\FBuild\FBuildWorker\FBuildWorker.exe
```

And describe your build artifacts:
```
artifacts:
- path: FASTBuild-Windows-x64.zip
  name: FASTBuild-Windows-x64
- path: FASTBuild-Windows-x86.zip
  name: FASTBuild-Windows-x86
```

Artifacts are then available through a special URL.
For instance :
https://ci.appveyor.com/api/projects/jairbubbles/fastbuild/artifacts/FASTBuild-Windows-x86.zip
Or https://ci.appveyor.com/api/projects/jairbubbles/fastbuild/artifacts/FASTBuild-Windows-x86.zip?branch=dev if you want only the dev branch.

(you can look at https://github.com/jairbubbles/fastbuild/blob/master/appveyor.yml)

ffulin:
So I've spent some time clearing out some of the issues identified by the analyzer (no bugs discovered, but nice improvements to overall code quality).  I've got a couple of issues remaining though:

*I can't work out how to disable a rule* - I feel like I must be stupid, but I can't work out how to disable some of the rules. From what I've discovered, I need to create a new Quality Profile, but I can't find anywhere in the sonarqube.com web UI to do that. I can view all the existing profiles, but I can't make any changes. Am I going about this wonrg? Is this something we should instead be defining in the .ps1 config file in GitHub?  All the stuff I've found on the web seems to relate to running your own SonarQube server and thus is of no help.

*I have to manually trigger a build* - Submits to GitHub don't seem to trigger any builds.  I always have to trigger a build manually on AppVeyor.

jairbubbles:
Hello,

I've been using SonarQube at work on an internal server and creating quality profiles was really easy. It's possible that you don't have such possibilities with the free plan on SonarQube.com. Please note that you can bulk close issues easily. 

For the triggers on AppVeyor it's weird as I didn't have to do anything to make it work. Maybe it's an issue regarding fastbuild vs ffulin depot ?

PR #163, opened 2016-11-07
Added SimpleDistributionMode, CustomEnvironmentVariables, and CompilerRoot

gkoreman:
These changes support distributing a custom toolchain and running arbitrary executables. This is useful when converting textures or building shaders.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -334,16 +334,16 @@ Node::BuildResult ObjectNode::DoBuildWithPreProcessor( Job * job, bool useDeopti
 
 	if (pass == PASS_PREPROCESSOR_ONLY)
 	{
-    if ( BuildPreprocessedOutput( fullArgs, job, useDeoptimization ) == false )
-    {
-        return NODE_RESULT_FAILED; // BuildPreprocessedOutput will have emitted an error
-    }
+		if ( BuildPreprocessedOutput( fullArgs, job, useDeoptimization ) == false )
+		{
+			return NODE_RESULT_FAILED; // BuildPreprocessedOutput will have emitted an error
+		}
 
-    // preprocessed ok, try to extract includes
-    if ( ProcessIncludesWithPreProcessor( job ) == false )
-    {
-        return NODE_RESULT_FAILED; // ProcessIncludesWithPreProcessor will have emitted an error
-    }
+		// preprocessed ok, try to extract includes
+		if ( ProcessIncludesWithPreProcessor( job ) == false )
+		{
+			return NODE_RESULT_FAILED; // ProcessIncludesWithPreProcessor will have emitted an error
+		}
 	}
 
 	if (pass == PASS_PREP_FOR_SIMPLE_DISTRIBUTION)
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -196,7 +196,8 @@ ObjectNode::~ObjectNode()
 
     bool useCache = ShouldUseCache();
     bool useDist = GetFlag( FLAG_CAN_BE_DISTRIBUTED ) && m_AllowDistribution && FBuild::Get().GetOptions().m_AllowDistributed;
-    bool usePreProcessor = ( useCache || useDist || GetFlag( FLAG_GCC ) || GetFlag( FLAG_SNC ) || GetFlag( FLAG_CLANG ) || GetFlag( CODEWARRIOR_WII ) || GetFlag( GREENHILLS_WIIU ) );
+	bool useSimpleDist = GetCompiler()->CastTo< CompilerNode >()->SimpleDistributionMode();
+	bool usePreProcessor = !useSimpleDist && ( useCache || useDist || GetFlag( FLAG_GCC ) || GetFlag( FLAG_SNC ) || GetFlag( FLAG_CLANG ) || GetFlag( CODEWARRIOR_WII ) || GetFlag( GREENHILLS_WIIU ) );
     if ( GetDedicatedPreprocessor() )
     {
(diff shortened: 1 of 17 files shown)


ffulin:
Thanks for the Pull Request. This change looks good overall, but there are a few issues we'll need to resolve before integration. Additionally, I'm in the process of stabilizing the v0.92 release, so with that in mind, I'll integrate this after v0.92 is released (so all things going well, these changes will be in the v0.93 release). I'll add comments to the review shortly for the issues we need to resolve.

gkoreman:
Sounds good to me.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -1728,7 +1740,36 @@ bool ObjectNode::BuildPreprocessedOutput( const Args & fullArgs, Job * job, bool
...
-    return true;
+	return true;
+}
+
+// PreProcessing for SimpleDistribution is just loading the source file
+//------------------------------------------------------------------------------
+bool ObjectNode::LoadStaticSourceFileForDistribution(const Args & fullArgs, Job * job, bool useDeoptimization) const
+{
+	const bool useDedicatedPreprocessor = (GetDedicatedPreprocessor() != nullptr);
+	EmitCompilationMessage(fullArgs, useDeoptimization, false, false, useDedicatedPreprocessor);
```
ffulin: I think the dedicated preprocessor can never be used, since we're skipping preprocessing. I think we can assume false here, or better still, emit an error during bff parsing if the option is set. What do you think?

Code/Tools/FBuild/FBuildCore/Helpers/ToolManifest.cpp
```diff
@@ -117,6 +120,11 @@ bool ToolManifest::Generate( const Node * mainExecutable, const Dependencies & d
         *pos = xxHash::Calc32( relativePath );
         ++pos;
     }
+	for (size_t i = 0; i < numEnvVars; ++i)
+	{
+		*pos = xxHash::Calc32(m_CustomEnvironmentVariables[i]);
+		++pos;
+	}
```
ffulin: I don't think we need to include the custom environment variables in the hash, since they don't impact the layout of the directory at all.  The env vars should be correctly set for each remote execution, even if two different builders use the same toolchain.
Is there an edge case I'm missing here?

Code/Tools/FBuild/FBuildWorker/FBuildWorkerOptions.cpp
```diff
@@ -19,18 +19,50 @@
...
+	: m_OverrideCPUAllocationWhenIdle( false )
+    , m_CPUAllocationWhenIdle( 0 )
+	, m_OverrideCPUAllocationDedicated(false)
+	, m_CPUAllocationDedicated(0)
 #if defined( __WINDOWS__ )
-    m_IsSubprocess( false ),
-    m_UseSubprocess( true ),
+	, m_IsSubprocess(false)
+	, m_UseSubprocess(true)
 #endif
```
ffulin: These properties were ordered to match the header, which is needed due to stricter warnings with Clang.

Code/Tools/FBuild/FBuildWorker/FBuildWorkerOptions.cpp
```diff
@@ -44,53 +76,20 @@ bool FBuildWorkerOptions::ProcessCommandLine( const AString & commandLine )
...
-            // problem... fall through
-        }
-        else if ( token == "-mode=disabled" )
-        {
-            m_WorkMode = WorkerSettings::DISABLED;
-            m_OverrideWorkMode = true;
-            continue;
-        }
-        else if ( token == "-mode=idle" )
+        if ( token.BeginsWith( "-cpusWhenIdle=" ) )
```
ffulin: Some users launch the FBuildWorker with a script or through other automation. It would be nice if we could extend the command line options while retaining backwards compatibility.
Perhaps we can keep -cpus as is, while adding -cpusWhenIdle?

Code/Tools/FBuild/FBuildWorker/Worker/WorkerSettings.cpp
```diff
@@ -82,10 +82,8 @@ void WorkerSettings::Load()
         }
 
         // settings
-        uint32_t mode;
-        f.Read( mode );
-        m_Mode = (Mode)mode;
-        f.Read( m_NumCPUsToUse );
+        f.Read( m_NumCPUsToUseWhenIdle );
+		f.Read(m_NumCPUsToUseDedicated);
```
ffulin: This looks like a bug. We've set the versions such that we'll load the old config, but we aren't correctly handling the old format.

ffulin:
Sorry it took so long, but I've put up the comments about things I think need minor modifications.  (Also for some reason I added half as comments and half as a review, so sorry for the wackyness)

How would you feel about splitting the "simple distribution" from the worker cpu changes (into two reviews)?  That would make it a little easier to manage the integration (since the simple distribution seems mostly good to go, while the worker changes have some backwards compat issues and a bug)

Code/Tools/FBuild/FBuildCore/Helpers/ToolManifest.cpp
gkoreman: Yes, you're correct, and removing them from the hash would allow much better toolchain re-use.

gkoreman:
I actually did not intend to submit a pull request for the worker CPU changes, but since I didn't branch it's all in one now. I see if I can do some git magic to pull it all apart if you like. There's a bug "fix" in here now too.

ffulin:
> I can see if I can do some git magic to pull it all apart if you like. There's a bug "fix" in here now too.

I know it's a bit of a hassle, but I think if we could split the changes apart into their functional parts (I guess 3 of them now), that would make it a lot more manageable. The changes mostly seem isolated, so cut and pasting of files into new git branches would do the trick if your git-fu is weak like mine :)

gkoreman:
I split this into #184 #185 #186.

PR #164, opened 2016-11-08
Fix AString::Replace taking extremly long time when "from" is an empty string

dummyunit:
When `AString::Replace` is called with empty `from` string, it never
advances `pos` pointer in the main loop and exits only when `maxReplaces`
reaches 0, which takes 2^32 iterations.
This problem can be observed with this input (parsing takes very long
time):

    .X = 'a'
    .X - ''

This patch adds early exit to catch this case.


ffulin:
Ouch.  I've confirmed the bug: on mp PC, each call costs around 4 seconds! This fix will be in the next release (v0.92).  Thanks for the fix!

PR #168, opened 2016-11-24
Use native slash in cache paths

dummyunit:
This re-enables grouping of cache files into separate directories on
Linux and OS X.
This also allows storage of cache files on filesystems that don't allow
\ as part of a file name. For example, I encountered this issue while trying to set up mounted VirtualBox shared folder (guest=Linux, host=Windows7) as FASTBUILD_CACHE_PATH (files weren't getting created).

Also, a slightly unrelated question: On Linux I am getting cache files with names like `EB65EB0AD97484EC_824A3E5A_30FBF3C59D93AF21-0000000000000000.7`, what `-0000000000000000` is supposed to encode? That part is always all zeroes, which doesn't look correct, also cache files on Windows doesn't have it.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/Cache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/Cache.cpp
@@ -143,11 +143,13 @@
 void Cache::GetCacheFileName( const AString & cacheId, AString & path ) const
 {
     // format example: N:\\fbuild.cache\\23\\77\\2377DE32_FED872A1_AB62FEAA23498AAC.3
-    path.Format( "%s%c%c\\%c%c\\%s", m_CachePath.Get(),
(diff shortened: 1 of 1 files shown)


ffulin:
Looks like a solid fix - thanks.

The final part of the cache key is to allow safe caching of precompiled headers and objects that use them when using the MS compiler. The MS compiler has a hard dependency between the specific precompiled header and an object file using it. It is normal for this to be 0 in other cases.

dummyunit:
Thanks for your explanation, now all that makes sense. I don't use precompiled headers and apparently was using older FASTBuild version on Windows (cache version 6), which doesn't have that part.

ffulin:
I've integrated this fix into the Perforce mainline. This fix will be in the next release (v0.93). Thankyou for your contribution!

Issue #82, opened 2016-01-03
default value for #import in case the envar doesn't exist


shooshx:
Using #import, if the given environment variable doesn't exist, the parsing of the .bff fails. this makes #import less useful than it could have been since if I use it anywhere, I now have to make sure this variable is always defined and I can't even do a simple ./FBuild in the .bff folder. 
A reasonable solution is to add a second space delimited, optional argument to #import that would be the default value, in case the given name doesn't exist. this is similar to python's dict.get()
**pro**: gives maximum flexibility
**con**: some might find it a weird syntax 
Another alternative is that #import just sets the variable to an empty string, 
**pro**: no syntax change, simpler concept
**con**: less powerful 
**con**: break backward compatibility as far as error handling is concerned.
let me know what you prefer and I'll code it.

kovarex:
I would also like to support it. The first variant.

ffulin:
I've been thinking about this, and I think adding an optional keyword as follows would be the way to go:
```C++
#import optional ENV_VAR
```
**If ENV_VAR exists**, the behaviour is the same as without the optional keyword.

**If ENV_VAR does not exist:**
 - if previously declared in the bff, it would remain unchanged (so you can have a default value overridden by an ENV_VAR)
 - if not previously declared, it would create and empty string

What do you think?

shooshx:
That sounds perfect.
Hopefully in the next few days I'll have time to be working on both this and the -define issue

kovarex:
This might be different topic, but it would be nice to be able to make conditional code depending on the import value. For example, I have visual studio 2013 and 2015 configurations in the build file, 
Visual studio 2015 configuration uses this import:
#import VS140COMNTOOLS
While visual studio 2013 configuration uses this import:
#import VS120COMNTOOLS
The default/optional value will solve the problem that it didn't work when just one of the versions was installed, but it still won't solve, that the other configuration will still be available. If I could define preprocessor value depending on the existance of environment variable, I can use it to control what configurations to define.

Manuzor:
:+1:
I'm looking for the same thing as @kovarex. It would be very useful to make some environment variables optional, possibly even having the ability to react on whether they exist or not.

I'm thinking
```
#if !VS140COMNTOOLS
#if !VS120COMNTOOLS
"Unable to find VS2015 or VS2013"
#endif
#endif
```

ffulin:
I've been thinking about this, and I think the following syntax would likely be the way to go:
```C++
[...]

PR #178, opened 2016-12-08
Added #if defined(ENV_VAR) support

jblazquez:
Added #if defined(ENV_VAR) support as discussed in #82.


ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.93).

Thankyou for your contribution!

Issue #105, opened 2016-06-13
"D8049: command line is too long to fit in debug record" when building UE4


arturmazurek:
I am building UE4  using the article from the wiki here. Sometimes when using remote workers the build fails with the following error:

> cl: Command line error D8049 : cannot execute 'C:\tmp[...]\c1xx.dll': command line is too long to fit in debug record

I have observed the following:
- it happens only when cl.exe process is spawned on remote workers, when the file is compiled on the same machine where the build was initiated it's ok
- it happens when the arguments count sent to cl.exe exceeds around 20k characters
- when I execute CreateProcess in my own test aplication with the same compilation command it fails with the same error (which is very odd, because as I write above, it works fine from Fastbuild)
- using a response file doesn't help

I'm attaching the command I used to reproduce it in my test application. It's actually saved from Process.cpp (`fullArgs` parameter) in Fastbuild.

I'm on Win7, using VS2015 professional. 

[Command.txt](https://github.com/fastbuild/fastbuild/files/311531/Command.txt)

ffulin:
On Windows, the normal command line args limit is just short of 32KiB.  When this limit is exceeded, FASTBuild writes the command line to a "response file" and passes a specially formatted arg to cl.exe that references that.

One thing to note is that the command line limit includes the exe path, which FASTBuild takes into account when deciding whether to use the response file.  It's possible that this might explain why a particular length arg works in FASTBuild, but doesn't in your test.

Now, this particular error seems to be something a little different. As you've probably noted, there's basically zero useful info on MSDN about it, but if I had to guess, I'd guess that under some circumstances, there is some additional constraint which might mean that we need to use response files for something even shorter than 32KiB (the ~20KiB you are seeing).

Interestingly, I've never seen this particular error before (while I have seen the 32KiB limit), which makes me wonder if there is something about the args that is relevant?  I looked at the args you sent and nothing out of the ordinary stood out.  I'll try to do some testing to see if I can narrow something down.

It'll also be interesting to see if response files fix this problem (you could test that by calling cl.exe directly as you've been doing, but with the response file args ([detailed here](https://msdn.microsoft.com/en-us/library/3te4xt0y.aspx)) or by changing the limit in FASTBuild to force response files for shorter limits.

I feel like we're probably not that far away from working out exactly what the problem case is, which means we'll likely be able to have a work around in place.

ffulin:
I re-read your message where you already noted that response files didn't solve the problem. Sorry I missed that the first time.

[...]

PR #185, opened 2016-12-31
For large projects that distribute their builds, having an extremely …

gkoreman:
…long path can cause error D8049 on MSVC. See https://github.com/fastbuild/fastbuild/issues/105

The error only presents itself on remote workers. Preprocessing only happens on the host machine. Removing all /I includes drastically reduces the command line length for the remote workers, and the includes are not required anyway (except for Recode? Not sure what that is).
This still feels like a workaround since it's unclear why the preprocessor doesn't choke on the long command line, and why the preprocessed code with the long command line does not fail on the host machine.


gkoreman:
Great! I should just close this pull request?

ffulin:
Yeah, if you're cool with the overall outcome.  (I'm keeping the https://github.com/fastbuild/fastbuild/issues/105 issue open until the next version is released)

PR #186, opened 2016-12-31
Dev worker options

gkoreman:


ffulin:
Thanks for splitting up the pull request. This one seems to still contain all the "simple distribution" changes. Since I'll integrate that first, this is likely to cause issues/conflicts.

gkoreman:
Still getting the hang of git. I made another pull request with only the desired changes.

PR #187, opened 2017-01-01
Dev worker options clean

gkoreman:


Code/Tools/FBuild/FBuildWorker/Worker/WorkerSettings.cpp
```diff
@@ -82,10 +84,11 @@ void WorkerSettings::Load()
...
         // settings
-        uint32_t mode;
-        f.Read( mode );
-        m_Mode = (Mode)mode;
-        f.Read( m_NumCPUsToUse );
+        f.Read( m_NumCPUsToUseWhenIdle );
+		if (header[3] >= FBUILDWORKER_SETTINGS_SUPPORT_DETAILED_CPU)
+		{
+			f.Read(m_NumCPUsToUseDedicated);
+		}
```
ffulin: I think there might still be a bug here.  If we're loading the previous version data, we'll read the "mode" as the number of cpus to use.
Seems like we need:
```C++
if ( header[3] < FBUILDWORKER_SETTINGS_SUPPORT_DETAILED_CPU )
{
   uint32_t mode;
   Read( mode );

   // Set NumCPUsWhenIdle/Dedicated based on the mode?
}
```

PR #192, opened 2017-01-13
Handle invalid EOL markers

wickerwaka:
When dealing with mixed-format files clang/gcc will sometimes add the line directives after '\r' rather than a '\n'

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/CIncludeParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/CIncludeParser.cpp
@@ -248,12 +248,17 @@ bool CIncludeParser::ParseGCC_Preprocessed( const char * compilerOutput,
 
     for (;;)
     {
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for tracking down this bug and for the fix. I added a regression test and confirmed that your fix resolves the bug, but unfortunately it also comes with a significant performance hit (include parsing goes from ~3.8GiB/s -> ~0.3GiB/s).

With that in mind, I re-implemented the fix (https://github.com/fastbuild/fastbuild/commit/ef84d20000031fe9bde2b24b5eb4d705c6cebe49) without use of strpbrk to avoid the slowdown (and got a small bonus perf boost compared to the original version - now ~4.0GiB/s).

The fix will be in the next release (v0.93). I'll leave this pull request open until that new version is released.

Thanks again for narrowing it down. It's great to have finally killed this rare but annoying edge-case!

PR #195, opened 2017-01-15
allow FASTBUILD_TEMP_PATH env var to override system temp dir

### LINKED ISSUE #194: Alternate TMP directory on worker machine not working
If a remote worker has set their TMP directory to something non-standard then fastbuild fails.

A few of our users moved their tmp directory to a secondary drive in order to free up space on their C: but this caused their remote workers to continuously fail.

Here is an example excerpt from a log:
PROBLEM: C:\Users\gkoreman\AppData\Local\Temp\UnrealShaderWorkingDir\263F66F64CCD23C0AEAEC0845284E7B5\FASTBuild\0\66\Worker.out
build	14-Jan-2017 13:20:44	UE4Editor-Cmd: Failed to build Object (error 0xc0000005) 'D:\TMP\.fbuild.tmp\0x00000000\core_1003\Worker.out'

On FASTBuild Workers that have standard TMP folders this issue does not occur.

comicfans:
some antivirus (such as McAfee) prevent executable run from system temp dir. this patch add a env var to override this temp location. it may also fix #194

### DIFF
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -375,10 +375,22 @@
 {
     #if defined( __WINDOWS__ )
         char buffer[ MAX_PATH ];
-        DWORD len = GetTempPath( MAX_PATH, buffer );
-        if ( len != 0 )
+        DWORD res = GetEnvironmentVariable("FASTBUILD_TEMP_PATH",
+                buffer,MAX_PATH);
+        if ( res == 0 )
+        {
+            DWORD len = GetTempPath( MAX_PATH, buffer );
+            if ( len != 0 )
+            {
+                output = buffer;
+                output += '\\';
+                return true;
+            }
+        }
+        else
         {
             output = buffer;
+            output += '\\';
             return true;
         }
     #elif defined( __LINUX__ ) || defined( __APPLE__ )
--- a/Code/Core/Process/Semaphore.cpp
+++ b/Code/Core/Process/Semaphore.cpp
@@ -111,7 +111,7 @@ void Semaphore::Wait( uint32_t timeoutMS )
             ts.tv_nsec = totalNS % 1000000000;
             if ( sem_timedwait( &m_Semaphore, &ts ) != 0 )
             {
-                ASSERT( errno == ETIMEDOUT ); // anything else is a usage error?
+                ASSERT( ( errno == ETIMEDOUT ) || (errno == EINTR ) ); // anything else is a usage error?
             }
         #endif
     }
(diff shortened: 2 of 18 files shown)


Code/Core/FileIO/FileIO.cpp
```diff
@@ -375,10 +375,22 @@
...
+            {
+                output = buffer;
+                output += '\\';
+                return true;
+            }
+        }
+        else
         {
             output = buffer;
+            output += '\\';
```
ffulin: Seems like this would add a trailing backslash even if the environment variable already has one. Also, according to the MSDN documentattion, the ::GetTempPath call should return the path with a trailing backslash already (though I've seen evidence that this might not be true).  Seems like the best bet would be to check if it's already there and add it if not in both cases.

Code/Core/Process/Semaphore.cpp
```diff
@@ -111,7 +111,7 @@ void Semaphore::Wait( uint32_t timeoutMS )
             ts.tv_nsec = totalNS % 1000000000;
             if ( sem_timedwait( &m_Semaphore, &ts ) != 0 )
             {
-                ASSERT( errno == ETIMEDOUT ); // anything else is a usage error?
+                ASSERT( ( errno == ETIMEDOUT ) || (errno == EINTR ) ); // anything else is a usage error?
```
ffulin: This change seems to be unrelated to the TMP path change. Can you provide some more details about this please?

Also, it would really help if you could isolate changes into their own pull requests - easier to integrate and test them in isolation.

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionExec.cpp
```diff
@@ -108,7 +110,8 @@ FunctionExec::FunctionExec()
                                            workingDir,
                                            expectedReturnCode,
                                            preBuildDependencies,
-                                           useStdOutAsOutput);
+                                           useStdOutAsOutput,
+                                           isGenerator);
```
ffulin: This change seems unrelated?

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionExecutable.cpp
```diff
@@ -209,7 +212,8 @@ FunctionExecutable::FunctionExecutable()
                               assemblyResources,
                               importLibName,
                               linkerStampExe,
-                              linkerStampExeArgs );
+                              linkerStampExeArgs,
+                              preBuildDependencies);
```
ffulin: This change seems unrelated?

ffulin:
This pull request seems to contain a lot of unrelated changes. That makes it quite difficult to analyze and integrate.  Could you please split the pull requests into individual changes?  That would make it a lot easier to integrate them.

Code/Core/Process/Semaphore.cpp
comicfans: Sorry, I forgot to rename my pull branch ,I 'll recreate it again

PR #196, opened 2017-01-16
added generic SAK source control strings in generated VSprojects

lucianm:
Hello Franta, it's been a while (like 2 years) since we last talked via email, at times before you published fastbuild on github. Recently I've convinced my colleagues we should use fastbuild for a large project, therefore I had to deal with several challenges in writing BFFs in such a manner they would also integrate with Visual Studio as lean as possible, to have them accept it more easily ;)

I was able to address one of them by adding support for generic source control binding status in the generated VSXproject files, just by adding 4 generic lines, according to the explanations found in [this blog](http://alinconstantin.blogspot.de/2008/02/sak-source-control-strings-in-visual.html). The effect of having generated project files containing these binding lines can be seen in the attached screenshot, in our case when using TFS, we can see these tiny source control status icons (blue locks or red check mark) beside every source file referenced in the generated project:

fastbuild_scc_icons_in_generated_vsxproj

I hope this is useful to others as well and can find its way into the next release. As a matter of fact, I'd actually rather contribute some more stuff in the near future, and also discuss some of the ideas.

Best regards,
Lucian

Later Edit: I forgot to mention, Visual Studio adds these lines in generated projects anyway, if they are not present and the solution referring to the projects has source control binding (which is quite common usage).


ffulin:
If I remember correctly, when these settings are present in the project, if you don't have a source control plugin installed, you get an annoying popup everytime you open the sln.  Do you know if that's still the case?  If it is, it might be necessary to expose an option to allow users to chose if they want this or not.

lucianm:
Hmm, I don't know exactly if that is still the case, and I also cannot really test it. I'd rather add that option you're suggesting (optional, default off), to this pull request, to be on the safe side, what do you think?

ffulin:
Thanks for the pull request update and sorry for all the places you have to update to shuffle the bool around (that's being fixed too but I didn't get to the VCXProjectNode yet!)

ffulin:
I've integrated these changes into the Perforce mainline. This will be in the next release (v0.94). Thanks for your changes!

PS - I verified that adding the option to activate the extra fields was a good choice - without it, Visual Studio spams a bunch of warnings about missing scc information on solution load if you don't have it setup.

PR #197, opened 2017-01-18
fix processing windows full path on other platforms

comicfans:
windows fullpath treats as none fullpath on other platforms incorrectly,
leads following FileNode constructor assert false

### DIFF
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -30,17 +30,21 @@
 //------------------------------------------------------------------------------
 /*static*/ bool PathUtils::IsFullPath( const AString & path )
 {
-    #if defined( __WINDOWS__ )
-        // full paths on Windows are in X: format
-        if ( path.GetLength() >= 2 )
-        {
-            return ( path[ 1 ] == ':' );
-        }
-        return false;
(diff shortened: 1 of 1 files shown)


ffulin:
Hi there.  What's the case where you would have a Windows path on Linux or OSX?

comicfans:
when windows style full path read  to construct FileNode, in CleanPath, it will makes a path as /home/comicfans/fastbuild/c:\program files\...\cl.exe  then FileNode.cpp line 23 assert false
this makes fastbuild crash under linux(or osx) if any windows style full path written as file Path

ffulin:
Finally coming back to this. A path with a colon in it is sadly valid on linux, so I don't think we can assume that a colon at particular position indicates a full path outside of Windows.

The assert in FileNode.cpp that was firing should have been a Windows-only assert. That bug has been fixed relatively recently (https://github.com/fastbuild/fastbuild/commit/bf46e19d622155eb79ff1435d1af560da0996cfc#diff-1b668c04a60c1a3850d99d4946a19c03).

I couldn't reproduce the crash you mentioned with any use of full paths under Linux. It's quite possible this has also been fixed in the internim. I'm going to close this issue for now. If you experience any crashes or incorrect behavior related to full path handling, please comment here and we can re-open the issue.

PR #198, opened 2017-01-18
Supports PreBuildDependencies in Exe Node

comicfans:
link libraries may be embedded in link options, not easy to extract them
out as Libraries variable, it's better to support PreBuildDependencies

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionExecutable.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionExecutable.cpp
@@ -199,6 +199,12 @@ FunctionExecutable::FunctionExecutable()
     }
     else
     {
+
+        Dependencies preBuildDependencies;
+        if ( !GetNodeList( nodeGraph, funcStartIter, ".PreBuildDependencies", preBuildDependencies, false ) )
+        {
+            return false; // GetNodeList will have emitted an error
+        }
         n = nodeGraph.CreateExeNode( linkerOutput,
                               libraryNodes,
                               otherLibraryNodes,
@@ -209,7 +215,8 @@ FunctionExecutable::FunctionExecutable()
                               assemblyResources,
                               importLibName,
                               linkerStampExe,
-                              linkerStampExeArgs );
+                              linkerStampExeArgs,
+                              preBuildDependencies);
     }
 
     return ProcessAlias( nodeGraph, funcStartIter, n );
--- a/Code/Tools/FBuild/FBuildCore/Graph/ExeNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ExeNode.cpp
@@ -25,10 +25,14 @@ ExeNode::ExeNode( const AString & linkerOutputName,
                   const Dependencies & assemblyResources,
                   const AString & importLibName,
                   Node * linkerStampExe,
-                  const AString & linkerStampExeArgs )
+                  const AString & linkerStampExeArgs,
+                  const Dependencies & preBuildDependencies
+        )
 : LinkerNode( linkerOutputName, inputLibraries, otherLibraries, linkerType, linker, linkerArgs, flags, assemblyResources, importLibName, linkerStampExe, linkerStampExeArgs )
 {
     m_Type = EXE_NODE;
+
+    m_PreBuildDependencies = preBuildDependencies;
 }
 
 // DESTRUCTOR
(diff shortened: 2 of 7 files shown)


comicfans:
following two commits implements two-stage build for fastbuild , if a Exec Node marks as generator, fastbuild will run it first if it's out of date, then reload and build targets. this is useful for cmake or script generated fastbuild project,  only command needs to run is just fbuild

ffulin:
Hi there. I'm finding it difficult to evaluate the "generator" part of your pull request. Could you please provide some more information about what problem this is trying to solve?

Also, would it be possible to move the PreBuildDependencies change into a different pull request? That would likely be trivially mergeable and would make things easier for me, as well as aiding the clarity of the rest of the changes.

comicfans:
these changes made mainly to support CMake(or maybe other bff generators), in CMake ,user write high level build language (CMakeLists.txt) and CMake lower it down to fastbuild bff, when CMakeLists.txt changed, fastbuild can automatically rerun CMake to generate new bff ,then build with new one(so called two stage build).  currently fastbuild only load file and build once, so user must manually rerun generator(if generator 's input changed), then run fastbuild

comicfans:
I've put these three commits together because if  bff is manual written , use can control link libraries, so my first commit is not very useful. but  for bff generator,link libraries maybe constructed in link flags.  it should be better to support PreBuildDependencies , so generator can use target instead of libraries to construct dependency.That's why I put it as first commit. If individual PR is better ,I'll recreate PR again

ffulin:
I did some refactoring to Executable/DLL. That created a conflict with this pull request, but I've gone and added PreBuildDependencies to both of those in a follow up change (https://github.com/fastbuild/fastbuild/commit/6b3443758dacbc18d17032ef1c5a136637e57c91).

One thing that should be noted is that .PreBuildDependencies are special compared to normal dependencies. FASTBuild will ensure they are up-to-date before building the target, but if any of the PreBuildDependencies change, it won't cause the target to rebuild. This allows you to have things that always run (like code generation) and not have the code gen itself cause an unnecessary rebuild.

comicfans:
glade to see this is implemented , I'd like to add some comments (I haven't touch fbuild codebase for a while, please correct me if I misunderstood it)

 I think "PreBuildDependencies change not cause target rebuild"  is not a good idea, for example a binary target link to a static lib (by link flags and use PreBuildDependencies to ensure correct order), will not relink if static lib changed. and for the codegenerator scenario , I think build tool can not "guess" user's decision in all cases,  maybe an always  rerun codegenerator want to download lastest resources from web at build time , and a executable target embedded them in,  if codegenerator rerun , but executable not rebuild, this leads wired behavior . I think in this case, user should makes a build structure as following to avoid always rebuild problem:
  
codegenerator(always rerun)=> outfiles   =>  copyfileifchanged (always rerun ,but output won't update if no filechanged)=> outfiles  =>  executable 

actually LLVM have such build script, it  have lots of headers which is generated by codegenerator (and codegenerator is built in source too),  when the codegenerator code changed, it rebuilt, and headers generated by it **may** change , so the codegenerator rerun to generate new headers, and these headers are copied to final headers (if changed).   

by this structure, if codegenerator output changed, final target always get latest headers, or final target won't rebuild if only the codegenerator itself code changed.

this is more complicate than "PreBuildDependencies change not trigger target rebuild" , but I think this can cover all use cases.

ffulin:
Things like ObjectList, Executable and DLL all have parameters to specify the actual dependencies, by:
a) Listing them explicitly (.CompilerInputFiles for ObjectList, .Libraries for DLL/Exectuable etc)
b) Specifying directories and search patterns (.CompilerInputPath etc for ObjectList)
c) Extracting them from the command line if specified manually (for libs in .LinkerOptions in DLL and Exectuable - this is not well documented so it might not be obvious that this is handled)

So it should be possible to have all the correct dependencies for "normal" use without .PreBuildDependencies being used. If you change any of the things specified by the above mechanisms, rebuilds should correctly occur.

.PreBuildDependencies were added to allow for the special case of needing to force something to occur before a directory listing is performed - i.e. the code-gen case. For that case, it's preferable for the act of running code-gen alone to not cause rebuilds (only if the files generated change should builds occur). This is why it has the behavior I described (again something that needs improved documentation).

So I think the cases you describe are handled:
 -  **libs on LinkerOptions** - These are auto detected by FASTBuild and managed as normal dependencies (FASTBuild understands the various -l/-L or /LIBPATH args)
 - **code gen generating/downloading files that are then built** - You would use .PreBuildDependencies to trigger the code-gen, but then use .CompilerInputPath or similar to find the files and build them which then has the proper dependencies for rebuilding if those change.

Have I missed some subtlety of your use-cases that isn't working properly (either due to an oversight in this design or a bug)?

comicfans:
Thanks for this detail explanation , during last time I tried (maybe 0.93), fbuild still not have support extract dependencies from link flags, (or maybe I've triggered a bug? lib rebuilt, but exe link to it not rebuilt). if fbuild can handle this and this is OK for my normal usage (but what will happen when two libs with same name appears in different link search folders ? will fbuild handle this order just like linker  ? or just choose first existed file as dependency? of course this is bad practice and rare )

and for codegen scenario, the rebuild is triggered from file inputs, not PreBuildDependencies, I've misunderstood it.

ffulin:
> during last time I tried (maybe 0.93), fbuild still not have support extract dependencies from link flags,

FASTBuild has been auto-extracting dependencies from the LinkerOptions for a very long time (since at least v0.63 - in 2014). If you have a case where this is not working as expected, we can work together to understand why and fix it.

> but what will happen when two libs with same name appears in different link search folders ?

I never checked this specific behavior. The documentation (at least for msvc's link.exe) doesn't specify what happens in this case. Right now, FASTBuild checks the LIBPATHs in the order they are specified on the command line. If this is wrong, it can be changed, but as you say, two libs with the same name differentiated by only LIBPATH is probably a bad idea.

comicfans:
I need some time to find the fbuid environment which trigger the problem,
that happens about half a year ago

2017年12月22日 上午12:54，"ffulin" <notifications@github.com>写道：

ffulin:
> I need some time to find the fbuid environment which trigger the problem,

No problem. If you do see the issue again, we can create a new issue to track and solve it.

ffulin:
v0.95 has now been released and includes .PreBuildDependencies support for Executable and DLL. Thanks for your help with this issue!

PR #199, opened 2017-01-18
added .BuildLogFile path in generated VSprojects

lucianm:
This allows to control where msbuild will write log files from FBuild-generated VSprojects, since they would otherwise be written anyway in some default location which may not suit anyone.


ffulin:
Seems like a good improvement. Thanks for the pull request!

ffulin:
I've integrated this change into the Perforce mainline. This will be in the next version (v0.94). Thanks for the pull request!

PR #201, opened 2017-01-21
allow FASTBUILD_TEMP_PATH environment variable to override system tem…

comicfans:
…p dir

some antivirus (such as McAfee) prevent executable run from system temp dir.

it apply to windows only , as I know most linux/osx user didn't use antivirus. if they are, we may extend this to all platforms.

this didn't check trailing backslash, because win32 api accept path with extra backslash, and if we really want to clean path , must reverse checking every char from trailing to ensure no extra backslash , I think leave a(or maybe some) extra backslash is a little easier.

### DIFF
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -375,10 +375,22 @@
 {
     #if defined( __WINDOWS__ )
         char buffer[ MAX_PATH ];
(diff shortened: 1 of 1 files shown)


ffulin:
I've integrated your changes into the Perforce mainline. They will be in the next release (v0.94). Thanks for the pull request!

comicfans:
my pleasure !

PR #202, opened 2017-01-21
implements Ctrl-C on linux

comicfans:
may also works for OSX, but I didn't have one to test. so I leave this code predefined for linux only

### DIFF
--- a/Code/Core/Process/Semaphore.cpp
+++ b/Code/Core/Process/Semaphore.cpp
@@ -111,7 +111,7 @@ void Semaphore::Wait( uint32_t timeoutMS )
             ts.tv_nsec = totalNS % 1000000000;
             if ( sem_timedwait( &m_Semaphore, &ts ) != 0 )
             {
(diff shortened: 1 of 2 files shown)


Leandros:
That is technically undefined behaviour. While it may work, the only legal thing you're allowed to do in a signal handler is setting a `volatile sig_atomic_t`.

comicfans:
Hmm, I should read your references first ...

ffulin:
I think forking a process in order to be strictly compliant is going to be unworkable (since we'd have to pass all the state across processes anyway).
I think even if it's not strictly per the C++ standards, it's worth handling the Ctrl+C as in the pull request, because it'll almost certain work just fine, and even if it doesn't, then in the absolute worst case we're no worse off than the current behaviour.

ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.94). Thanks for the pull request.

PR #207, opened 2017-01-26
Handle both LF & CRLF output in MSCL include parser

poppolopoppo:
LLVM for windows is mimicking cl.exe, with a slight difference though : the output is LF terminated instead of CRLF.

This patch allows the parser to be line-end agnostic.


ffulin:
I'm sorry this took so long to integrate, especially for such a simple but important fix! I have integrated this into the Perforce mainline. The fix will be in the next release (v0.95). I extended the CIncludeParser tests to catch the case this fixes, and add coverage for various other situations (https://github.com/fastbuild/fastbuild/commit/fb8ff776e42d7bb4fa848c786aae35ec9468fe7e)

PR #208, opened 2017-02-06
PreprocessorGenericDependencies

wickerwaka:
Experimental change, open to suggestions for other ways to do this.

Added support for PreprocessorGenericDependencies option to the
ObjectList node. When enabled the custom preprocessor will be used one
to generate dependency information. The preprocessor must write out the
list of dependencies to stdout.

This flag makes it possible to add support for additional compilers and
tools that introduce dynamic dependencies. For instance, the QT resource
compiler could use this method rather than needing to be handled as a
special case in the code.

### DIFF
--- a/Code/Core/Mem/MemDebug.cpp
+++ b/Code/Core/Mem/MemDebug.cpp
@@ -12,6 +12,9 @@
 // Core
 #include "Core/Env/Assert.h"
 
+// System
+#include <stdint.h>
+
 // FillMem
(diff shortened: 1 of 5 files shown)


ffulin:
The idea here seems pretty reasonable, though I'm wondering if this is the right way to move forward. As you've no doubt noticed, the web of what to call with what arguments is getting a little out of control.

I wonder if how the compiler is detected, and how the code is structure needs a bit of a re-think?

PR #209, opened 2017-02-12
Added #importfullpath directive - similar to #import…

lucianm:
Sometimes environment variables which are imported by FASTBuild contain a single file or folder path which is possibly passed only from a local environment context, like setting the variable in a command line used in a FASTBuild-generated VCXProject right before calling fbuild. This might trigger a rebuild if the same actual path is given as different symlinks or relative path segments (../../) combinations, even if they actually resolve to the same location. If FASTBuild would know that this environment variable is supposed to be a path, it could resolve it to the most compact form, and between runs this would result in the same actual path for which the hash is calculated on parsing BFF. This pull request adds exactly this feature, as a new directive pair, #importfullpath and #existsfullpath, the second one being needed to inform that the enclosed #import is supposed to be a path, and handle it accordingly. The path must exist (since we know it has to be a path, otherwise we would use plain #import), otherwise parsing ill fail.


ffulin:
I'm struggling to understand the problem this pull request is trying to solve. Do you think you could please provide some more context on the situation in which you would be importing a variable which would be different between runs, but somehow actually pointing to the same location?

lucianm:
Well, imagine some more complex setups where FASTBuild is used from several generated VS projects (even in the same solution, but not necessarily), still sharing the same BFF structure with its unique config and FDB.
If for whatever reason, the VS projects need to pass to the fbuild.exe call some path via environment variable, constructed from its own visual studio variable semantics, and this is related to things like relative path constructs around $(ProjectDir) and similar, the invoked fbuild.exe will import a string possibly resolving to some path (imagine it might be needed in some test target to be passed as an argument). If this string just looks different between runs (because it was constructed from a VS project possibly located in a different folder hierarchy) but in fact would still point to the same location, if FASTBuild would know that a path is meant, it would only consider it different if it truly is different (as a path), and not also if it just looks different (as a string). This of course saves rebuilding, this is my patch about :)

lucianm:
Hello Franta, if you need more clarification, please let me know. As a matter of fact, I managed to get along without this feature in the mean time, by completely separating test targets from build targets (so they have a different config BFF and FDB, so rebuilding in the "build" setup is at least no longer triggered unnecessarily by imported paths needed in the "test" setup, so not having the _#importfullpath_ no longer hurts me so much. So it's not that important to me anymore. Nevertheless, I rebased my pull request branch onto more recent status of your "dev" branch, in case you should decide to adopt it.

ffulin:
Thanks for the clarifications!

PR #210, opened 2017-02-15
VCXProjectNode uses full file path for deterministic GUID generation.

rickvanprim:


rickvanprim:
Looked at the blame for VCXProjectNode and using the just the name (plus the slash right before it) seems to pre-date git.  From the comment I'm guessing there's a reason for that, but I couldn't discern what.

PR #211, opened 2017-02-15
VCXProjectNode uses full file path for deterministic GUID generation.

rickvanprim:
Looked at the blame for VCXProjectNode and using the just the name (plus the slash right before it) seems to pre-date git. From the comment I'm guessing there's a reason for that, but I couldn't discern what.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
@@ -154,9 +154,7 @@ void SLNGenerator::WriteProjectListings( const AString& solutionBasePath,
         AStackString<> projectGuid;
         if ( (*it)->GetProjectGuid().GetLength() == 0 )
         {
-            // For backward compatibility, keep the preceding slash and .vcxproj extension for GUID generation
-            AStackString<> projectNameForGuid( lastSlash ? lastSlash : projectPath.Get() );
-            VSProjectGenerator::FormatDeterministicProjectGUID( projectGuid, projectNameForGuid );
+            VSProjectGenerator::FormatDeterministicProjectGUID( projectGuid, projectPath );
         }
         else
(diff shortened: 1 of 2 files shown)


Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
```diff
@@ -189,12 +187,8 @@ void SLNGenerator::WriteProjectListings( const AString& solutionBasePath,
                 // get all the projects this project depends on
                 for ( const AString & dependency : deps.m_Dependencies )
                 {
-                    // For backward compatibility, keep the preceding slash and .vcxproj extension for GUID generation
-                    const char * projNameFromSlash = dependency.FindLast( NATIVE_SLASH );
-                    AStackString<> projectNameForGuid( projNameFromSlash ? projNameFromSlash : dependency.Get() );
-
                     AStackString<> newGUID;
-                    VSProjectGenerator::FormatDeterministicProjectGUID( newGUID, projectNameForGuid );
+                    VSProjectGenerator::FormatDeterministicProjectGUID( newGUID, dependency );
```
rickvanprim: This section seems wrong in that it always uses the path for the GUID and doesn't check if one has been explicitly set in the BFF file.

ffulin:
The project name was originally chosen as the source of the hash to generate deterministic GUIDs for any given project, while being independent of the location of the project.  Including the preceding slash was actually a bug, but has been retained in order to avoid unnecessarily changing the guids of projects generated by previous versions of FASTBuild (in case users were depending on that).

The change you propose would cause two users with different folder setups to generate projects with different guids, which if being mixed with non-FASTBuild generated projects could break compilation.

I guess you could argue that if you need explicitly stable path-independent guids, you could set them manually instead of relying on the generation.

Is the problem you are trying to fix that you have two projects with the same name in a single solution?

PR #212, opened 2017-02-15
FastBuildOptions gets ProgramName from command line and uses that in …

rickvanprim:
…place of hard-coded 'FBuild.exe' string.


ffulin:
Was this to fix running in wrapper mode if you renamed the exe on Windows?  Or to fix the help output on OSX/Linux?

ffulin:
I've integrated your change into the Perforce mainline. Your changes will be in the next release (v0.94). Thanks!

PR #214, opened 2017-03-06
Checkout path independant cache

poppolopoppo:
Hi,

This PR adds a .RootPath argument to Settings() function, which should point to current checkout directory. RootPath will be automagically replaced by a special char in hash input, making cache keys path agnostic : ie we can share cache between 2 different paths.

The PR is at prototype stage (no doc no tests) and this is  a brute force approach, but I was able to share the cache with just that and I would like to get your feedback.

Thanks!

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionSettings.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionSettings.cpp
@@ -6,6 +6,8 @@
 #include "Tools/FBuild/FBuildCore/PrecompiledHeader.h"
 
 #include "FunctionSettings.h"
+#include "Tools/FBuild/FBuildCore/FBuild.h"
+#include "Tools/FBuild/FBuildCore/FLog.h"
 #include "Tools/FBuild/FBuildCore/Graph/NodeGraph.h"
 #include "Tools/FBuild/FBuildCore/Graph/SettingsNode.h"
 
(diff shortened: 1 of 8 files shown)


ffulin:
I think there are a few problems with this approach:

*1) Incorrect debug info*
Most compilers I've seen embed full paths to source files in object file debug info. Retrieving an object on a machine that didn't build at the same location would result in mismatched debug info (i.e. being able to set breakpoints or find the right source file when debuggging)

*2) Incorrect __FILE__ directives*
It seems this approach will replace instances of __FILE__ in the preprocessed output which will cause use of that macro (typically in things like ASSERT macros) to display the placeholder token instead of the correct file path.

poppolopoppo:
Thanks for your feedback !

**2) Incorrect FILE directives** : the proposed PR only operates on hash input, and will never change the actual preprocessor output. So the replacement token will never be seen, but you could argue that a different path could be shown depending on the path used by the machine which actually compiled the cached TU.

**1) Incorrect debug info** : I admit that it needs more testing, but for instance MSVC is able to handle a different root path while debugging (just need to locate the first file not found).

Do you think of a better approach ?
Don't you think that if the behavior was kept optional and if these restrictions were well explained that the gain will still be interesting for many developers ?

I'll do more tests about 2) and see what limitations we would be dealing with.

ffulin:
I misread the pull request. You're right that the result won't have the placeholder token in it, but you're also right that the path will be arbitrarily wrong (whatever path happened to be used when populating the cache).

On the topic of the debug info, I think if you had multiple users populating the cache then you might end up having to browse the file several times in a debug session. I think this could get quite annoying.

One idea I had previously was to make FASTBuild (if you enable a toggle) map a drive before compiling, and compile within that drive to normalize the path across all users. it would still suffer from having to select a file for debugging the first time, but at least it would be normalized for all users. Additionally, we wouldn't need to touch the preprocessed output for hashing at all (which is good because I think performance-wise it'll be hard to make that acceptably performant).

ffulin:
I'm going to close this issue, since it won't be integrated as is. I think the auto-drive mapping option has a lot of promise for fixing this in the future.

PR #220, opened 2017-03-20
Isolate files list for UnityNode

poppolopoppo:
Hi,

This PR adds an optional field "UnityInputIsolateListFile" to Unity() functon.
It should point to a file containing a list of files which will be isolated from the unity builds, following the same behavior than writable files when "UnityInputIsolateWritableFiles" is true.
The list must contain absolute paths, and will be read at each execution of FastBuild, allowing it's content to change without unnecessarily invalidating the fdb.

Using this new option a developer can feed fastbuild with an external tool to drive unity builds, like git :)
By filling the specified files with modified files (for instance with a simple wrapper around fbuild.exe), git users can use this feature reserved to this day to Perforce users without putting the burden of handling every SCM on FastBuild.

This PR is not complete yet as it lacks documentation, and maybe a cache to avoid reading the external file every time when using several unity nodes with the same list.

### DIFF
--- a/Code/Core/FileIO/IOStream.cpp
+++ b/Code/Core/FileIO/IOStream.cpp
@@ -58,4 +58,70 @@ void IOStream::AlignWrite( size_t alignment )
     ASSERT( ( Tell() % alignment ) == 0 );
 }
 
+// AlignWrite
+//------------------------------------------------------------------------------
+bool IOStream::ReadLines( Array< AString > & lines )
+{
+    const uint64_t size = GetFileSize();
+
+    const int bufferSize = 2048;
+    char buffer[2048];
+
+    uint64_t start = Tell();
+    uint64_t i = start;
+    for ( ; i<size ; ++i )
+    {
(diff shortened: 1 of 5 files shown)


ffulin:

I made a follow up commit (https://travis-ci.com/github/fastbuild/fastbuild/builds/158323151). The most notable changes are:
 - the text file is correctly parsed when it contains blank lines
 - information about which files are found in the isolation list file is gated behind -verbose
 - if an isolation file is specified, it is required to be present (doesn't silently fail if you mistype the filename)

ffulin:
I made an additional change (https://github.com/fastbuild/fastbuild/commit/3047cb6a5ffc112910fd3429694a096ebf7e51ee) which adds the UnityInputIsolateListFile as a dependency for the UnityNode. This allows this file to be generated by another step (like a call to Exec for example).

poppolopoppo:
Thank you for your hard work :)

PR #222, opened 2017-03-23
Adding the prx compiler linker pass through argument for detecting a dll node

jrf2:
The orbis-clang compiler allows users to use the same executable to call
the linker but linker options need to be prefixed "-Wl," so fastbuild
doesn't see that "--oformat=prx" is used because of the extra "-Wl,"
characters before it.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
@@ -612,7 +612,7 @@ void LinkerNode::GetAssemblyResourceFiles( Args & fullArgs, const AString & pre,
         for ( const AString * it=tokens.Begin(); it!=end; ++it )
         {
             const AString & token = *it;
-            if ( ( token == "-shared" ) || ( token == "-dynamiclib" ) || ( token == "--oformat=prx" ) )
+            if ( ( token == "-shared" ) || ( token == "-dynamiclib" ) || ( token == "--oformat=prx" ) || ( token == "-Wl,--oformat=prx" ) )
             {
                 flags |= LinkerNode::LINK_FLAG_DLL;
                 continue;


Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
```diff
@@ -612,7 +612,7 @@ void LinkerNode::GetAssemblyResourceFiles( Args & fullArgs, const AString & pre,
         for ( const AString * it=tokens.Begin(); it!=end; ++it )
         {
             const AString & token = *it;
-            if ( ( token == "-shared" ) || ( token == "-dynamiclib" ) || ( token == "--oformat=prx" ) )
+            if ( ( token == "-shared" ) || ( token == "-dynamiclib" ) || ( token == "--oformat=prx" ) || ( token == "-Wl,--oformat=prx" ) )
             {
```
ffulin: Seems like this would miss cases like:
```C++
-Wl,--someotherthing,--oformat=prx
```
How about:
```C++
|| ( token.BeginsWith( "-Wl" ) && token.Find( "--oformat=prx" ) )
```

fitzymj:
If it helps, I think _all_ clang and gcc front-ends let you pass linker parameters through like that (not just `orbis-clang`).  It might be worth generalizing?

ffulin:
I wasn't able to merge this pull request as it was created against the master branch (instead of the dev branch). I have however manually applied the fix to the dev branch (https://github.com/fastbuild/fastbuild/commit/16a776bf7064fbdea9b7a0479feb391dda975c29). I ended up adding the slightly more robust fix as follows:
```C++
|| ( token.BeginsWith( "-Wl" ) && token.Find( "--oformat=prx" ) )
```
This fix will be in the next release (v0.98).

Thanks for reporting the bug and thanks for the fix!

PR #227, opened 2017-03-27
[Fix] Do not report success if the wrapper crashes early on.

samhocevar:
The return code in shared data is initialised to zero by the main process,
so if the child crashed it wrongly assumes its exit code was zero.


ffulin:
I've integrated this change into the Perforce mainline. This will be in the next release (v0.94). Thanks for the fix!

PR #231, opened 2017-04-10
treat C1001 (internal compiler error) as system error

samhocevar:
Error C1001 may be caused by bugs in the compiler, but most of the time it
is a sign of faulty hardware. We do not want a single malfunctioning host
to cause all compilations to fail, so treat these as system errors.


ffulin:
Thanks for the pull request. I've integrated this change into the Perforce mainline. This will be in the next release (v0.94)

PR #235, opened 2017-04-19
Fix incorrect mutex usage in ServerState::m_Jobs access.

samhocevar:


ffulin:
At a glance this looks like a sensible change. I'll integrate this and do a more thorough analysis of the change.

PR #238, opened 2017-05-03
Add support for new optional parameters: .TargetName, .TargetExt, .AdditionalIncludeDirectories, .WebBrowserDebuggerHttpUrl and .ProjectBuildType to VCXProject, and support CustomBuild steps on Utility projects for fast iteration.

missmah:
Add support for new optional parameters: .TargetName, .TargetExt, .AdditionalIncludeDirectories, .WebBrowserDebuggerHttpUrl and .ProjectBuildType to VCXProject, and support CustomBuild steps on Utility projects for fast iteration.

This is an updated PR which cherry picks just the changes to these files, without the history that they had in the other PR. I will close out the other PR next.

Signed-off-by: Layla Mah <layla@insightfulvr.com>

PR #240, opened 2017-05-05
This change adds support for generating usable Visual Studio 2015 .androidproj projects via the FastBuild VCXProject Function…

missmah:
…. These generated .androidproj projects enable native C++ code debugging on attached Android devices from within Visual Studio 2015.

NOTE: This change is a superset of #238.

To generate a `.androidproj` via `VCXProject`, simply use the `.androidproj` extension instead of the `.vcxproj extension` inside your .bff code.

This change also adds a couple of new and needed features to VCXProj to support functional .androidproj generation:
1. "`.PackagePath`" - This is the location of your APK file on disk [e.g. `Output/ARM/ProjectName/APK/bin/ProjectName-debug.apk`]
2. "`.LaunchActivity`" - This is the activity name to launch when the APK is deployed. [e.g. `com.android.NativeActivity`]
3. "`.PlatformIncludePattern`" - This is a string which can be used to specify which matching platforms should be included in the `.vcxproj` [e.g. `ARM` - would match `ARM` and `ARM64`]
4. "`.TargetArchitectureAlias`" - This is a string which can be used to add an alias between a user-defined target platform (e.g. `Android-ARM`) and another MSBuild target platform (e.g. `ARM`)
5. "`.AdditionalSymbolSearchPaths`" - This is a semi-colon delimited string fed to Visual Studio which is used to tell the `.androidproj` where to find debugging symbols for the code deployed inside the APK.
6. "`.AdditionalIncludePaths`" - This was defined in #238 and is re-used here to give the android debugger access to paths where it can find source files for debugging.

This change also adds code to the solution generation (`SLNGenerator.cpp`) to use a different GUID `{39E2626F-3545-4960-A6E8-258AD8476CE5}` for referencing .androidproj(s).
I found this out the difficult way, by trial and error, using VS to make things work, and then looking what it did differently in its generated XML.

There is also the preliminary mention of .csproj here. This change could be extended to cover Visual Studio .csproj and .iosproj projects eventually.

Signed-off-by: Layla <layla@insightfulvr.com>

---------------------------------------------------------------------------

For reference, here is the commit message for #238:

Add support for new optional parameters: `.TargetName`, `.TargetExt`, `.AdditionalIncludeDirectories`, `.WebBrowserDebuggerHttpUrl` and `.ProjectBuildType` to `VCXProject`, and support `CustomBuild` steps on `Utility` projects for fast iteration.

This is an updated PR which cherry picks just the changes to these files, without the history that they had in the other PR. I will close out the other PR next.

Signed-off-by: Layla Mah layla@insightfulvr.com

---------------------------------------------------------------------------

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionVCXProject.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionVCXProject.cpp
@@ -155,7 +155,10 @@ FunctionVCXProject::FunctionVCXProject()
     if ( !GetString( funcStartIter, baseConfig.m_BuildCommand,  ".ProjectBuildCommand", false ) ||
          !GetString( funcStartIter, baseConfig.m_RebuildCommand,".ProjectRebuildCommand", false ) ||
          !GetString( funcStartIter, baseConfig.m_CleanCommand,  ".ProjectCleanCommand", false ) ||
+         !GetString( funcStartIter, baseConfig.m_TargetName,    ".TargetName", false ) ||
+         !GetString( funcStartIter, baseConfig.m_TargetExt,     ".TargetExtension", false ) ||
          !GetString( funcStartIter, baseConfig.m_Output,        ".Output", false ) ||
+         !GetString( funcStartIter, baseConfig.m_AdditionalIncludePaths,    ".AdditionalIncludeDirectories", false ) ||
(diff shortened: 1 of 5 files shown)


ffulin:
This pull request is against the master branch. Could you re-create against the dev branch please?

ffulin:
Replaced by new pull request.

PR #242, opened 2017-05-14
Add support for new optional parameters: .TargetName, .TargetExt, .AdditionalIncludeDirectories, .WebBrowserDebuggerHttpUrl and .ProjectBuildType to VCXProject, and support CustomBuild steps on Utility projects for fast iteration.

missmah:
Duplicate of https://github.com/fastbuild/fastbuild/pull/238, but now against dev branch :)

Add support for new optional parameters: .TargetName, .TargetExt, .AdditionalIncludeDirectories, .WebBrowserDebuggerHttpUrl and .ProjectBuildType to VCXProject, and support CustomBuild steps on Utility projects for fast iteration.

This is an updated PR which cherry picks just the changes to these files, without the history that they had in the other PR. I will close out the other PR next.

Signed-off-by: Layla Mah layla@insightfulvr.com

PR #244, opened 2017-05-14
This change adds support for generating usable Visual Studio 2015 .androidproj projects via the FastBuild VCXProject Function…

missmah:
…. These generated .androidproj projects enable native C++ code debugging on attached Android devices from within Visual Studio 2015.

NOTE: This change is a superset of #238.

To generate a .androidproj via VCXProject, simply use the .androidproj extension instead of the .vcxproj extension inside your .bff code.

This change also adds a couple of new and needed features to VCXProj to support functional .androidproj generation:

    ".PackagePath" - This is the location of your APK file on disk [e.g. Output/ARM/ProjectName/APK/bin/ProjectName-debug.apk]
    ".LaunchActivity" - This is the activity name to launch when the APK is deployed. [e.g. com.android.NativeActivity]
    ".PlatformIncludePattern" - This is a string which can be used to specify which matching platforms should be included in the .vcxproj [e.g. ARM - would match ARM and ARM64]
    ".TargetArchitectureAlias" - This is a string which can be used to add an alias between a user-defined target platform (e.g. Android-ARM) and another MSBuild target platform (e.g. ARM)
    ".AdditionalSymbolSearchPaths" - This is a semi-colon delimited string fed to Visual Studio which is used to tell the .androidproj where to find debugging symbols for the code deployed inside the APK.
    ".AdditionalIncludePaths" - This was defined in #238 and is re-used here to give the android debugger access to paths where it can find source files for debugging.

This change also adds code to the solution generation (SLNGenerator.cpp) to use a different GUID {39E2626F-3545-4960-A6E8-258AD8476CE5} for referencing .androidproj(s).
I found this out the difficult way, by trial and error, using VS to make things work, and then looking what it did differently in its generated XML.

There is also the preliminary mention of .csproj here. This change could be extended to cover Visual Studio .csproj and .iosproj projects eventually.

Signed-off-by: Layla layla@insightfulvr.com

For reference, here is the commit message for #238:

Add support for new optional parameters: .TargetName, .TargetExt, .AdditionalIncludeDirectories, .WebBrowserDebuggerHttpUrl and .ProjectBuildType to VCXProject, and support CustomBuild steps on Utility projects for fast iteration.

This is an updated PR which cherry picks just the changes to these files, without the history that they had in the other PR. I will close out the other PR next.

Signed-off-by: Layla Mah layla@insightfulvr.com

This is a duplicate of https://github.com/fastbuild/fastbuild/pull/240, but against the dev branch.


missmah:
I'm not sure why it now shows three commits instead of just the one?

missmah:
@ffulin Does this branch need to be conflict free before it's eligible for merge, or can you resolve the conflicts on your end?

ffulin:
The conflicts look pretty manageable, the issue holding this up is just me finding the time to integrate because it's a non-trivial change that adds a lot of new options (all of which need documenting). I also need to make sure no regressions are introduced because it also changes the existing behaviour in a number of ways.  Sorry it's taking so long :(

missmah:
@ffulin I just added a couple more commits to this (sorry!) - they're super important for debugging C# projects generated by fastbuild inside visual studio, but the changes are highly related/correlated to the rest of this change, so I submitted them to the same branch.

Essentially, to properly get the VS debugger to debug a C# binary+pdb built by fastbuild, it's important to set the LocalDebuggerType = Mixed.

This change enables fastbuild VCXProject function to have the .LocalDebuggerType variable.

So, .LocalDebuggerType = 'Mixed', will generated <LocalDebuggerDebuggerType>Mixed</LocalDebuggerDebuggerType> in the .vcxproj, which will enable Visual Studio F5 debugger to properly load the PDB for the mixed code. Otherwise, debugging works, but the PDB refuses to load, and breakpoints won't work.

kostasvl:
Hi @missmah. I was looking for potential solutions to my issue (see #290), and I found your pull request. It kinda sounds from your posts like you added support for generating C# projects, but I can only find the android code (I see that writeCSProj variable, but it doesn't seem to be used). Am I missing something? :)  Thanks!

missmah:
@ffulin Is there any chance of this (and https://github.com/fastbuild/fastbuild/pull/242) making it into `0.95`? Is there any way I can help make the merge more manageable? It's painful to have to reverse-merge our changes back into `0.94` :-)

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -137,17 +224,51 @@ const AString & VSProjectGenerator::GenerateVCXProj( const AString & projectFile
...
+                    const ObjectListNode * oln = ProjectGeneratorBase::FindTargetForIntellisenseInfo( cIt->m_Target );
+                    if ( oln != nullptr )
+                    {
+                        AString my_compilerName( ( oln != nullptr ) ? oln->GetCompiler( )->GetName( ).Get( ) : "error: compiler unknown" );
+                        AString my_compilerOptions( ( oln != nullptr ) ? oln->GetCompilerOptions( ).Get( ) : "error: compiler options unknown" );
+
+                        my_compilerOptions.Replace( "%1", "%(FullPath)" );
+                        my_compilerOptions.Replace( "%2", "$(OutDir)%(Identity).obj" );
+
+                        Write( "        <Command Condition=\"'$(Configuration)|$(Platform)'=='%s|%s'\">@echo on &amp; SET %s; &amp; %s %s</Command>\n", cIt->m_Config.Get( ), GetPossiblyAliasedPlatformStringInUppercase( cIt->m_TargetArchitectureAlias, cIt->m_Platform ), environment, my_compilerName.Get( ), my_compilerOptions.Get( ) );
```
ffulin: environment is a double-nul terminated string, so this will only print the first entry in the environment list. Does this need to be processed somehow?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -137,17 +224,51 @@ const AString & VSProjectGenerator::GenerateVCXProj( const AString & projectFile
...
-            else
+
+            if ( writeCustomBuildCommands && !isHeader )
             {
-                Write( "    <CustomBuild Include=\"%s\" />\n", fileName );
+                // Build Command [Conditional]
+                Write( "        <Message>Compiling %%(Identity)</Message>\n" );
+                const VSProjectConfig * const cEnd = configs.End( );
+                for ( const VSProjectConfig * cIt = configs.Begin( ); cIt != cEnd; ++cIt )
+                {
```
ffulin: I'm a little confused about what this custom build command is for. In what situation is this used?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -86,6 +88,48 @@ void VSProjectGenerator::AddFiles( const Array< AString > & files )
     guid.Format( "{%08x-6c94-4f93-bc2a-7f5284b7d434}", CRC32::Calc( projectNameNormalized ) );
 }
 
+// IsFilenameAHeaderFile
+//------------------------------------------------------------------------------
+static inline const bool IsFilenameAHeaderFile( const AString & fileName )
```
ffulin: returning const bool won't compile with Clang.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -86,6 +88,48 @@ void VSProjectGenerator::AddFiles( const Array< AString > & files )
...
+{
+    // TODO: Add support for other header types?
+    return ( fileName.EndsWith( 'h' ) || fileName.EndsWith( ".hpp" ) || fileName.EndsWith( ".hxx" ) || fileName.EndsWith( ".inl" ) );
```
ffulin: Should the first check be EndsWith( ".h" )?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
missmah: This is for `CTRL+F5` iterative builds, where you can select a single source file to build via visual studio. This uses the build command directly to build that source file at the command line (no linking).

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
ffulin: Returning const bool won't compile with clang.
```C++
fatal error : 'const' type qualifier on return type has no effect [-Wignored-qualifiers]
```

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
ffulin: These are all case sensitive - I think they need to be case insensitive.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -86,6 +88,48 @@ void VSProjectGenerator::AddFiles( const Array< AString > & files )
...
+}
+
+static inline const bool ShouldWeSkipThisPlatform( const AString & platformIncludePattern, const AString & platform )
+{
+    // If we have a .PlatformIncludePattern filter defined, and this platform doesn't match it, return true [So we can skip emitting this platform to the .vcxproj]
+    const char *thisPlatformIncludePattern = platformIncludePattern.Get();
+    ASSERT( (thisPlatformIncludePattern != nullptr) && (platformIncludePattern != "") );
```
ffulin: An AString can never return a nullptr, so asserting on that is not needed.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
missmah: It sure looks like it! :)

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
ffulin: platformIncludePattern is empty by default, so this looks lik it would assert by default.
Seems like this function should just return false if that string is empty?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -86,6 +88,48 @@ void VSProjectGenerator::AddFiles( const Array< AString > & files )
...
+{
+    // Handle Platform Aliases - these define new platforms and alias existing platforms to them
+    if( !targetArchitectureAlias.IsEmpty() )
+    {
+        const char *thisPlatformAlias = targetArchitectureAlias.Get();
+        ASSERT( (thisPlatformAlias != nullptr) && (targetArchitectureAlias != "") );
+        AStackString<>							thisPlatformAliasString( thisPlatformAlias );
+                                                thisPlatformAliasString.ToUpper();
+        const char *thisPlatformAliasUpper =	thisPlatformAliasString.Get();
+        return thisPlatformAliasUpper;
```
ffulin: This is returning a pointer into the stack that falls out of scope when this function returns. This is definitely a bug.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
missmah: Another good catch - would that be `EndsWithI`? I think I wrote this code in a bit too much haste given that I'm not an expert at the codebase :/

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -86,6 +88,48 @@ void VSProjectGenerator::AddFiles( const Array< AString > & files )
...
+    }
+    return platform.Get();
+}
+
+static inline const char* GetPossiblyAliasedPlatformString( const AString & targetArchitectureAlias, const AString & platform )
+{
+    // Handle Platform Aliases - these define new platforms and alias existing platforms to them
+    if( !targetArchitectureAlias.IsEmpty() )
+    {
+        ASSERT( (targetArchitectureAlias.Get() != nullptr) && (targetArchitectureAlias != "") );
```
ffulin: This assert can never fire - AString never returns nullptr, and the IsEmpty() checkes means it can never be == ""

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -102,26 +146,69 @@ const AString & VSProjectGenerator::GenerateVCXProj( const AString & projectFile
...
+    const bool writeAndroidProj = projectFile.EndsWithI( ".androidproj" );
+
+    // Don't write a custom build command by default, because as of Visual Studio 2015 R3, Makefile projects can't currently utilize this functionality
+    // However, if we have a Utility project, then we can execute per-file custom build steps
+    bool writeCustomBuildCommands = false;
+
     // header
     Write( "<?xml version=\"1.0\" encoding=\"utf-8\"?>\n" );
-    Write( "<Project DefaultTargets=\"Build\" ToolsVersion=\"4.0\" xmlns=\"http://schemas.microsoft.com/developer/msbuild/2003\">\n" );
+    Write( "<Project DefaultTargets=\"Build\" ToolsVersion=\"14.0\" xmlns=\"http://schemas.microsoft.com/developer/msbuild/2003\">\n" );
```
ffulin: I wonder where this 4.0 came from originally.  That certainly seems wrong.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -253,12 +453,28 @@ const AString & VSProjectGenerator::GenerateVCXProj( const AString & projectFile
...
             WritePGItem( "NMakeCleanCommandLine",           cIt->m_CleanCommand );
             WritePGItem( "NMakeOutput",                     cIt->m_Output );
+            WritePGItem( "TargetName",                      cIt->m_TargetName );
+            WritePGItem( "TargetExt",                       cIt->m_TargetExt );
+
+            AStackString<> vsIncludeSearchPath;
+            if ( !cIt->m_AdditionalIncludePaths.IsEmpty( ) )
+            {
+                vsIncludeSearchPath = cIt->m_AdditionalIncludePaths;
+                vsIncludeSearchPath += "$(IncludePath);";
```
ffulin: Why is $(IncludePath) automatically added?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
missmah: It sounds like the `!= ""` check is not needed either because it's the same as the `.IsEmpty()` check?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
```diff
@@ -86,6 +88,48 @@ void VSProjectGenerator::AddFiles( const Array< AString > & files )
...
+    return ( (!platformIncludePattern.IsEmpty()) && (!platform.Find( thisPlatformIncludePattern )) );
+}
+
+static inline const char* GetPossiblyAliasedPlatformStringInUppercase( const AString & targetArchitectureAlias, const AString & platform )
+{
+    // Handle Platform Aliases - these define new platforms and alias existing platforms to them
+    if( !targetArchitectureAlias.IsEmpty() )
+    {
+        const char *thisPlatformAlias = targetArchitectureAlias.Get();
+        ASSERT( (thisPlatformAlias != nullptr) && (targetArchitectureAlias != "") );
```
ffulin: Assert can never fire (AString never returns nullptr and we already checked the string is not empty)

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
ffulin: While this might work (in that the file will compile) it's not really safe - i.e. you can't mix and match compiling file directly outside of FASTBuild and then using FASTBuild.  Best case it will rebuild when you switch back to a proper FASTBuild build, but worst case you will get weird failures.  I have also had problems with Visual Studio project conditionals (not working properly when switching between build configurations, only being re-evaluated when quitting VS and reloading the sln).

I'd like to explore getting Ctrl+F5 working, but if we could decouple this from the Android support that would be good.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
ffulin: Yeah, EndsWithI :)

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
ffulin: I think at the moment this assert will just fail, because the string is empty by default (see next comment)

ffulin:
I definitely feel bad about how long these two changes have been sitting around. The problem is that they both change the project generation a lot and in really non-trivial ways.  That makes me really nervous to integrate them.

If we could do two things, I think we could start to get these sorted out:
a) split them up into smaller changes so we can integrate the easy/safe bits without holding up the more complex parts (for example, Ctrl+F5 seperate from android support)
b) keep pure functional extensions separate from behaviour changes (many of the new properties you added only change the output if the new property is specified) These could be trivially integrated without holding up the other changes.

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
missmah: Maybe a typo where the 1 was accidentally deleted?

Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.cpp
missmah: Thanks - I'll remove all the `== ""` and `nullptr` asserts - they were there because I was writing code quickly and wanted to know if my assumptions were correct or not about how your string classes worked, without taking the time to read all of the code. I think I planned to come back and review it later and it seems I never did. Sorry :(

missmah:
I meant to work on this more today; but, got swamped by tons of other things and didn't get back to it. I'll try to implement these changes soon. Thanks for the review!

missmah:
@kostasvl It's been a long time since I've looked at this code; but, I believe this statement I made back when I created the PR is still accurate, unfortunately: "There is also the preliminary mention of .csproj here. This change could be extended to cover Visual Studio .csproj and .iosproj projects eventually."

But, on a positive note, I don't think it will take much to extend the changes here to properly support C# and iOS projects. I did already add some support for debugging C# properly via a generated .vcxproj which contains projects that happen to build some C# code via fastbuild - so that part is already supported.

kostasvl:
@missmah Thanks, that's useful to know.

ffulin:
This PR contains a lot of useful stuff, but is unfortunately: 
 - very much out of sync with the current state, with many large conflicts
 - contains incomplete and commented out code

I'd still like to see these changes get integrated. It we could split the change, at least dividing it into:
 - exposing new, purely opt-in functionality
 - behavioral changes that might impact existing code
then I think we will have more success getting it merged.

PR #252, opened 2017-06-03
always use absolute dir path of config file as working directory

comicfans:
it brings two benefits: 1.we can ensure relative path in bff always
resolved to same file , no matter which current folder we run fbuild.
2.always create same global process lock when different instance build
same bff,even they run from different dirs. without this, second instance will
report 'open file error' instead of 'another instance already running'

this implementation didn't handle relative path with '.' or '..', which I think should be part of 
FixupFilePath or maybe a new path normalize function.

### DIFF
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -194,4 +194,28 @@
     }
 }
 
+// SplitDirFileName
 //------------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


ffulin:
I am a worried that this change in behavior will break things for existing users. I think we'd need to have some way to opt-in to this alternative behaviour (a command line option for example).

comicfans:
how about add -unstablepath to keep old behavior ? 
this change only break behaviour when relative path is written against running dir instead of bff dir,  which I think this is rare . and this change allow running multi instance to build different bff in same dir, or provides better diagnostic when building same bff from different dir.

PR #253, opened 2017-06-05
Support flattening nested libraries when creating static library

jblazquez:
When a `Library` target uses `LibrarianAdditionalInputs` to reference other static libraries, the librarian tool is invoked with those static libraries as inputs instead of the object files contained in those libraries. This is OK on Windows because the MSVC librarian flattens the list and creates a new library with only object files in it, but with GNU `ar` that's not the case [without some additional help](https://stackoverflow.com/questions/3821916/how-to-merge-two-ar-static-libraries-into-one/23621751#23621751) and the resulting library contains nested libraries that can cause problems during linking.

This change extends the existing logic of flattening nested object lists to also handle nested libraries.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
@@ -14,6 +14,7 @@
 #include "Tools/FBuild/FBuildCore/FLog.h"
 #include "Tools/FBuild/FBuildCore/Graph/CompilerNode.h"
 #include "Tools/FBuild/FBuildCore/Graph/DirectoryListNode.h"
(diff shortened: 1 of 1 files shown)


ffulin:
I've integrated your change (https://github.com/fastbuild/fastbuild/commit/5c426a8c00cd116b3da003bcc117980622fe5521) but I needed to make few small changes:
 - libs being merged no longer use object files when using the MSVC librarian
 - linking resepects the .LinkerLinkObjects flag once again
Let me know if the changes seem ok for you.

This will be in the next release (v0.94).

jblazquez:
The changes sounds great, thanks!

ffulin:
v0.94 has now been released and includes this change as discussed.

PR #257, opened 2017-06-07
Fix CopyDir failing when trying to copy files that had been deleted.

kwayle:
DynamicDependencies was never cleared so it would still list files that had been previously copied but since deleted.

Add test case to catch this regression in the future.


ffulin:
Thanks for the fix (and in particular the tests to go with it). I've integrates the fix into the Perforce mainline. The fix will be in the next release (v0.94)

PR #260, opened 2017-06-27
Add optional .CompilerFamily option to the Compiler node, so that we …

missmah:
…can explicitly specify the compiler type (e.g. Clang) from within the .bff file.

This is useful for i.e. Android and Emscripten - both of these use Clang back-end but do not have the normal clang.exe binary name for the compiler (e.g. Emscripten is emcc).

Signed-off-by: Layla Mah <layla@insightfulvr.com>


ffulin:
Looks pretty good. Thanks for the pull request.

ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.95). Thanks for the pull request!

I made some further changes/modifications/enhancements: (https://github.com/fastbuild/fastbuild/commit/c3f60a2792adf51777fb0845424d5835ac4b2c56)
 - I changed the valid strings to use hyphens instead of underscores to match the similar LinkerType (i.e. "qt-rcc" instead of "qt_rcc")
 - I added "custom" as a valid value, which allows you to specify to FASTBuild that you know it is not a supported compiler and only basic dependencies (in/out) will be tracked.
 - An error is reported if auto-detect can't determine the compiler automatically (#1500)
 - An error is reported if an invalid value is specificed for .CompilerType (#1501)
 - I moved the detection logic into the CompilerNode (something I wanted to do for a while)
 - Documentation

PR #261, opened 2017-06-27
Added -showdeps command to print out the dependency graph in a primit…

missmah:
…ive manner.

Signed-off-by: Layla Mah <layla@insightfulvr.com>

### DIFF
--- a/Code/Tools/FBuild/FBuildApp/Main.cpp
+++ b/Code/Tools/FBuild/FBuildApp/Main.cpp
@@ -110,6 +110,7 @@ int Main(int argc, char * argv[])
     bool waitMode = false;
     bool noStopOnError = false;
     bool displayTargetList = false;
+    bool displayDependencyDB = false;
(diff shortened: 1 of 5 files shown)


ffulin:
Nice work. I think I'll extend this slightly so that if you provide a target on the command line, it prints the dependencies only for that sub-tree.

ffulin:
I've integrated this change into the Perforce mainline. This will be in the next release (v0.94). Thanks for the pull request, I think this will come in handy!

I also made a few improvements (https://github.com/fastbuild/fastbuild/commit/e576c79fc38470bfe98ed8409b1979342b0e8e28):
 - show "..." when dependencies are not traversed due to already being seen
 - respect targets on the command line so you can specify what you want to see dependencies for, defaults to "all" just like building
 - format via a temporary buffer to massively speedup output (as much as 40x)
 - change recursion logic to avoid redundant printing of leaf nodes at end of output

PR #269, opened 2017-08-01
Refactored "did you mean" feature for better results

poppolopoppo:
I wasn't satisfied with my previous pull request, which was rarely useful.
This new version is very close to what git does, and should be more pertinent hopefully.


Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -1413,90 +1413,75 @@ Node * NodeGraph::FindNodeInternal( const AString & fullPath ) const
...
 
-    for ( size_t i = 0 ; i < NODEMAP_TABLE_SIZE ; i++ )
+    const size_t totalNodes = GetNodeCount();
+    for ( size_t i = 0 ; i < totalNodes ; ++i )
     {
-        for ( Node * node = m_NodeMap[i] ; nullptr != node ; node = node->m_Next )
+        Node * node = GetNodeByIndex( i );
+        bool displayName = false;
+        switch ( node->GetType() )
         {
```
ffulin: What's the thinking behind limiting to certain node types? While the ones you limited the search to are the most useful, I could see the others still being useful. Was is just to de-clutter the output or was it for performance reasons?

poppolopoppo:
Sounded better to mimic the behavior already implemented to list targets in FBuild::DisplayTargetList().
https://github.com/fastbuild/fastbuild/blob/master/Code/Tools/FBuild/FBuildCore/FBuild.cpp#L656

But I can add back every node types you'd think useful :)

ffulin:
Hmmm... it would be a little weird for them to not be consistent. I might need to think about this for a little bit.

Code/Core/Strings/LevenshteinDistance.cpp
```diff
@@ -37,89 +37,140 @@ namespace
 }
 
 // Levenshtein distance
-// https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Levenshtein_distance#C
+// https://github.com/gitster/git/blob/master/levenshtein.c
```
ffulin: I took a look at the licensing terms of this source code and it seems to be GPLv2. Unfortunately, that makes it incompatible with the more permissive licensing of FASTBuild's (which for example allows for commercial use without source code release).

PR #275, opened 2017-08-25
Change boolean IdleDetection to continuous.

Nesquick0:
I'm not fan of boolean (on/off) idle detection on workers, because I don't have dedicated machines as compilation helpers.

So I tried to create something continuous so if user/machine is using 30% cputime there would be for example 5 of 8 idle.
I was trying to keep IdleSmoother behavior so it doesn't change too much.


ffulin:
The "When Idle" setting is designed for use in the non-dedicated worker scenario. The idea is that you don't want to impact the user while they are using their machine. The "Always" setting on the other hand is for dedicated machines.

If a user's machine is at 30% load and you allow FASTBuild to to use the other 70%, the user will see a significant degradation in their machine performance. As such, I won't integrate the "When Idle" setting as you've changed it.

I wouldn't be opposed to integrating a new setting however, that behaves in a scaling fashion along the lines of what you've implemented.

PR #278, opened 2017-08-29
Linux: Transfer file permissions on copy

jblazquez:
The `CopyFile` and `CopyDir` functions weren't properly setting file permissions on the destination files to be the same as the source files on Linux.


ffulin:
Thanks for the fix! (And double thanks for extending the tests!) I've integrated this into the Perforce mainline and this will be in the next release (v0.94)

PR #279, opened 2017-08-30
Linux: Fix transfer of permissions on file copy for CIFS shares

jblazquez:
This change fixes a problem with #278 where destination files would not be created due to access denied errors when the destination is a CIFS share. It looks like Samba fails when setting the file mode (for some file modes only, e.g. `0555`) during an `open` call but it "succeeds" (see below) if done during a `chmod` call, so we use that now instead.

Note that in either case, if the SMB server is a Windows box then the actual file permissions are ignored (see FILE AND DIRECTORY OWNERSHIP AND PERMISSIONS [here](https://www.samba.org/samba/docs/man/manpages-3/mount.cifs.8.html)).

### DIFF
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -180,10 +180,19 @@
     struct stat stat_source;
     VERIFY( fstat( source, &stat_source ) == 0 );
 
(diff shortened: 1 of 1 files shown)


ffulin:
I've integrated this into the Perforce mainline. This will be in the next release (v0.94)

PR #283, opened 2017-09-05
Add "Work when available" as continuous idle detection.

Nesquick0:
Add "Work For Others When Available" as continuous idle detection which is something between Idle and Always to use PC power even when user consume more than 10% CPU.

Not sure about name of new worker state. "Available" was the best one I came up with.

Original pull request: #275

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/options.html
+++ b/Code/Tools/FBuild/Documentation/docs/options.html
@@ -143,7 +143,7 @@
 
 <table width=700>
   <tr>
-    <th width=250 align=left>Option</th>
+    <th width=320 align=left>Option</th>
     <th align=left>Summary</th>
   </tr>
   <tr>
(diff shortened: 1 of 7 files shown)


Nesquick0:
After long time I've updated this to work with your changes in v0.95.

One thing I noticed what v0.95 changed in IdleDetection.cpp that:
`if ( m_IdleSmoother == 10 ) // 5 secs (called every ~500ms)`
was 50 before. But there is also Clamp before this:
`m_IdleSmoother = Math::Clamp(m_IdleSmoother, 0, 50);`

I think this isn't correct, so I changed that in my commit to Clamp 0 to 10. Because with 50 in clamp, that means when PC is idle for long enough to reach 50 in m_IdleSmoother, and then it would take 25x0.5s to get to 0 which is 12.5s. That seems to me not what you wanted.

Nesquick0:
Sorry, I've done it again. I was trying to update my Dev branch to match your current Dev branch. Is it possible to reopen pull request?

Also I expect you to be busy and don't have time to get through each pull request, but is there way how to make this merged please? I changed things you mentioned last time and check regularly that my branch is compatible with yours.

ffulin:
I wasn't able to re-open this pull request (github complains that "The dev branch was force-pushed or recreated"). I created a new pull request: https://github.com/fastbuild/fastbuild/pull/554

PR #288, opened 2017-09-13
Output remote errors to monitor log

hillin:
The title is self-explanatory.


ffulin:
I've integrated this fix into the mainline. This will be in the next release (v0.96). Thanks for the fix!

Issue #290, opened 2017-09-21
C# project to go with the CSAssembly?


kostasvl:
Hi. I wanted to ask what the intention is with regards to CSAssembly, in terms of having a backing C# project for the .cs files, in the same Visual Studio solution as everything else. I couldn't find a way to do it. I'm probably missing something, because CSAssembly doesn't seem as useful as it could be, if I'm better off having a separate solution for my C# code (to get the best editing experience of course).

Thanks!
Kostas

ffulin:
FASTBuild has a distinction between the config used to define compilation of something (Library(), CSAssembly() and others) and the way to integrate that in Visual Studio. (So that you can always compile from the command line, even if VS is not installed - as long as you have the compiler available).  With this in mind, the VCXProject() can be used to create a project that will compile your C# code.

Alternatively, if you prefer to not have FASTBuild manage the C# part of your compilation, you can add manually created projects into your solution.

C# is not widely used by FASTBuild users, so it's possible that either of those approaches has limitations or problems for you. I'm open to suggestions about how to improve support!  (I see that you added a pull request, but I didn't get a chance to go over that yet)

kostasvl:
"...With this in mind, the VCXProject() can be used to create a project that will compile your C# code."

I assume that would be a C++ nmake project with the .cs files included? A google search makes me think nmake projects are only meant for C++. It would be nice if we could get intellisense for C#, breakpoints during debugging, Edit & Continue, etc. I only seem to get syntax highlighting, after trying it quickly. Anyone with prior experience?

"Alternatively, if you prefer to not have FASTBuild manage the C# part of your compilation, you can add manually created projects into your solution."

We currently use 'fbuild.exe projects' to generate our solution. We could switch to using a solution created in the traditional way, which would manually include the vcxproj files created by FB, and the csproj files we created. Build order between cross-language projects could be handled via solution-level dependencies. I tried this quickly just now, and it seems to work - at least from within Visual Studio. Not sure why I didn't think of this sooner :). I'd have to see how it will behave when building from the command line (via devenv.exe from now on, I guess), but it sounds very promising.

Thanks @ffulin!

ffulin:
I'm definitely open to ideas for improving the C# integration. Honestly, C# is not use a lot by FASTBuild users, so the support is very basic/limited at the moment.

[...]

PR #291, opened 2017-09-22
Referencing external (non-FASTBuild generated VS projects) in new VSProjectExternal function

lucianm:
Hi,

in our team which is gladly adopting FASTBuild via the VS-generated projects more and more we also often use non-VC++ projects our VS Installation, augmented with third party extensions understands (for example Python, WiX toolset to name only two).

If we want to have those in the same solution as the one generated by FASTBuild, the ~~VCXProject~~ **_new VSProjectExternal_** function needs to be able to reference external, existing projects without overwriting them. This pull request adds exactly this enhancement.

To use it, ~~the correct project GUID, as well as~~ the correct project type GUID (necessary for VS to correctly load those projects from the generated solution) ~~have~~ has to be specified, ~~and nothing more but the external flag~~ as in the example below:
```diff
-VCXProject( 'ExtProject-vsproj' )
+VSProjectExternal( 'ExtProject-vsproj' )
{
-    .ProjectIsExternal = true
-    .ProjectOutput     = 'Setup\Source\ExtProject\ExtProject.wixproj'
+    .ExternalProjectPath     = 'Setup\Source\ExtProject\ExtProject.wixproj'
-    .ProjectGuid       = '{32aa1ad4-da90-48b6-8d09-84bc693734d2}'
    .ProjectTypeGuid   = '{930c7802-8a8c-48f9-8165-68863bccd9dd}'       // set the correct WiX Toolset project type GUID
}
```

- ~~_ProjectOutput_~~ _ExternalProjectPath_ matches the actual project file just ~~as in the case of generated (the default .ProjectIsExternal = false) projects~~ which has to exist;

~~- _ProjectGuid_ becomes mandatory and **should** match the actual project GUID of the referenced Project (should - because FASTBuild will not check the match, but the resulted solution will be inconsistent and VS will not be too happy loading it);~~
- _ProjectTypeGuid_ **should** actually be provided ~~unless the Project is of the type VC++ which FASTBuild already knows internally~~.

Best regards,
Lucian


kostasvl:
Hi @lucianm. Thanks for this :). It could be a potential solution to my issue (see #290). I assume you use the VCXProject function's .ProjectProjectReferences property to add any C++ ==> Python dependencies, or you add those project references manually for the Python ==> C++ dependencies. If I do the same with my C# projects, I guess I won't have to use the CSAssembly function. Do you have any thoughts?

Also, may I suggest an improvement :)? I think a new ExternalProject function might be more discoverable and more self-documenting.

lucianm:
Hi @kostasvl, I do not use .ProjectProjectReferences in this case, as the added feature does not touch the referenced ("external") Project. However, dependencies could be added manually to it as references, from the VS GUI interactively, then they get saved to that Project and not as solution dependencies. The only Thing with references is that every different Project type seems to understand something slightly different under the concept of project references. **ProjectProjectReferences** is bound to VC++/NMake Projects.

Having a completely new function ExternalProject might be a good idea, as it would indeed be more generic and possibly more discoverable and documentable as you suggest, besides VCXProject seems to be more C(++)-centric. Let's hear what @ffulin  has to say about it.

ffulin:
I'm a little confused as to why this is needed. The .ProjectProjectReferences property is designed to allow a FASTBuild generated project to depend on another project not generated by FASTBuild - FASTBuild does not touch that project in any way in this case.

This functionality was initially added for the use case of a FASTBuild generated project depending on an already existing .csproj.

The warnings you get about "project not selected for this configuration" are normal (sadly there is no way to suppress those) and while inconvenient are nothing to be worried about. There is no way to avoid msbuild being invoked when compiling from VS, but if the project dependencies are set correctly, it will first invoke your non-FASTBuild projects to build first, and then the FASTBuild project you have set as the master project (that is the one that needs to depend on the other project(s)).

All of this is pretty messy, but it seems this is the best we can manage mixing non-FASTBuild projects (it might be possible to achieve better integration with a Visual Studio plugin in the future).

So in summery, I feel like maybe I'm missing something about your use case, because I don't think the functionality in this pull request is needed.

lucianm:
Hello Franta,

.ProjectProjectReferences just adds references as dependencies, but does not add such a project to a solution generated by FASTBuild, that's what this new functionality is needed for, to be able to maintain a FASTBuild-generated solution even including "foreign" projects in it, at the place one wants to actually see them organized in virtual solution folders maybe, in the solution, ready to be edited from within Visual Studio. Without it, those would have to be added manually and then they would get lost when the solution is generated again by FASTBuild and would have to be added again manually and organized. I hope this clarifies things.

lucianm:
@ffulin  hehe, maybe my subject line for this pull request was misleading as well :-)

ffulin:
@lucianm Thanks for the extra context - this issue makes sense now. Mixing FASTBuild and "foreign" projects indeed only really works right now if you make the .sln file by hand. Making it so you can use these other projects in a generated sln is absolutely desirable.

Rather than extend VCXProject in this way, or create a new ExternalProject, I think it might be ideal if instead the VSSolution simply accepted external projects and tried to automatically extract the necessary information (guids) out of them. Do you see any drawbacks or issues with this approach?

lucianm:
@ffulin  Well, how do you mean the VSSolution accepting external projects and try to automatically extract the GUID information from the foreign project, like in the process of generating it, or leave it to the loading mechanism of VS do it on first loading that solution? If you mean to parse the XML of the foreign project when referencing one in a solution to be generated, then well, the ProjectGuid should always be there, but the ProjectTypeGuid is not found in the project very often, as it seems. A project type may have even more than one Type Guid, the XML tag for that is even called _<ProjectTypeGuids>_ but as mentioned, it is not present in every project, it's the system registry where they're all maintained by the vanilla VS installation and any other extensions which may add new project types.
If you on the other hand meant something like letting VS figure them project type Guids out on first time loading such a solution, AFAIK it won't even load those projects as it does not know what type they are and what extensions to use for them. It's a pity, as it's as well VS (or MSBuild) which has this information available, since it finds it and writes it to the solution when interactively adding a project type known to the current installation to the solution.
This is why I thought the type of implementation I've chosen is the more simple and effective one under the given circumstances.

lucianm:
Hello Franta,

I rebased my pull request again (there wasn't really any other conflict to solve than the one-line keyword enumeration in the Notepad++ markup file, really annoying that notepad++ expects those in just one line so regular merge tools are having a hard time) :-)
So what do you think of the feature, still reluctant of integrating it, because of the approach I used? I did not exactly understand what approach you were suggesting instead.

lucianm:
Hello Franta, I rebased my pull request on top of v0.97 and adapted it to the reflection macros. Would be awesome if you would adopt it.

ffulin:
Hello @lucianm 

I've been thinking about the possible ways to solve this problem. I'm not set on a particular solution, as I don't see a super clear best path forward. Here are the 3 options I can imagine right now:

**1) Extend VCXProject()** - This is the pull request you've made.
Pros: 
 - You did it and it works
Cons: 
 - It's a little odd for a node to sometimes generate a file, and sometimes reference an existing one
 - VCXProject is very complex, and this makes it more-so
 - It's a little strange to have a single toggle that disables most of the functionality
 - Manually specifying the guids seems painful (though this could potentially be remedied)

**2) Make VSSolution() automatically support external projects** - This was the solution I alluded to in my previous comments, but never fully explained. The idea is that is VSSolution could reference a project that is just a file (and not a VCXProject) and when it sees this, i would open the existing file and parse out the needed information (like the guids) and "just work".
Pros:
 - Easier for end user
Cons:
 - Might have some complexities that are not foreseen
 - Might be considered a bit "magic"

**3) Make a new VCXProjectExternal** - We could make a new project that contains the functionality you have provided.
Pro:
 - Can leverage some of what you created
 - Breaks apart unique functionality
 - Is very explicit
Cons:
 - Might end up causing some duplication/special casing in VCXProject and VSSolution to handle

I'm leaning towards option 2), but I'm concerned there might be some complexities I'm missing. The hesitation is that whatever we decide, we'll be stuck with for a long time, so I want to make sure we get it right.

I'd love to get your thoughts on the relative merits of these approaches given all the time you've spent thinking about this problem. Let me know what you think.

lucianm:
Hello @ffulin,

it took me some time to read github updates, and then some more to find a moment to answer...

I would love to see FASTBuild having a solution like described in option 2) for this, but as far as I can imagine it right now, the automatism could only be realized in parsing the project GUID from the existing project, but not the project _**Type**_ GUID , wich is needed by Visual Studio to be able to load a certain type of project, especially those it "learned about" only after some extension or addon has been installed. It seems it need to know this type GUID to be present in the solution, in order to manage to chose the right project properties "renderer". It is not common that these project type GUIDs are present in the project files themselves, although possible. The only user-friendly way to go with option 2) I think is providing some additional helper tool or better, a DLL written in .NET which will query directly the Visual Studio installation to provide the project type GUID of a certain project, I found some code going into that direction [here](https://www.mztools.com/articles/2007/MZ2007016.aspx), maybe it would also work with recent versions of Visual Studio, I'll investigate as time will allow me to.  Of course, it would be a Windows-Only solution to the problem, as these .NET services might only be available there, but then again, one would only need this on Windows.

If this option does not prove to be feasible, re-implementing the technique used in my current PR into a VCXProjectExternal shouldn't be that difficult.

So, what do you think about such an approach? It will be kind of "magic", that's for sure ;-)

ffulin:
HI @lucianm 

Thanks for the extra info. I was only thinking about the project GUID and not the project **type** GUID. With this new information in mind, I think perhaps option 3) is the way to go, doing something like the following:

**1) Create a new Function: VSProjectExternal** This can create a new VSProjectExternalNode which exposes a property which is the path of an existing project. It can also expose any other properties we need to control or hint how this project should be used (see below)

**2) For the project GUID:** In the DoBuild of this new VSProjectExternalNode, we can open the existing file on disk and extract the project GUID. The GUID can be saved in the node as a MetaHidden member, so that we only do this parsing when the external project changes.

**3) For the type GUID** We can attempt to detect the project type GUID, either by parsing the project or based on the file extension (or a combo - whatever seems more reliable). At least for the "normal" types, I think this should be possible. We can handle only the simplest cases to begin with, and add better detection logic or support for more types later if needed. to handle exotic types (say a custom project type) we can expose a .ProjectTypeGUID property that disables the basic auto-detection and instead allows you to explicitly specify the type GUID.  As with 2), we can save the auto-detected type GUID in a hidden property to avoid re-parsing the file when it hasn't changed.

**4) Update SLN** We can make the appropriate input properties on Solution() that take a VCXProject also accept one of these new VSProjectExternal objects as inputs. We can insert a common base class for VCXProjectNode and the new VSExternalProjectNode (VSProjectBaseNode perhaps) to abstract away which one is being used, providing a consistent API for things like GUID retrieval.

What do you think? Does this cover all the cases?

lucianm:
Hi @ffulin,

yes, that looks like a really good specification to me. It would still optionally allow for the .ProjectTypeGUID auto-detection to involve dyn-loading a DLL if available (not necessarily part of FBuild), which could extract the actual type-guid using the Visual Studio .NET SDK, if guessing from a file extension based lookup fails. If you agree, I'll refactor my PR again in the near future, according to what we discussed.

Best regards,
Lucian

ffulin:
Sounds good to me. Let me know if you run into any troubles implementing it.

lucianm:
Hi @ffulin,

I refactored the pull request according to your proposal, for now without yet parsing the ProjectGuid into a MetaHidden member (position 2.) and without any sort of project type detection (position 3.), because even this refactoring took me some efforts while being really busy with other things, so the rest will have to wait a bit now, but for sure will follow soon.
Nevertheless I thought I'll share already my current state (which is in fact resembling the exact same functionality as the initial attempt, so only the candy stuff 2. and 3. is missing now), so maybe you will have a look at it already and give some feedback.

The new VSProjectExternal for now has 3 mandatory input options:
```
VSProjectExternal( alias )               // (optional) Alias
{
  // Input Options
  .ExternalProjectPath            // Path to project file
  .ProjectGuid                    // Project Guid exactly as found in the actual external
                                  // project (has to be consistent)
  .ProjectTypeGuid                // Project Type Guid, please read the details below
}
```

for example:

```
VSProjectExternal( alias )               // (optional) Alias
{
  .ExternalProjectPath = 'Setup\Source\ExtProject\ExtProject.wixproj'
  .ProjectGuid = '{32aa1ad4-da90-48b6-8d09-84bc693734d2}'
  .ProjectTypeGuid = '{930c7802-8a8c-48f9-8165-68863bccd9dd}' // WiX Toolset project type GUID
}
```

Both nodes VCXProjectNode and the new VSProjectExternalNode now share a common base class VSProjectBaseNode which holds all of the configuration data, but only persisatance / reflection is distributed respectively. I thought that was the most straightforward way in order to not mess with the Array<VSProjectConfig> too badly since it has to be propagated in various places.

There is also a basic test included (working), along with documentation, syntax integration of the few new keywords .

I will switch our build systems to this new refactored version in the following days and let you know if there are any problems.

Maybe I'll manage to have a look into providing a Visual Studio SDK based module for pos 3.), I'll let you know as soon as I have something, or find out that it's maybe too complicated.

lucianm:
@ffulin 

Ok, the `.ProjectGuid` is gone (hidden) now in VSProjectExternal, I added some ugly parsing from the project file :-)

lucianm:
@ffulin 

So, it was not as difficult as I first thought to write a [VSProjTypeExtractor](https://github.com/lucianm/VSProjTypeExtractor), I can now further add to this pull request linking against it on Windows and if `.ProjectTypeGuid = 'auto'` is found, it would be used to automatically determine the project type GUID. As this actually is done by automating a Visual Studio instance in the background for loading that project into a volatile solution, just for querying the type GUID, it might take several seconds for each project. What do you think, if I store some file hash to the database I could skip calling the DLL code if the hash is not changed, in order to save time?
Looking forward for your opinion.

lucianm:
@ffulin 
Ok, so I did not link against [VSProjTypeExtractor](https://github.com/lucianm/VSProjTypeExtractor), but just added its export header file to the "external" directory, it provides convenience code for loading the library at runtime, so the actual code in VSProjectExternalNode.cpp is minimal invasive. It also passed all tests for me, it successfully extracts project type Guids for external projects, by instantiating a configurable VS version under the hood, load the projects and returning the type Guids. Instanciating may take around 9-12s, depending on configuration, but after that it scales pretty well, processing a project is done very quickly.
The fact that it needs 2 external DLLs and a VS installation is also the reason why Travis CI checks now failed, maybe the ExternalProjects test should be actually commented out, disabled.
I've been working a lot at this refactoring and the external module, it would be nice if we could discuss the current implementation rather soon (while I'm still fresh at it) and hopefully this time you can adopt it in mainline FASTBuild.
Awaiting your feedback ;-)
Greetings, Lucian

ffulin:
I've taken a fresh look at this PR and have summarized what I see as the remaining issues below. They vary in complexity and importance. I'll list them in no particular order and provide thoughts/suggestions about how we can resolve them for you to consider.

I appreciate all the effort you've put in so far, and it'd be great to finally get this sorted out and removed from the PR backlog!

Please let me know what you think and let me know if anything I wrote doesn't make sense :)

<h2>1) DoBuild implementation</h2>

In the DoBuild function of VSProjectExternalNode, a hash of the file contents is calculated/stored into m_FileChecksum and the m_Stamp is always left at zero. FASTBuild's dependency graph expects (for externally provided items which are files) that the m_Stamp contains a representation of the file state, which is usually the modification time. This is used to determine if a build is out-of-date or not and can have other special meaning. Except for rare cases, DoBuild should not execute when there is no work to do. Currently only things that generate output should run every build, and even that will change over time.

The implementation as it stands requires the external projects be loaded/hashed every time and that DoBuild be called every time. It duplicates a subset of the "do i need to be built" logic that is provided by the DependencyGraph. In addition to unnecessary rebuilding, -verbose output is inconsistent and the build statistics are potentially incorrect.

**Proposed solution:** Remove the file hashing logic, assume DoBuild is called only when the file has changed and set m_Stamp at the end of DoBuild by calling the RecordStampFromBuiltFile() function. Optional: Extend the tests to validate that the nodes are not built when nothing has changed.

<h2>2) VSProjectExternalNode exposes many unused properties</h2>

VSProjectExternalNode derives from VSProjectBaseNode, which in turn exposes all the properties that can be used to configure a generated project. These properties can therefore be set on VSProjectExternalNode, but they don't do anything. This could be confusing (to users and to someone working on the cdoe). It's also inefficient (many unnecessary properties are serialized and modifying them will cause unnecessary rebuilds).

I'm sorry about this - this confusion is probably my fault. I didn't consider that you might interpret my suggestion of "adding a base class" as referring to this. This class is unrelated to the base interface I think we need to add that VCProjectNode and your new node derive from. For simplicity, let's not rename the existing structure even though we might want to do that later to disambiguate things.

**Proposed Solution:** Revert the changes/move of this helper structure and create a new base class which implements only the virtual functions that Solution() needs (like GetProjectGUIDType())

<h2>3) Tests</h2>

As you noted, the tests are failing. We'll need to find a way to have the tests pass out-of-the box for end users and on the CI without a dependency on external/optional components.

The test also doesn't validate that the project nodes in the test were actually built.

**Proposed solution:** Modify the test to validate the built-in "auto" behavior and remove the need for the external dependencies. Add a "node count" check for the 4 nodes being referenced.

<h2>4) Dependency on external VspteModuleWrapper</h2>

I've gone to great lengths to keep FASTBuild free of 3rd party code, and with the exception of 2 things which are entirely embedded inside FASTBuild, there are no external dependencies (ignoring platform SDKs/compilers etc of course).

**Proposed Solution:** This is a philosophical concern, but there are some practical implications that follow.

<h2>5) MIT license</h2>

VspteModuleWrapper uses the MIT license. By my interpretation, the PR as it stands doesn't meet the requirements of this license as it doesn't provide a mechanism to distribute the license along with the binaries.

The MIT license is more restrictive than the FASTBuild license, which makes it somewhat incompatible.

**Proposed Solution:** Avoid the use of this header interface, perhaps by re-implementing the library loading via an interface that FASTBuild controls (like the optional Cache plugin). i.e. make the external tool buildable as a dll against the interface which FASTBuild provides rather than the other way around.  Optionally (as with the cache plugin) we could provide a dummy implementation in the tests, though we could certainly leave this for later.

<h2>6) VspteModuleWrapper doesn't appear to be thread-safe</h2>

The library can be loaded multiple times, simultaneously on multiple threads. Looking at the code, it seems to be safe, but nothing is enforcing that. The calling code in FASTBuild is therfore fragile - should the implementation of the external DLL change, or someone implements their own, this could break. Additionally, the library can be loaded mutliple times, but will only ever be freed once. The lifetime of the wrapper is also asymetrical, which looks to cause no problems at the moment, but could cause issues with resource leak checks in the future should the DLL behavior change.

**Proposed Solution:** Explicitily manage the new/delete of the wrapper, protected by a mutex and assume construction and destruction must be serialized and symmetrical.

<h2>7) VspteModuleWrapper uses exceptions</h2>

FASTBuild is built without exception support.

**Proposed Solution:** Avoid the use of the external header by taking the cache interface approach.  If we want to protect against crashes during plugin initialization, use a SEH wrapper instead of C++ exceptions. 

<h2>8) Other</h2>

<h3>NODE_RESLULT_OK_CACHE is used incorrectly</h3>

This return value is for objects retrieved from the network cache, which doesn't apply here. This node should either build, or be already up-to-date. If we resolve the "DoBuild" point above, this will be taken care of (by this code vanishing)

<h3>Project parsing code is unsafe</h3>

There are some places in the code that assume certain strings will be found in the project file and will derference nullptr if provided a malformed project. Malformed inputs or misconfigured external files should not cause FASTBuild to crash.

**Proposed Solution:** Robustify the parsing code.

<h3>Parsing code has some minor inefficiencies</h3>

The parsing code (when extracting the guid) copies the entire file several times into temporary strings when obtaining the delimiters for the guid.

**Proposed Solution:** Use the found delimiters to directly construct the GUID instead of copying/truncating.

<h2>9) Coding standard/style</h2>

There are several coding standard/style differences in the proposed code (formatting/whitespace, brace style, variable naming etc). There are a few whitespace changes to unrelated code.

**Proposed Solution:** I can fix these. I mention them only for completeness. If any of these are improved/fixed while addressing any of the above it would be appreciated though.

lucianm:
Hello @ffulin ,

> ## 1) DoBuild implementation
> In the DoBuild function of VSProjectExternalNode, a hash of the file contents is calculated/stored into m_FileChecksum and the m_Stamp is always left at zero. FASTBuild's dependency graph expects (for externally provided items which are files) that the m_Stamp contains a representation of the file state, which is usually the modification time. This is used to determine if a build is out-of-date or not and can have other special meaning. Except for rare cases, DoBuild should not execute when there is no work to do. Currently only things that generate output should run every build, and even that will change over time.
> 
> The implementation as it stands requires the external projects be loaded/hashed every time and that DoBuild be called every time. It duplicates a subset of the "do i need to be built" logic that is provided by the DependencyGraph. In addition to unnecessary rebuilding, -verbose output is inconsistent and the build statistics are potentially incorrect.
> 
> **Proposed solution:** Remove the file hashing logic, assume DoBuild is called only when the file has changed and set m_Stamp at the end of DoBuild by calling the RecordStampFromBuiltFile() function. Optional: Extend the tests to validate that the nodes are not built when nothing has changed.

This was actually a point I was unhappy about myself, as it seems I did not fully understood the "do i need to be built" logic how you called is relying on m_Stamp, but then again, I thought that deciding that an external file has to be parsed/investigated again just because its file attributes have changed might not always be what we actually want to happen. Often checking out from version control again might renew file time, but still give us the exact same contents. But well, that also happens with any source file, so why not, I think you're right, RecordStampFromBuiltFile() should be enough.

> ## 2) VSProjectExternalNode exposes many unused properties
> 
> VSProjectExternalNode derives from VSProjectBaseNode, which in turn exposes all the properties that can be used to configure a generated project. These properties can therefore be set on VSProjectExternalNode, but they don't do anything. This could be confusing (to users and to someone working on the cdoe). It's also inefficient (many unnecessary properties are serialized and modifying them will cause unnecessary rebuilds).
> 
> I'm sorry about this - this confusion is probably my fault. I didn't consider that you might interpret my suggestion of "adding a base class" as referring to this. This class is unrelated to the base interface I think we need to add that VCProjectNode and your new node derive from. For simplicity, let's not rename the existing structure even though we might want to do that later to disambiguate things.
> 
> **Proposed Solution:** Revert the changes/move of this helper structure and create a new base class which implements only the virtual functions that Solution() needs (like GetProjectGUIDType())

Well, I wasn't happy having to come up with moving the properties to the base class, but there was a problem to overcome: Having to deal now with the abstract base class of the 2 siblings (the original VCXProjectNode and the VSProjectExternalNode) made it difficult to get the proper `Array< VSProjectConfig>` and more so access to its elements in SLNNode.cpp:228 `project->GetConfigs()` (as project here is just the base class, both implementations of GetConfigs need to provide access to m_Platform and m_Config), but on the other hand in `VSProjectGenerator::GenerateVCXProj` where the input argument is actually the same `Array< VSProjectConfig>` thes absolutely need to hold the complete set of the many properties needed in VCXProjectNode. My first attempt with some ugly casting of each element failed. I hope you're getting my point, I'm not saying reverting to the original structure as you suggest won't work, but how to do it in a clean way? Maybe the array should be of pointers to a basic VSProjectConfig and each node should fill it with instances of its own derived config, I'll look into that, it will allow to revert to the original structure, but I'd really use proper, suggestive names of these config types.

> ## 3) Tests
> 
> As you noted, the tests are failing. We'll need to find a way to have the tests pass out-of-the box for end users and on the CI without a dependency on external/optional components.
> 
> The test also doesn't validate that the project nodes in the test were actually built.
> 
> **Proposed solution:** Modify the test to validate the built-in "auto" behavior and remove the need for the external dependencies. Add a "node count" check for the 4 nodes being referenced.

Actually my intention was to checkin a version of the test BFF in which the project type GUID is provided and therefore no need for the external dependency in order to pass the test, with the visible comment in the BFF to add also the commented out projects (without type GUIDs) only if the external DLLs are available. My mistake was that I deleted the explicit type GUIDs for Proj1 and Proj2, so this is easily fixed and the test should pass out of the box. What do you mean by validating the project nodes?

> ## 4) Dependency on external VspteModuleWrapper
> 
> I've gone to great lengths to keep FASTBuild free of 3rd party code, and with the exception of 2 things which are entirely embedded inside FASTBuild, there are no external dependencies (ignoring platform SDKs/compilers etc of course).
> 
> **Proposed Solution:** This is a philosophical concern, but there are some practical implications that follow.

Well, my first approach was to just let the user figure out somehow what project type GUID to fill in, then you expressed the wish it should be done easy for the user :-). I couldn't find any reasonable approach other than with .NET (which isn't my area actually), therefore I assumed you would not want it inside FASTBuild. Do you? I'd have no problem at all, I do not insist on maintaining such a specialized module which is useless for anything else than FASTBuild, then all the practical implications you're mentioning, plus the license issue (I just didn't know better at that moment) would be easily mitigated ;-)

> ## 5) MIT license
> 
> VspteModuleWrapper uses the MIT license. By my interpretation, the PR as it stands doesn't meet the requirements of this license as it doesn't provide a mechanism to distribute the license along with the binaries.
> 
> The MIT license is more restrictive than the FASTBuild license, which makes it somewhat incompatible.
> 
> **Proposed Solution:** Avoid the use of this header interface, perhaps by re-implementing the library loading via an interface that FASTBuild controls (like the optional Cache plugin). i.e. make the external tool buildable as a dll against the interface which FASTBuild provides rather than the other way around. Optionally (as with the cache plugin) we could provide a dummy implementation in the tests, though we could certainly leave this for later.

I need tofind the code and look into that approach used in the optional cache plugin you're talking about.

> ## 6) VspteModuleWrapper doesn't appear to be thread-safe
> 
> The library can be loaded multiple times, simultaneously on multiple threads. Looking at the code, it seems to be safe, but nothing is enforcing that. The calling code in FASTBuild is therfore fragile - should the implementation of the external DLL change, or someone implements their own, this could break. Additionally, the library can be loaded mutliple times, but will only ever be freed once. The lifetime of the wrapper is also asymetrical, which looks to cause no problems at the moment, but could cause issues with resource leak checks in the future should the DLL behavior change.
> 
> **Proposed Solution:** Explicitily manage the new/delete of the wrapper, protected by a mutex and assume construction and destruction must be serialized and symmetrical.

When used through the singleton VspteModuleWrapper provided in the header, the library is only loaded once in the constructor, regardless how many headers attempt this. I even have a GTest with that, of course, maybe I should spawn even more than just 3 threads there. The VspteModuleWrapper destructor frees the library. The behaviour of the library should not really change, maybe you still think about adopting it, or another possibility would be making use of the API version.

> ## 7) VspteModuleWrapper uses exceptions
> 
> FASTBuild is built without exception support.
> 
> **Proposed Solution:** Avoid the use of the external header by taking the cache interface approach. If we want to protect against crashes during plugin initialization, use a SEH wrapper instead of C++ exceptions.

Will look into that, let's see where to start.

> ## 8) Other
> ### NODE_RESLULT_OK_CACHE is used incorrectly
> 
> This return value is for objects retrieved from the network cache, which doesn't apply here. This node should either build, or be already up-to-date. If we resolve the "DoBuild" point above, this will be taken care of (by this code vanishing)

Alright, seems I did not correctly understood what that value was for, I'll change that as well.

> ### Project parsing code is unsafe
> 
> There are some places in the code that assume certain strings will be found in the project file and will derference nullptr if provided a malformed project. Malformed inputs or misconfigured external files should not cause FASTBuild to crash.
> 
> **Proposed Solution:** Robustify the parsing code.
> ### Parsing code has some minor inefficiencies
> 
> The parsing code (when extracting the guid) copies the entire file several times into temporary strings when obtaining the delimiters for the guid.
> 
> **Proposed Solution:** Use the found delimiters to directly construct the GUID instead of copying/truncating.

Agreed.

> ## 9) Coding standard/style
> 
> There are several coding standard/style differences in the proposed code (formatting/whitespace, brace style, variable naming etc). There are a few whitespace changes to unrelated code.
> 
> **Proposed Solution:** I can fix these. I mention them only for completeness. If any of these are improved/fixed while addressing any of the above it would be appreciated though.

Hehe, it was almost impossible not to have these, as I incrementally worked on all of this. Now since a whole lot of refactoring is needed again, I will try to sort out as much as possible of these coding style issues.
My biggest concern right now is with 2.)

lucianm:
Hello @ffulin,

so I refactored yet again my PR and was able to solve few of the positions enumerated by you, which I will summarize below. This time I consolidated everything in one commit for this iteration, as I'll progress with the remainders I'll add new commits based on this one:

1. `VSProjectExternalNode::DoBuild` was pretty straightforward to adapt according to your proposal
2. This one gave me headaches because I started implementing in the wrong direction, when the pretty slick solution was to leave `VCXProject` almost in its original, working state and introduce an array of platform / configuration tuples which is only needed in `SLNNode::Initialize` and can be consistently copied from the actual `VSProjectConfig` array in `VCXProjectNode::Initialize`, I'm assuming these are always called only once and due to dependency enforcement in the right order (projects before solution).
3. The tests issue on my previous push was not caused by the dependency to the external module (I had that on my mind and commented out the 2 projects which would require it, in the test BFF file), but by the node statistics check at the end of my new test. Now I got it right by peeking a bit at the actual numbers, so the test should not fail. More than this, at the present time, Travis CI checks are failing because of other tests, even on my laptop, without this PR, the following test groups are failing (I commented them out locally), and adding my PR does not make the others and my new test fail:
     - TestBuildFBuild
     - TestDistributed
     - TestObjectList
     - TestPrecompiledHeaders
     - TestUnity
     - TestWarnings
8. Regarding NODE_RESULT_OK_CACHE usage, I removed it, as I wasn't aware of its actual purpose until you explained.

The rest is still open, so now I'll work on making the parsing more robust and looking at what you suggested about the caching plugin interface.

Code/Tools/FBuild/FBuildCore/Graph/VSProjectExternalNode.cpp
```diff
@@ -0,0 +1,152 @@
...
+        extProjFileAsString.SetLength(uint32_t(fileSize));
+        if (fs.ReadBuffer(extProjFileAsString.Get(), fileSize) != fileSize)
+        {
+            FLOG_ERROR("VSProjectExternalNode - Failed to read external project file '%s'", m_Name.Get());
+            return NODE_RESULT_FAILED;
+        }
+
+        // parse Project GUID from string buffer
+        AString strPGStart(extProjFileAsString.FindI("<ProjectGuid>"));
+        AString strPGEnd(extProjFileAsString.FindI("</ProjectGuid>"));
```
ffulin: This will crash in ProjectGuid is not found

Code/Tools/FBuild/FBuildCore/Graph/VSProjectBaseNode.cpp
```diff
@@ -0,0 +1,53 @@
...
+    m_Type = Node::PROXY_NODE;
+    m_LastBuildTimeMs = 100; // higher default than a file node
+}
+
+
+// Initialize
+//------------------------------------------------------------------------------
+/*virtual*/ bool VSProjectBaseNode::Initialize(NodeGraph& /*nodeGraph*/, const BFFIterator& /*iter*/, const Function* /*function*/)
+{
+    // Generate default platform / config tuples if not specified
```
ffulin: The external project will have these configurations by default, which have no relation to what's in the project file, so these defaults being in this base class don't seem to make sense.

I imagine the ideal outcome would be for the External node to parse and extract the projects present in the file?

ffulin:
Thanks for the fresh PR revision. I've poked the tests to ensure there are no issues with those and can confirm they are nothing you need to worry about (sadly there are some intermittent test failures on Travis CI that I cannot reproduce locally yet).

The latest revision overall looks pretty good, and I think we're really close to resolving this. There are a few minor issues that I can resolve while integrating, so I won't bother calling out anything but the non-trivial remaining issues.

The only things I think we need to resolve are:

**1) The new node exposes a list of configurations**

Having a default set of project configurations makese sense for the VCXProject generator that already exists, but for the new VSExternalProject it doesn't really make sense as it will almost never match what is actually in the project.

I think we have a few options:
a) Make the new node parse the existing project file and extract the configs. Pro: best user experience. Cons: Would need implementing. Could be complicated for some reason.
b) Require the user to specify the configurations always. Pro: Simple Con: Annoying for the user
c) Not require the configurations, and generate the sln file with the assumption that the needed configs exist. Pro: Simple Con: Could hide configuration errors and make them really hard to diagnose.

Do you have any thoughts on this?

**2) The dependency on an external tool**

We're still left with the issues around a) licensing and b) dependency on an external tool.

I think we should simply implement within FASTBuild a list of file extension to ProjectTypeGuid mappings. This would completely remove the need for this external tool and all the associated issues. It's pretty easy to find a list of common project types and we can easily add any types you need (like the wix project). Although this isn't entirely future proof in an automated way as your current solution would be, we can trivially add new types the the mapping in the future and having the ability to set it explicitly in the bff means nobody would ever be blocked waiting for support in a new version.

Likewise, any thoughts on this?

lucianm:
> **1) The new node exposes a list of configurations**
> 
> Having a default set of project configurations makese sense for the VCXProject generator that already exists, but for the new VSExternalProject it doesn't really make sense as it will almost never match what is actually in the project.
> 
> I think we have a few options:
> a) Make the new node parse the existing project file and extract the configs. Pro: best user experience. Cons: Would need implementing. Could be complicated for some reason.
> b) Require the user to specify the configurations always. Pro: Simple Con: Annoying for the user
> c) Not require the configurations, and generate the sln file with the assumption that the needed configs exist. Pro: Simple Con: Could hide configuration errors and make them really hard to diagnose.
> 
> Do you have any thoughts on this?

Yeah, this matter felt strange to me as well, I guess I just neglected it and then didn't give it anymore thoughts due to inheriting it from my initial implementation doing it all in `VCXProject`.
Going with **a)** should not be that complicated, provided I'll be using the official API exposed by MS which I used also for extracting the `ProjectTypeGuid`. I had a look in the documentation, seems to be all there, the instantiated Project object exposes the so-called ConfigurationManager, that ought to be the source of the actual information.
Alternatively, if the user chooses not to use the external tool (which I'm reminding, _is optional_ the way I implemented this PR so far), they should be provided in the BFF, just like the `ProjectTypeGuid`. This would also hopefully allow for helpful diagnostics.


> **2) The dependency on an external tool**
> 
> We're still left with the issues around a) licensing and b) dependency on an external tool.
> 
> I think we should simply implement within FASTBuild a list of file extension to ProjectTypeGuid mappings. This would completely remove the need for this external tool and all the associated issues. It's pretty easy to find a list of common project types and we can easily add any types you need (like the wix project). Although this isn't entirely future proof in an automated way as your current solution would be, we can trivially add new types the the mapping in the future and having the ability to set it explicitly in the bff means nobody would ever be blocked waiting for support in a new version.
> 
> Likewise, any thoughts on this?

I don't really like the idea of a hard-coded list mapping files-extensions to known project type Guids, otherwise I would have done it this way from the beginning. First of all, they are not unequivocal (there are multiple guid's sharing the same file extension out there), and then I started to investigate how to retrieve them the proper way, and to me, the proper way seems to be using an official MS API. Well, unfortunately such API is not available for plain unmanaged code, only for .NET, that's why I thought of dynamically loading a DLL if necessary and if available, the diagnostic message on why loading it may have failed could be improved.

I have absolutely no problem of changing anything necessary with the licensing (please make suggestions, as I just picked one which I thought would be ok, not having too much experience with that sort of matter).

Last but not least, I'm not seeing the whole dependency matter as being too much of a problem, since it's not a _required_ dependency, using the external tool is still _completely optional_, the user has the possibility to provide the project type guid (and also the configurations if I'll implemented as proposed in my answer to **1)** ) himself. It's only getting tedious if those external projects change configurations a lot, otherwise they usually are added to a solution only the first time. Even finding out the Project Guid by the user is easily done, (s)he only has to manually load such a project in any solution, save it and look it up in the saved solution file, also only necessary once when adding that project. But having a tool which is able to properly retrieve this information and let FASTBuild update it because the timestamp has changed comes in very handy in my opinion. If you like, I can even imagine to make the whole external tool really part of FASTBuild, I think FASTBuild can already build C# and managed C++? It would only be a matter of building it only on supported platforms, but that shouldn't be a problem.

ffulin:
**List of configurations in project**

> if the user chooses not to use the external tool ... they should be provided in the BFF

I don't think there's any benefit in making the user specify the project configs since FASTBuild can't use this information to detect any mismatches against the actual file on disk (which is the thing you'd presumably care about).

Unless we can detect the configs in the project, this just seems like an inconvenience to me. (And if we can, by parsing the file or any other means, then we don't need the list to be manually specified)

**Dependency on external tool**

> there are multiple guid's sharing the same file extension out there

Do you have an example of this I can take a look at? I'd like to understand whether:
a) this is a case we need to worry about
b) if there are other properties inside the object that are important or can be used to differentiate those cases
c) whether either GUID could be used in that case

I definitely understand the desire to solve this problem 100%, but it's not just this initial implementation I'm thinking about. A 95% solution might be better if it:
a) is simpler
b) is easier to test and maintain
c) works on all platforms
d) provides more functionality out of the box
e) doesn't require collaboration with an external depot

This is of course provided it can solve the actual problem.

I'd like this functionality to work out of the box, so I think even if we keep the external tool, I'd make the default behavior auto-detect the GUIDs from a built in list anyway. Of course I appreciate I might be missing some aspect of this that makes that completely unmaintainable that I didn't consider!

lucianm:
> Do you have an example of this I can take a look at? I'd like to understand whether:

I found several examples on the net:
- [here](https://social.msdn.microsoft.com/Forums/vstudio/en-US/b0ce091e-ce19-43ae-8a95-fe9573847717/projecttypeguids-and-subprojects?forum=vsx)
- [here](https://weblogs.asp.net/leftslipper/opening-an-asp-net-mvc-project-without-having-asp-net-mvc-installed-the-project-type-is-not-supported-by-this-installation)
- [or here](https://dzimchuk.net/projecttypeguids-node-in-vs-project-files/)

just quickly found again now after such long time I didn't bother anymore on how to solve the issue since the solution became obvious for me, to effectively use the **installed instance of Visual Studio** (because that's what this project type Guids matter boils down to, seems like on multiple guids, VS will pick up the proper one based on what project types it is aware of on that machine installation). The external tool I'm proposing  (of which you seem to get increasingly reluctant about ) is doing exactly that: instrumenting the installed VS to do the job right.

> a) this is a case we need to worry about
> b) if there are other properties inside the object that are important or can be used to differentiate those cases
> c) whether either GUID could be used in that case

Only the "proper" one. The installed Visual Studio "knows" this best.

> I definitely understand the desire to solve this problem 100%, but it's not just this initial implementation I'm thinking about. A 95% solution might be better if it:
> a) is simpler

The simplest solution is to let the user fill in the Guid, either found in lists on the internet, or more reliable and useful, inspected by himself from a temporary solution in which he loads that project **with his own installation** of Visual Studio. Now after the issue you mentioned with the configurations, I don't know anymore if this is getting "simple", anyway, I don't think that parsing configurations information from the XML code and getting it 100% right will be trivial, the way I've seen it is spread and repeated several times in such project files.

> b) is easier to test and maintain
> c) works on all platforms
> d) provides more functionality out of the box
> e) doesn't require collaboration with an external depot

This functionality in itself is of no benefit of anyone NOT using Visual Studio's solutions / projects. Building code with FASTBuild does not require this, only bootstrapping via MSBuild through these generated projects.  Which means, Visual Studio is required anyway. And afaik, other than Windows, there is only MacOS for which a sort of Visual Studio exists, but I do no know if the format of solutions/projects is in any way compatible, this is something that should be investigated, but on all other supported platforms, generating projects/solutions should not be of any relevance.
The external dll has quite simple business-logic code, as it merely starts an instance of VS, creates a temporary solution and loads external projects one after the other and queries them for information, that's it. Testing this of course requires an installed VS supporting the project Guids used in the test code, but that are also the requirements for real life usage of the tool. To still keep this functionality usable even if the external tool (which by this occasion I'm re-iterating, doesn't have to stay external, the whole code can be part of FASTBuild, just like the distributed worker which is not used by everyone, lots of  users are happy to use just FASTBuild itself) should misbehave due to some regression, the information needed can still be provided explicitly.
 
> This is of course provided it can solve the actual problem.
> 
> I'd like this functionality to work out of the box, so I think even if we keep the external tool, I'd make the default behavior auto-detect the GUIDs from a built in list anyway. Of course I appreciate I might be missing some aspect of this that makes that completely unmaintainable that I didn't consider!

For me, being able to manually provide the Guid is enough and the only thing I would trust beside VS itself (instrumented in the external tool).

ffulin:
Thanks again for the extra info. I'd like to get this merged so that we can avoid any conflicts getting introduced,so I took another look at the review to refresh my memory.

There are a few minor outstanding things (comments/questions on the review), but in the interests of getting this integrated before more conflicts arise, I think I can address those as part of integration. 

The only blocker to integration that remains then is the use of your interface header due to licensing issues (VSProjTypeExtractor.h)

I think we can either: 
a) remove that header from the PR and directly call the functions by making a minimal wrapper inside FASTBuild
OR
b) change the licensing terms of your library (or actually even just the header if you prefer). Something likethezlib license is compatible

Option a) would allow more flexibility on the FASTBuild side and would allow us to make things like the singleton creation/destruction well scoped (i.e. create/destroy around a single test instead of the first test creating it and it carrying through to application shutdown) It also avoids any licensing issues.

The second option would require fewer code changes, at least in the short term.

Do you have a preference? I think I'm leaning towards option a) but I could be swayed either way.

lucianm:
Hi, I'm ok with option a) as well, what I did with the header was just because I thought due to the nature of this contribution you would prefer something quite minimal invasive to the FASTBuild code. I can nevertheless as well change the licensing to whatever you prefer even if we go for option a) :-)

ffulin:
Just to clarify - what I meant by option a) was to keep your external DLL, but instead of using the interface header, we make a new interface header just to define the dll functions inside the FASTBuild source. This keeps all the logic in your external project/dll while avoiding the licensing issue as well as making the way the dll is loaded more easily controlled on the FASTBuild side (for tests etc).

lucianm:
Alright, got your point better now. How about this:

1. I'll remove the part inside #ifdef VSPROJTYPEEXTRACTOR_DYNLOAD from the library-own VSProjTypeExtractor.h and merely document it in the developer information part of the Readme addressing the "Dynamic (less invasive) integration", use the same in my GTest for example.
2. As a first version which I think you would merge, I'd use the removed part in the new interface header you're refering to, which will be part of FASTBuild, I'll give it a slightly different name (how about VSProjLoaderInterface.h ?), the licensing issue will be avoided and the code can still be adapted, improved if you think, as necessary, even later.
3. The list of project configurations issue you mentioned in an earlier reply can be addressed as well in the external code and exposed through the interface and taken care of in the PR.

ffulin:
The first two items sound fine - they key thing being that if we just make a header inside FASTBuild itself, I'm sure we can adapt it to meet our needs.

For the third point (which I forgot about :) ) I think we just wand to leave the list of configs as empty by default and require the user to populate it. For a later update it might be nice to try and discover those automatically somehow, but in the interests of moving forward I think we should leave that for later.

This means in the bff, the user will (for now at least) have to specify the configs they know are present in the externally generated project file (only the ones they want the generated Solution to reference). For me that seems pretty reasonable.

lucianm:
As a matter of fact, in the meantime I already added the functionality of retrieving config/platform pairs as well, to the [external DLL](https://github.com/lucianm/VSProjTypeExtractor), only had to finish / test it, now it's published.
The only thing remaining would be to adapt the PR again.

You can already have a look at how the [header added to FASTBuild would look like](https://github.com/lucianm/VSProjTypeExtractor/blob/master/VSProjTypeExtractorTest/VSProjLoaderInterface.h), as for now also as documented in the README.

lucianm:
So, I refactored the PR again, now the external module is able to retrieve existing configurations as well. Providing all information in the BFF will prevent VSProjectExternal to spawn de external module for that respective project:
```
VSProjectExternal( alias )               // (optional) Alias
{
  // Input Options
  .ExternalProjectPath            // (mandatory) Path to project file
  .ProjectGuid                    // (only when avoiding external module) Project Guid
  .ProjectTypeGuid                // (only when avoiding external module) Project Type Guid
  
  // Build Config Options
  .ProjectConfigs                 // (only when avoiding external module) List of project configurations (see below)
                                  // (default: Debug Win32, Release Win32, Debug X64, Release X64)
}
```

If any of it is missing, the external module will try to query all information from the project, but only what is not specified in the BFF will be taken from the module. So a minimal VSProjectExternal declaration will look just like:
```
VSProjectExternal( alias )               // (optional) Alias
{
   .ExternalProjectPath            // (mandatory) Path to project file
}
```
This way, the external module is not at all mandatory, but just convenient if available.

ffulin:
I've integrated this change. It will be in the next release (v1.00). Thanks :)

I've made a few followup changes that you might like to review (listed below).

https://github.com/fastbuild/fastbuild/commit/5f8f53a9c18632fe3c00a22da282b3d583da8d55
 - code style/convention
 - Add missing fdb version bump for new node

https://github.com/fastbuild/fastbuild/commit/34afd78aa28834727212a0f88bc62016e4392b5d
 - Regenerate integrations files using generation script

https://github.com/fastbuild/fastbuild/commit/3db9e596d42f09967e364fe49e79d410df373445
 - Fix .fdb files being written in the test source directory
 - Fix crash when trying to extract a ProjectGuid out of a malformed project file (and add test)
 - Avoid unnecessary FileExists() check in DoBuild
 - minor whitespace fixes

There are a few unresolved issues that I still plan to address:

**1. Loading the plugin dll is not threadsafe.** Multiple threads can call the loading function in a manner which is not thread-safe. (Function pointers in the dll can be used before being initialized for example)

**2. External project default configs*
External projects get a matrix of Debug/Release * x86/x86 configs by default. These don't necessarily (or usually) match the actual configs in the project at all, and this causes us to silently generate a project file that will generate errors on solution load.

When the external plugin is not being used, we need to default this list to empty and require the user to set the configs.

lucianm:
Hi,

I've been looking over the follow-up changes. I'm ok with them. Did not increase the node version myself because I could not know if other fdb changing commits would be merged and in which order, but you're right, maybe I should have changed it anyway just to make sure it's not forgotten.

About the remaining issues:

1. The thread safety of loading the plugin (1) is easily obtained by making use of the singleton, I'll write a new PR.
2. I had my difficulties with the platform/config pairs matrix for the external project node, because if not supplying them at all, before the external module gets to extract them when `VSProjectExternalNode::DoBuild` is executed (possibly concurrently if necessary), FASTBuild already complains because they're missing. The other possibility would have been (and I also had it running for a while like that) to use the external module already in `VSProjectExternalNode::Initialize`, but then there were problems with reflecting/persisting the values. Maybe I did wrong somewhere, but I think, especially now after hinting you on this you'll come up with the proper fix ;-).
If I did not completely misunderstood how/when Initialize / DoBuild are being called, the way it is implemented now, the private method `VSProjectExternalNode::CopyConfigs` first initializes the array passed to the solution node later with the values provided in BFF in Initialize, and if that happened, even if the external module is used, they are no longer overwritten. If they are not provided in the BFF, they are retrieved from the external module, and defaults are added only if missing, maybe this last part is debatable.

PR #296, opened 2017-09-30
Fix readlink usage in GetExePath

dummyunit:
`readlink` doesn't add terminating null character to the supplied buffer [1]. Therefore we shouldn't treat `path` as a null-terminated string in `GetExePath`.

[1] https://linux.die.net/man/3/readlink Section "Application Usage".

### DIFF
--- a/Code/Core/Env/Env.cpp
+++ b/Code/Core/Env/Env.cpp
@@ -221,8 +221,10 @@ void Env::GetExePath( AString & output )
         output = argv[0];
     #else
         char path[ PATH_MAX ];
-        VERIFY( readlink( "/proc/self/exe", path, PATH_MAX ) != -1 );
-        output = path;
+        const ssize_t length = readlink( "/proc/self/exe", path, PATH_MAX );
+        VERIFY( length != -1 );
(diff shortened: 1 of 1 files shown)


Code/Core/Env/Env.cpp
```diff
@@ -221,8 +221,10 @@ void Env::GetExePath( AString & output )
         output = argv[0];
     #else
         char path[ PATH_MAX ];
-        VERIFY( readlink( "/proc/self/exe", path, PATH_MAX ) != -1 );
-        output = path;
+        const ssize_t length = readlink( "/proc/self/exe", path, PATH_MAX );
+        VERIFY( length != -1 );
+        VERIFY( length <= PATH_MAX );
```
ffulin: These VERIFY statements should be ASSERTs. I'll fix that post-merge.
ASSERT = stripped in Release builds
VERIFY = code remains, but assertion of result is stripped in Release builds.

ffulin:
I've integrated this into the Perforce mainline. The fix will be in the next release (v0.95). Thanks for the fix!

PR #297, opened 2017-09-30
Limit debug check for ':' in FileNode to Windows

dummyunit:
Second `ASSERT` in the `FileNode` ctor verifies that `:` can only appear as the second character in the `fileName`. This check should only apply to Windows because Linux and Mac OS allow to use `:` anywhere in file path.

This fixes TestSerialization test in Linux debug build.


ffulin:
This fix look legit. I never noticed it because I don't have any ':' in my paths on Linux or OSX :)

ffulin:
I've integrated this into the Perforce mainline. The fix will be in the next release (v0.95). Thanks!

PR #298, opened 2017-10-01
Extend sleep workaround in BFFDrifted test to Linux

dummyunit:
Some file systems used on Linux (ext2, ext3, reiserfs) have only a second resolution for timestamps. And even file systems with a higher timestamp resolution can in practice be limited to a millisecond resolution because of time caching in the kernel (see `current_fs_time()` in the kernel and this answer from stackoverflow: https://stackoverflow.com/questions/14392975/timestamp-accuracy-on-ext4-sub-millsecond#answer-14393315).
Thus to avoid spurious failures we should sleep for 1s before clearing fbuild.bff file in the BFFDrifted test.

I have decided to duplicate call to `Sleep` in both `#if` blocks to better separate reasons for doing so on each platform.

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestGraph.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestGraph.cpp
@@ -595,6 +595,8 @@ void TestGraph::BFFDirtied() const
 
     #if defined( __OSX__ )
         Thread::Sleep( 1000 ); // Work around low time resolution of HFS+
+    #elif defined( __LINUX__ )
+        Thread::Sleep( 1000 ); // Work around low time resolution of ext2/ext3/reiserfs and time caching used by used by others
     #endif
 
     // Modity BFF (make it empty)


ffulin:
Thanks for the fix. I've integrated this into the Perforce mainline. This will be in the next source code release (v0.95).  I went ahead and checked the tests for other similar issues, and added the equivalent work around for a few other places (https://github.com/fastbuild/fastbuild/commit/46f0b61f877cda8c836973f405054ea3b19b2dc4).

dummyunit:
Looks like commit 46f0b61 broke non-unity build on Linux:
```
/home/dummyunit/fastbuild/Code/Tools/FBuild/FBuildTest/Tests/TestObject.cpp: In member function 'void TestObject::TestStaleDynamicDeps() const':
/home/dummyunit/fastbuild/Code/Tools/FBuild/FBuildTest/Tests/TestObject.cpp:193:9: error: 'Thread' has not been declared
         Thread::Sleep( 1000 ); // Work around low time resolution of ext2/ext3/reiserfs and time caching used by used by others
         ^
compilation terminated due to -Wfatal-errors.
```

ffulin:
Fixed: https://github.com/fastbuild/fastbuild/commit/4444a930939e2ec8742fe8babe402252fe84d8c4 - Thanks!

PR #299, opened 2017-10-03
- Fixed compilation errors(and assertions in debug) with DeoptimizeWritableFiles for precompiled header.

jspelletier:
Hi Franta,

Trying to use merge requests to send you patches. Hopefully this will be much easier for you than our previous way of doing things.

The merge request is to fix build errors when making the pch input file writable when you have a .DeoptimizeWritableFiles setting.

The problem was that for the precomp.cpp, the m_CompilerOptionsDeoptimized was empty. This was then resulting in an invalid commandline passed to cl.exe. I don't know why but It seems that .PCHOptionsDeoptimized has been removed from fastbuild at some point(we are still generating it in sharpmake!).

Thus, this was resulting in the following command "cl /showincludes" which obviously is wrong.

I just ended up with a quick fix that disables deoptimization when compiling the precompiled header and this merge request is the result of this fix!

I've included below a truncated bff section extracted from our source tree.

I haven't added any new test case for this... Sorry about that.

Cheers
JS

ObjectList( 'Engine_FastBuild_Blob_Engine_Dll_Release_win64_objects' )
{
     Using( .win64Config )
    .Intermediate           = '..\..\tmp\win64\engine\engine_release_dll_vs2015_fastbuild\'

    .CompilerExtraOptions   = ''

    // Optimizations options
    // ---------------------
    .CompilerOptimizations = ''
            + ' /Ox'
            + ' /Ob2'
            + ' /Oi'
            + ' /Ot'
            + ' /Oy-'

    // Precompiled Headers options
    // ---------------------------
    .PCHInputFile           = '..\source\engine\precomp.cpp'
    .PCHOutputFile          = '..\..\tmp\win64\engine\engine_release_dll_vs2015_fastbuild\engine.pch'
    .PCHOptions             = '"%1" /Fp"%2" /Fo"%3" /c'
                            + ' /Yc"Engine\precomp.h"'
                            + ' $CompilerExtraOptions$'
                            + ' $CompilerOptimizations$'

    // Compiler options
    // ----------------
    .CompilerOptions        = '"%1" /Fo"%2" /c'
                            + ' /Yu"Engine\precomp.h" /Fp"$PCHOutputFile$"'
                            + ' $CompilerExtraOptions$'
                            + ' $CompilerOptimizations$'

    .CompilerInputUnity       = 'Engine_FastBuild_Blob_Engine_Dll_Release_win64_unity' 
    .CompilerOutputPath       = '$Intermediate$'
    .CompilerInputFiles       = { 'somefiles.cpp' }

    .PCHOptionsDeoptimized = .PCHOptions

    .CompilerOptionsDeoptimized = '"%1" /Fo"%2" /c'
                            + ' /Yu"Engine\precomp.h" /Fp"$PCHOutputFile$"'
                            + ' $CompilerExtraOptions$'
                            + ' /Od'

    .DeoptimizeWritableFiles = true
}

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1350,6 +1350,8 @@ bool ObjectNode::BuildArgs( const Job * job, Args & fullArgs, Pass pass, bool us
     PROFILE_FUNCTION
 
     Array< AString > tokens( 1024, true );
+    if ( GetFlag( FLAG_CREATING_PCH ) )
+        useDeoptimization = false; // Deoptimization not supported for precompiled header.
 
     const bool useDedicatedPreprocessor = ( ( pass == PASS_PREPROCESSOR_ONLY ) && GetDedicatedPreprocessor() );
     if ( useDedicatedPreprocessor )


ffulin:
I investigated this a bit, and while I think your change will fix the issue, the root bug actually lies in:
```C++
ObjectNode * ObjectListNode::CreateObjectNode()
```
That function is incorrectly passing the shouldUseOptimization flags to the precompiled header object when it should not. It seems this bug was introduced in v0.93.
```C++
    node->m_DeoptimizeWritableFiles = m_DeoptimizeWritableFiles;
    node->m_DeoptimizeWritableFilesWithToken = m_DeoptimizeWritableFilesWithToken;
```
when it should be doing something like:
```C++
    if (flags & FLAG_CREATING_PCH)
    {
        node->m_DeoptimizeWritableFiles = false;
        node->m_DeoptimizeWritableFilesWithToken = false;
    }
    else
    {
        node->m_DeoptimizeWritableFiles = m_DeoptimizeWritableFiles;
        node->m_DeoptimizeWritableFilesWithToken = m_DeoptimizeWritableFilesWithToken;
    }
```
I'll make that change. I'll leave this pull request open to track the issue.

jspelletier:
Hello Franta,

I have no problem with your fix. My fix was more like a patch to workaround the fact that we were unable to edit our precomp.cpp!

Thanks
JS

ffulin:
I've submitted the more "root" fix. I also added a test to catch this regression. (https://github.com/fastbuild/fastbuild/commit/07254dcf79492ce92ae0faf1d0f959923743219c) This fix will be in the next release (v0.95). (You should be able to continue using your fix locally with no problem of course)

Thanks for the bug report/pull request!  (PS - Going through Github is the way to go if that's ok for you)

Issue #286, opened 2017-09-11
Generated .sln should include a BOM for the Microsoft Visual Studio Version Selector to work


ChuangTseu:
The *Microsoft Visual Studio Version Selector* used to open .sln files with the correct Visual Studio version doesn't work if the .sln files is encoded as simple UTF-8 without the BOM.

See this SO answer for reference: https://stackoverflow.com/questions/2804759/vslauncher-starts-wrong-version/2829550#2829550

I have tested it myself just right now, with a fresh install of Visual Studio 2017, and having Visual Studio 2015 still installed on my PC: now double-clicking the VS2015 .sln files generated by FastBuild opens it with VS2017. Then, manually saving the .sln with encoding UTF-8 with BOM restores the expected and working behavior (assuming the default file opener for .sln is the VS Selector): it opens with VS2015, and as an added bonus the file has a VS14 icon in the Windows Explorer.

ffulin:
I just opened a sln file create by FASTBuild and saved it out from VS2017, and the behaviour I see is that the leading CR/LF FASTBuild writes is stripped, but no BOM is added. If I create a new sln file from VS2017, I see it created with a BOM.

I created the Solution generation for compatibility with an older version of VS, but I don't recall exactly which version :) (I think it was at least 2013, but it might have been 2012 or even 2010). The only concern I have is making sure that we don't break compatibility for exisiting users.  Best case we can just write the BOM if it doesn't break compatibility (with whatever versions of VS currently work). Worst case we can use the SolutionMinimumVisualStudioVersion property to control the behaviour (writing the BOM only if it's > some value).

ChuangTseu:
Almost forgot about that one, a few days ago I got hooked and tried many things to refine my understanding of how the Selector works, here are my observations: https://gitlab.com/ChuangTseu/WikiTseu/wikis/Visual-Studio-Version-Selector:-observations-on-the-solution-format .

Basically, creating a new C++ project/solution from the Visual Studio project creator with VS 2012 up to 2017 yields a .sln file with a leading CR/LF _AND_ the BOM.

But interestingly enough, removing the leading CR/LF _AND_ saving the file WITHOUT BOM makes it work too! (It might even be the default for older than 2010 Visual Studio versions, can't test though).

Conclusion: I guess doing the same thing Visual Studio 11/14/15 does is a safe bet (i.e. a leading newline _AND_ a BOM).

ffulin:
[...]

PR #300, opened 2017-10-05
Fix solution compatibility with Visual Studio 2010

slonopotamus:
Microsoft Visual Studio Selector from Visual Studio 2010 doesn't understand
how to open solution generated with fastbuild even when specifying

	.SolutionVisualStudioVersion = '10.0.40219.1'
	.SolutionMinimumVisualStudioVersion = '10.0.40219.1'

This is caused by blank line inserted at the beginning of .sln file.

Unfortunately, code comment just says that adding this empty line is deliberate,
without explaining what purpose it is supposed to serve, so I am not sure what can break.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
@@ -116,7 +116,6 @@ void SLNGenerator::WriteHeader( const AString & solutionVisualStudioVersion,
     AStackString<> shortVersion( shortVersionStart, shortVersionEnd );
 
     // header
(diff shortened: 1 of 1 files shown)


ffulin:
I also wish I added a better comment :)  As far as I can recall, I added that because it was present in the Visual Studio solution I had available when I originally implemented the solution generation.

The only concern I had with removing it was compatibility with VS2010 (as detailed here https://github.com/fastbuild/fastbuild/issues/286) which I am no longer able to test. Since you have conveniently fixed it for that version specifically, I think we should be ok to integrate this fix!

ffulin:
I've integrated this fix into the Perforce mainline. This will be in the next release (v0.95) Thanks for the pull request!

PR #302, opened 2017-10-12
Handle a DT_UNKNOWN value for the type field of the returned dirent

fuzzac:
The d_type field of dirent (returned by readdir()) is not valid for all filesystems. Applications need to handle a value of DT_UNKNOWN and instead do a stat() on the file to determine type.


ffulin:
Thanks for the fix. I've integrated this change into the Perforce mainline. This will be in the next release (v0.95).

Issue #303, opened 2017-10-12
UnityInputExcludedFiles doesn't handle partial paths containing ".."


belkiss:
Hello!
I'm using the technique mentioned here https://github.com/fastbuild/fastbuild/issues/169 to have path relative to the bff file, and be able to reuse the same bff from multiple master bff.

Unfortunately, this doesn't work with .UnityInputExcludedFiles, in my case, I have:
` .UnityInputExcludedFiles = '$CurrentBffPath$\..\..\extern\libs\render\debug.cpp'`

and in the master bff,
`.CurrentBffPath = "project\vs2015"`

This is resolved by FastBuild as:
`project\vs2015\..\..\extern\libs\render\debug.cpp`

Which can never match the end of a path.

I found that always calling CleanPath there https://github.com/fastbuild/fastbuild/blob/28eb2aba32b083abd3479f05ac6d2fd26cccc67f/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp#L477 fixes the issue, but is not really nice as it converts the path as a full path by prepending the working directory.

My proposed fix is to have a CleanPath variant that doesn't prepend the working dir, but instead will leave the path as clean as possible.

So in my case,
`project\vs2015\..\..\extern\libs\render\debug.cpp`
would become
`extern\libs\render\debug.cpp`

And the current code will proceed with the partial path.
But, if after the call it still begins with "..", we would keep the current behavior, which is call the regular CleanPath and prepend the working directory.

What do you think?

Thanks!
Lambert

ffulin:
I've integrated your fix for this into the Perforce mainline. This change will be in the next release (v0.96). I'll leave this issue open until that version is released. Thanks for the pull request!

ffulin:
A version containing a change for this issue has now been released (v0.96).

PR #305, opened 2017-10-15
Allow UnityInputExcludedFiles to handle partial paths (Issue #303)

belkiss:
Heya,
Those commits fix the issue described in #303.

Best,
Lambert


belkiss:
I also joined to this a cleanup of the path passed as LIBPATH to the linker, as it sometimes creates an issue if they are not simplified.

ffulin:
This change will be in the next release (v0.96). Thanks!

belkiss:
Thank you!!

PR #306, opened 2017-10-15
Add /Zo to MS compiler command line to enable enhanced optimized debugging.

belkiss:
Heya,
This helps debugging FastBuild in release (the flag /Zo used to be called /d2Zi+, and was renamed in vs2013 update4).

### DIFF
--- a/Code/fbuild.bff
+++ b/Code/fbuild.bff
@@ -118,7 +118,7 @@ Compiler( 'Compiler-x64-OSX' )
 
     // Optimizations
     .CompilerDebugOptimizations     = ' /MTd /Od /RTC1 /GS /Oy- /GR- /analyze'
-    .CompilerReleaseOptimizations   = ' /MT /Ox /Oy /Oi /GS- /GF /GL /Gy /Gw /GR- /analyze'
+    .CompilerReleaseOptimizations   = ' /MT /Ox /Oy /Oi /GS- /GF /GL /Gy /Gw /GR- /analyze /Zo'
     .LibrarianDebugOptimizations    = ''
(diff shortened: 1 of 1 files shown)


Code/fbuild.bff
```diff
@@ -117,8 +117,8 @@ Compiler( 'Compiler-x64-OSX' )
     .LinkerOptions          = '/NODEFAULTLIB /WX /NOLOGO /INCREMENTAL:NO /OUT:"%2" "%1" /DEBUG'
 
     // Optimizations
-    .CompilerDebugOptimizations     = ' /MTd /Od /RTC1 /GS /Oy- /GR- /analyze'
-    .CompilerReleaseOptimizations   = ' /MT /Ox /Oy /Oi /GS- /GF /GL /Gy /Gw /GR- /analyze'
+    .CompilerDebugOptimizations     = ' /MTd /Od /RTC1 /GS /Oy- /GR- /analyze /Zo'
```
ffulin: Seems like we don't want/need it for debug (when using /Od).

belkiss:
You're right, I just fixed it, it was unclear when it was undocumented :)
Also, I was incorrect, /Zo started with vs2013 *Update 3*.

Thanks!

ffulin:
I've integrated this into the Perforce mainline. Weirdly, I am using the 2013 Update 4 compiler, and I get a warning about /Zo being an unknown option.  I've left the existing /d2Zi+ option, so hopefully this will at least result in good debug info regardless of VS2013 version. Both options will be removed when updating to VS2017 anyway (on by default since 2015).

Thanks for the pull request. This will be in the next release (v0.95) (unless I update to 2017 :) )

Issue #304, opened 2017-10-15
TestDLLWithPCH doesn't pass on 0.94


belkiss:
Heya!

The TestDLLWithPCH doesn't pass with the latest release.
It seems the commit introducing the regression is 1cc889e06bee91feba0fcf1bcedf997a931a1f7f.

ffulin:
I'm not able to reproduce this failure. Is there anything special about your setup that might help identify why it fails for you?

belkiss:
Hi!
This seems indeed to be a bug in vs2015, which I use to build FastBuild.
Here is the output of two consecutive launch of cl.exe from the TestDLLWithPCH (I have manually removed /nologo to get compiler version). Note that the second invocation doesn't have the "including file" line, causing the failure:

```
U:\>"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\cl.exe" C:\RD\FastBuild\Code\Tools\FBuild\FBuildTest\Data\TestDLL\PrecompiledHeader.cpp /Fo"C:\RD\FastBuild\tmp\Test\DLL\PrecompiledHeader.pch.obj" /Yc"PrecompiledHeader.h" /W4 /WX /Z7 /Fp"C:\RD\FastBuild\tmp\Test\DLL\PrecompiledHeader.pch" /c -DDLL_A_EXPORT  /showIncludes
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

PrecompiledHeader.cpp
Note: including file: c:\rd\fastbuild\code\tools\fbuild\fbuildtest\data\testdll\PrecompiledHeader.h

U:\>"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\bin\amd64\cl.exe" C:\RD\FastBuild\Code\Tools\FBuild\FBuildTest\Data\TestDLL\PrecompiledHeader.cpp /Fo"C:\RD\FastBuild\tmp\Test\DLL\PrecompiledHeader.pch.obj" /Yc"PrecompiledHeader.h" /W4 /WX /Z7 /Fp"C:\RD\FastBuild\tmp\Test\DLL\PrecompiledHeader.pch" /c -DDLL_A_EXPORT  /showIncludes
Microsoft (R) C/C++ Optimizing Compiler Version 19.00.24215.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

[...]

PR #307, opened 2017-10-15
Fix for TestDLLWithPCH failure (Issue 304)

belkiss:
Heya,
This fixes the issue descibed at #304 .
The problem was caused by the /showIncludes. Apparently, building the .pch with /showIncludes only gives the full list of includes if the .pch doesn't exist...

The regression introduced upstream at 1cc889e was leaving the file on disk except on clean builds, so now we also delete the old file in the case where we create the pch with msvc.


ffulin:
I've submitted a fix (https://github.com/fastbuild/fastbuild/commit/60c09e3ea864e0eecc88fc52e4c5a9f1b896d9c3) as detailed in https://github.com/fastbuild/fastbuild/issues/304.

PR #308, opened 2017-10-15
Fix nasty and rare bug that could delete source files when using fastcancel

belkiss:
Heya,

This fixes a bug I've been chasing for a while. In case of cancellation of build with fastcancel, the node result is forced to NODE_RESULT_FAILED.
But, with that status, we delete the old stale files from disk.

Most of the time everything went fine, because the "input" nodes where up to date, and in that case we weren't entering that loop.

The fix is just to add a test for the type of file node, ie, we only allow deletion of output file nodes.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
+++ b/Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
@@ -627,7 +627,8 @@ void JobQueue::FinishedProcessingJob( Job * job, bool success, bool wasARemoteJo
 
     // make sure the output path exists for files
     // (but don't bother for input files)
-    if ( node->IsAFile() && ( node->GetType() != Node::FILE_NODE ) && ( node->GetType() != Node::COMPILER_NODE ) )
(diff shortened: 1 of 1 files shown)


ffulin:
I looked into the code and this makes sense. There is a very small window where a worker thread takes a job to start, but before it starts, the build is cancelled, if using -fastcancel the node is set to FAILED and we try to clean it up.  Additionally, if the source file was read-only the delete would fail which would make it a lot less likely (at least for Perforce users).

ffulin:
On a related note, I'm thinking about removing the on-failure deletion in the future entirely. At the moment, it's mostly a fail-safe for FASTBuild, but I don't think it should really be needed.

ffulin:
I've integrated this fix into the Perforce mainline. It'll be in the next release (v0.95). Thanks!

belkiss:
In most of the cases where I saw the bug happen, the files that were deleted were actually headers in the visual studio folder in program files!
Thanks for integrating the fix!

ffulin:
> the files that were deleted were actually headers in the visual studio folder

Ouch :(  Anyway, thankfully fixed now!

PR #311, opened 2017-10-22
Include optimization options into .CompilerOptionsC

dummyunit:
LZ4 uses `.CompilerOptionsC` instead of `.CompilerOptions` in X86Clang and x64Linux configurations. But optimization options were only added to `.CompilerOptions` and not to `.CompilerOptionsC`.
This resulted in LZ4 compiled without `-O3` in Release configuration, which on machine lead to 10-12 times lower throughput of xxHash in CompareHashTimes_Large test.

### DIFF
--- a/Code/fbuild.bff
+++ b/Code/fbuild.bff
@@ -328,6 +328,7 @@ Compiler( 'Compiler-x64-OSX' )
     .Config                 = 'Debug'
     .CompilerOptions        + ' -DDEBUG -DPROFILING_ENABLED'
                             + .CompilerDebugOptimizations
+    .CompilerOptionsC       + .CompilerDebugOptimizations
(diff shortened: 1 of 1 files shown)


ffulin:
Good find!  The use of CompilerOptionsC only in those configs is an error - for consistency it should be using that everywhere (which would have made this bug consistent at least :) )  I'll do a fixup of the options after integrating.  I also noticed that we're not setting the defines properly for those targets/configs either (I'll fix that too).

ffulin:
I've integrated your fix and I've done the additional cleanup (https://github.com/fastbuild/fastbuild/commit/1c32b249d737d117a652970285885730c580c246):
 - All platforms/configs use .CompilerOptionsC consistently
 - Missing defines are added for .CompilerOptionsC (-DRELEASE etc)

Thanks for the fix! This will be in the next release (v0.95)

PR #312, opened 2017-10-22
Deduplicate alias generation, provide useful per-configuration aliases on Linux and OSX

dummyunit:
All identical code for generating per-platform, per-configuration and `All` aliases is moved to separate file. Main reason is to simplify second commit and further PRs (building with Clang on Linux, building with sanitizers).

Second commit changes per-configuration aliases to use `.Targets` useful on each platform.

### DIFF
--- a/Code/Core/Core.bff
+++ b/Code/Core/Core.bff
@@ -120,23 +120,7 @@
 
     // Aliases
     //--------------------------------------------------------------------------
-    // Per-Config
-    Alias( '$ProjectName$-Debug' )      { .Targets = { '$ProjectName$-X86-Debug',   '$ProjectName$-X64-Debug', '$ProjectName$-X86Clang-Debug' } }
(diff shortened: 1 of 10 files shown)


ffulin:
Thanks for the cleanup! I've integrated this into the Perforce mainline. This will be in the next source code release (v0.95)

dummyunit:
Well, that is embarrassing...  
In commit message for e0a30747ab2c57b751ac6664983a49ba51025ea6 I wrote that it fixes a typo: `X64-Release` was included in `X86` alias instead of `X86-Release` in 2 places. And now I realized that I have managed to copy the wrong version of that alias into `gen_default_aliases.bff`. :(  
So line 21 in `gen_default_aliases.bff` is incorrect.

ffulin:
Fixed: https://github.com/fastbuild/fastbuild/commit/ebab52c01bc7402a5ffc240de339960c94e37008

PR #313, opened 2017-10-22
Use AString::Copy overload with length when assigning/appending AString to AString

dummyunit:
There are two reasons to prefer Copy that accepts length:
   1. That version is much faster that the other, especially on long strings.
   2. This avoids incomplete string copies when source string contains an embedded null-terminator. In that case data is copied only up to the null-terminator but `m_Length` is set to the same value as in       the source string. Such incomplete copies can later lead to uninitialized memory reads.

Issue described in (2) was found in practice with MemorySanitizer:
   * In `Cache:Publish` last slash in `cacheFileName` is replaced with `'\0'`.
   * `FileIO::EnsurePathExists` is called, it makes (incomplete) copy of its argument.
   * `AString::Replace` is called (via `PathUtils::FixupFolderPath`) on that copy, it checks all characters in range `[m_Contents, m_Contents + m_Length)` including ones that were not copied.

### DIFF
--- a/Code/Core/Strings/AString.cpp
+++ b/Code/Core/Strings/AString.cpp
@@ -53,7 +53,7 @@ AString::AString( const AString & string )
     uint32_t reserved = Math::RoundUp( len, (uint32_t)2 );
     m_Contents = (char *)ALLOC( reserved + 1 );
     SetReserved( reserved, true );
-    Copy( string.Get(), m_Contents ); // copy handles terminator
+    Copy( string.Get(), m_Contents, len ); // copy handles terminator
 }
 
 // CONSTRUCTOR (const char *)
@@ -355,7 +355,7 @@ void AString::Assign( const AString & string )
         // didn't resize then the passed in string is empty too
         return;
     }
-    Copy( string.Get(), m_Contents ); // handles terminator
+    Copy( string.Get(), m_Contents, len ); // handles terminator
     m_Length = len;
 }
 
@@ -456,7 +456,7 @@ AString & AString::operator += ( const AString & string )
             Grow( newLen );
         }
 
-        Copy( string.Get(), m_Contents + m_Length ); // handles terminator
+        Copy( string.Get(), m_Contents + m_Length, suffixLen ); // handles terminator
         m_Length += suffixLen;
     }
     return *this;


ffulin:
Your changes make sense: from a performance standpoint it's better, and it means a copy or assignment results in the same destination AString as the source (that this is not the case right now is pretty odd/wrong).

However, it actually looks like Cache::Publish is currently working ok. It seems it is relying on the fact that a string copy inside EnsurePathExists will copy only to the null inserted at the last slash.  Copying the entire string (as a result of this change) actually causes it to search further in the string than it should.

The entire behaviour of Cache::Publish is pretty dangerous (passing the weird-state string), so I think that should be changed (we can just make a proper copy and truncate that with SetLength() instead of the weird null poking it is doing). It seems like we'll need to make that fix before this change can be integrated.

Let me know if I've misunderstood something :)

dummyunit:
Yes, `Cache::Publish` works now because the only observable behavior of `FileIO::EnsurePathExists` are calls to `FileIO::DirectoryCreate(pathCopy)` and that function only uses data up to the first null.
But non-observable operations (modifications on `pathCopy`) will still work with data in range `[m_Contents, m_Contents + m_Length)` even though there are garbage after the inserted null:
   1. `PathUtils::FixupFolderPath` will apply fixes (convert slashes, append slash) to all data.
   2. Loop at the end of `FileIO::EnsurePathExists` will iterate over all slashes, including ones that are happen to be after the inserted null, this may result in unnecessary (but harmless) calls to `FileIO::DirectoryExists` and `FileIO::DirectoryCreate`.

I don't think this patch will break `Cache::Publish` (it only ensures that after the inserted null there will be filename instead of arbitrary bytes). If anything it may reduce unnecessary calls to `FileIO::DirectoryExists` and `FileIO::DirectoryCreate`.

ffulin:
I think the behaviour of Cache::Plugin is pretty sketchy before and after, so I went ahead and just fixed it (https://github.com/fastbuild/fastbuild/commit/30cdf3a51876971ed10518cbe99cb162da6d4bac) so we don't have to worry about it anymore :)

ffulin:
I've integrated your AString fix into the Perforce mainline. This will be in the next release (v0.95). Thanks for the fix!

PR #314, opened 2017-10-22
Support building with Clang on Linux, support building with Address and Memory sanitizers

dummyunit:
This PR adds support for building with Clang on Linux (Debug, Release, Profile) and building with [AddressSanitizer](https://clang.llvm.org/docs/AddressSanitizer.html) (with GCC and Clang) and with [MemorySanitizer](https://clang.llvm.org/docs/MemorySanitizer.html) (Clang only). More details in the corresponding commits messages.

Main goal is to add required build configurations for a fuzzer based on [libFuzzer](http://llvm.org/docs/LibFuzzer.html) for `BFFParser` which is currently work-in-progress (branch `bff_fuzzer` in my fork).

There is a separate PR (#312) for the first two commits. I also fixed some minor errors found by sanitizers. Let me know if you think that this PR should be split further.

### DIFF
--- a/Code/Core/Core.bff
+++ b/Code/Core/Core.bff
@@ -78,9 +78,9 @@
         Alias( '$ProjectName$-$Platform$-$Config$' ) { .Targets = '$ProjectName$-Lib-$Platform$-$Config$' }
     }
 
-    // Linux (GCC)
+    // Linux (GCC, Clang)
     //--------------------------------------------------------------------------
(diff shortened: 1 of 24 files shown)


dummyunit:
Added 2 more commits (8f1cada, 18bd34a) fixing the last issue reported by MSan: Parts of protocol messages (`IMessage` hierarchy) used for padding and alignment contained arbitrary data when they were sent to the network.

PR #319, opened 2017-10-25
Use traditional file name extension for static libraries on Linux and OSX

dummyunit:
Just a minor consistency fix.


ffulin:
I've integrated this into the Perforce mainline. It will be part of the next source release (v0.95). Thanks!

I hope to do some more general cleanup of the bffs in the future - there is a lot of boilerplate that could be avoided with use of features from more recent versions that would avoid some of these inconsistencies.

PR #321, opened 2017-10-25
Fix crash when trying to access variable from parent scope at global scope

dummyunit:
Found with libFuzzer. Minimal reproducer:
```
.a={^A}
```


ffulin:
Thanks for the fix! I've integrated this into the Perforce mainline. I will be in the next release (v0.95). I added a regression test for this bug as well (https://github.com/fastbuild/fastbuild/commit/4797c698ee00ea9772a62cc266fe5d457d5da856)

PR #322, opened 2017-10-27
Store current variable type explicitly in BFFParser::StoreVariableArray

dummyunit:
This simplifies type compatibility checks when appending new values.

It also fixes crashes in error reporting caused by calls to `var->GetType()` when creating new variable (in which case `var == nullptr`).
Crashes were found with libFuzzer. Minimal reproducers:
```
.s=[].a={.s''}
```
```
.s=[].a={''.s}
```


ffulin:
Thanks for the bug fix and the great repro! I've integrated this into the Perforce mainline. The fix will be in the next release (v0.95). I also added a test to cover this case (https://github.com/fastbuild/fastbuild/commit/3b6867654b8a535734c4f190d2d91be9af83e8ef)

PR #323, opened 2017-10-28
Fixes remote compilation errors treated as system errors on clang

arturmazurek:
On remote workers whenever compilation fails it is reported as a system error when clang is used. This is because clang returns 1 on failed compilations. 

I am unsure about gcc, but it seems to me that this version will be safer anyway.


ffulin:
Oh dear - that looks like a very silly bug. Thanks for the fix!

ffulin:
I've integrated the fix into the Perforce mainline. This will be in the next release (v0.95). I also went and added a test to exercise this code path (https://github.com/fastbuild/fastbuild/commit/a903b161c7d43452344e651810e5253e5aca9b88).

Thanks for the fix!

PR #330, opened 2017-11-05
Allow GCC to check format strings passed to custom printf-like functions, fix related issues

dummyunit:
This adds [attribute `format`](https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html#index-format-function-attribute) to all functions that format messages using printf-like formatting rules.
After adding that attribute 3 kinds of issues were found:
1. Two actual errors (fixed by first commit).
1. Multiple warnings about passing a wrong type (`unsigned long` instead of `unsigned long long`) when passing `int64_t` or `uint64_t`. These aren't real problems because there is no Linux 32-bit build, but to avoid such warnings second commit converts all uses of `%llu` and similar format specifiers to C99 `PRI` macros to allow portable formatting of these types.
1. Multiple warnings about passing a wrong type (`unsigned long` instead of `unsigned int`) when passing `size_t`. Third commit adds conversions to `unsigned int` when `size_t` is being formatted, since in all such cases overflow isn't possible in practice.


dummyunit:
So, it looks like my changes related to `int64_t` broke Windows build:
```
C:\Program Files (x86)\Microsoft Visual Studio 12.0\VC\include\inttypes.h(132) : warning C4005: 'PRIi64' : macro redefinition
        D:\fastbuild\Code\Core/Env/Types.h(22) : see previous definition of 'PRIi64'
```
Sorry about that :) Apparently `ProfileManager.cpp` was including `inttypes.h` for some reason. I fixed it now.

But this raises a question: Why are all sized integer types are defined manually instead of just using `inttypes.h` directly? That include for `inttypes.h` appeared in `ProfileManager.cpp` in 0.90 and by the looks of it didn't cause any issues since.

ffulin:
> Why are all sized integer types are defined manually instead of just using inttypes.h directly? 

Generally I try to avoid all external dependencies (including headers) to minimize compile times and keep the codebase as lean as possible. It's not always possible, or practical, and sometimes (like here) something slips though.

ffulin:
Thanks for the fixes!  Funnily enough, I've added this to pretty much every code base I've ever worked on (since it always finds/prevents some problems), but didn't add it to FASTBuild's source for some reason :). The fixes will be in the next release (v0.95)

PR #331, opened 2017-11-05
Correctly interpret status code of waitpid(), add test

dummyunit:
`WEXITSTATUS` can be used only if `WIFEXITED` returns true, in fact `WEXITSTATUS` returns 0 if the process was terminated by a signal. This resulted in crashed tests being treated as passed by TestNode.

This change expands status checks to handle all state changes (normal termination, termination by a signal, process receiving `SIGSTOP`).

Also added a test for TestNode that verifies that crash of a test executable is detected and treated as test failure.

### DIFF
--- a/Code/Core/Process/Process.cpp
+++ b/Code/Core/Process/Process.cpp
@@ -379,7 +379,22 @@ bool Process::IsRunning() const
 
         // store wait result: can't call again if we just cleaned up process
         ASSERT( result == m_ChildPID );
-        m_ReturnStatus = WEXITSTATUS(status);
+        if ( WIFEXITED(status) )
+        {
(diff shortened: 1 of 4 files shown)


ffulin:
Thanks for the fix! I've integrated your changes into the Perforce mainline. This fix will be in the next release (v0.95).

dummyunit:
It looks like that after the cleanup (4e1b2ca54f8b00d2d69c26a700cd2f05c12ef4f8) files `test_crash.bff` and `test_crash.cpp` are not used anymore, maybe delete them?

ffulin:
Done! (https://github.com/fastbuild/fastbuild/commit/b1a9616c4ca25c04f3d2c508b10e64ed3c5b628e)

PR #332, opened 2017-11-09
Added If() function

clevijoki:
The only operator is a set comparison, that is intended to work like this:

ForEach(.config in .CONFIGS)
{
If(.config in .DEBUG_CONFIGS) {}
If(.config not in .DEBUG_CONFIGS) {}
}

You can also test lists vs lists:
.AB = {'a', 'b'}
.BC = {'b', 'c'}
If(.AB in .BC) {} // is true, evaluates to any of AB in BC

The goal is to allow for branching to happen, so you can use the same platform/config loops to do slightly different things inside of them.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
@@ -15,6 +15,7 @@
 #include "FunctionExec.h"
 #include "FunctionExecutable.h"
 #include "FunctionForEach.h"
+#include "FunctionIf.h"
 #include "FunctionLibrary.h"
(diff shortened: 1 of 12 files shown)


ffulin:
Thanks for the pull request. I've integrated this into the Perforce mainline. This change will be in the next release (v0.96).  

I added some basic documentation and updated the various markup files in (https://github.com/fastbuild/fastbuild/commit/4ae514cc121ccf2046e616cc57c882461449fac6). Of particular note, I removed support for the comma operator functioning as an "or". I can see that this would be useful, but I think overloading , in this way is a little confusing. Instead, I think we'll want to add a variety of explicit operators in a future update.

PR #352, opened 2017-12-19
Fix crash when pressing CTRL-C in a commandline because FBuild singleton was null.

jspelletier:
Hi Franta,

This fixes a 100% crash occuring when building from commandline and pressing CTRL-C because the FBuild singleton was null in the CTRL-C handler. I fixed it by making the FBuildOption static and renamed it to follow the naming convention.

Note: The crash occurs in fastcancel mode.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/FBuild.cpp
+++ b/Code/Tools/FBuild/FBuildCore/FBuild.cpp
@@ -50,6 +50,7 @@
 //------------------------------------------------------------------------------
 /*static*/ bool FBuild::s_StopBuild( false );
 /*static*/ volatile bool FBuild::s_AbortBuild( false );
+FBuildOptions FBuild::s_Options;
 
 // CONSTRUCTOR - FBuild
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for the bug fix. There seem to be two problems. The Ctrl+C handler is created before and is destroyed after the FBuild singleton, so there is a race condition.  Secondly, if using -wrapper, -vs or -ide on the command line, the FBuild singleton is never created. I was able to reproduce the crash in the second case. Is that what you are doing too?

Thought it will fix this crash, I'm a little hesitant to make m_Options static. Firstly, it complicates the lifetime for the object in a way that could cause future bugs for tests which create/destroy multiple builds in one process (I think it's fine right now, but a bug would be easy to introduce - you can see that s_StopBuild and s_AbortBuild are already special because of this).

Secondly, I would like -fastcancel to become the default behavior, removing the access of the option and allowing for the s_StopBuild and s_AbortBuild flags to be merged together. (I will consider -fastcancel to be no longer experimental in v0.95 and will likely make it on by default in v0.96)

With all that in mind, I think we could instead fix the problem by:
 - have AbortBuild check for the validity of the FBuild singleton before checking the "fast cancel" option
 - clear the Ctrl+C handler before destroying the FBuild singleton

I'm happy to make these changes.

Le me know what you think. Is there a scenario I am missing?

jspelletier:
Hello Franta,
I think that should work. Will revert my change once you made the fix.
BTW I was reproducing this using the following options:
-vs -summary -monitor -fastcancel -cache -wait  -dist and was executing this from a command line prompt.

ffulin:
I'll leave this open until we have a fix in place, just for tracking.  (Thanks for the confirm about using -vs)

ffulin:
I went and implemented the CtrlCHandler changes (after doing some much needed cleanup to main.cpp) https://github.com/fastbuild/fastbuild/commit/49e3b46178225d7b0425268ea931da67c351c841

This will be in the next release (v0.95). I'll keep this pull request open until that version is released.

Thanks again for the bug report/pull request.

jspelletier:
Hi Franta,
I have reverted my fix and applied the if (FBuild::IsValid() in the condition as a temporary workaround until we integrate your code(the commit was causing some conflicts because of earlier changes you did for 0.95 so I ended up putting this as a temp workaround.

ffulin:
> I have reverted my fix and applied the if (FBuild::IsValid() in the condition as a temporary workaround

Makes sense. While the IsValid check alone isn't a complete fix, it'll cover the most problematic case and likely solve your issue (all that remains is the very small period of time during shutdown where a Ctrl+C could cause a freed memory deref but you're incredibly unlikely to hit that)

ffulin:
v0.95 has now been released and includes the refactor/fix for the crash. Thanks again for reporting this issue and helping to get it fixed!

Issue #353, opened 2017-12-20
disable compress lead to crash


changshuang:
When i disable the compress of DoBuildWithPreProcessor function, compile with '-dist' crash.
Here is my modify:
```
Compressor c;
c.Compress(job->GetData(), job->GetDataSize());
size_t compressedSize = c.GetResultSize();
job->OwnData(c.ReleaseResult(), compressedSize, true);
```
to
`job->OwnData(job->GetData(), job->GetDataSize());`

Is there some place depend on compress ?

changshuang:
![image](https://user-images.githubusercontent.com/8707792/34201275-efaa9750-e5ae-11e7-98c9-8ca6bac9ad79.png)
Compiling unreal4 crash.

As you see, it seems that if don't use compress， the destruct of job will do not detect the end of m_Data.

changshuang:
In our distributed network，there are machines with 56-core. When building with fastbuild, the memory is overhead. The compress take double memory. So i want to disable it to have a try.

changshuang:
Can you figure out which places depend on compress? I  have a try to find it, but failed. Remote jobs read the IsDataCompressed when being deserialized.

ffulin:
Do you mean the memory usage on the fbuildworker.exe (and not the fbuild.exe) is a problem? (I'm assuming that's the case, since the local fbuild.exe has a 1GiB memory cap for distributed jobs.)

Turning off compression for data sent to the worker will actually increase overall memory usage during compilation on the worker given how things are implemented right now.

Right now the worker keeps the buffer it receives in memory for the lifetime of the compilation job. If this is compressed (and it usually is) then before compiling, it will temporarily allocate a new buffer to decompress into, write that data to disk and free this temporary buffer. 

If you change the fbuild.exe to send an uncompressed buffer, it will no longer create this temporary allocation, but it now will keep the uncompressed buffer for the duration of the compilation (and it's usually more than 2x the size of the compressed buffer).

The good news is that we actually don't need to keep the source buffer (compressed or not) after we've written to disk, so we can actually greatly reduce the overall memory footprint without disabling compression.

If you change the end of ObjectNode::WriteTmpFile to to look like this:
```C++
    // On remote workers, free compressed buffer as we don't need it anymore
    // This reduces memory consumed on the remote worker.
    if ( job->IsLocal() == false )
    {
        job->OwnData( nullptr, 0, false ); // Free compressed buffer
    }
    return true;
```
[...]

PR #356, opened 2017-12-21
Job::OwnData add self protect

changshuang:


ffulin:
Is there a need for this change? Set a job to own the data it already owns seems like it would be a bug and should not be handled gracefully.

changshuang:
This is just a prevention of free self. E.g: #353 . If user pass its own data to function "owndata", the function will free data firstly, then the data is invalid. But next it will be assigned to m_Data as normal data. It will lead crash sometimes because of using the invalid data in memroy.

Of cause, this can be avoided by user's judgement. But as a offical c++ code, i thind it is necessary to add a self protect. That is just my opinion.

ffulin:
I've reworked this code a little bit in order to expose the memory limit via Settings(). As part of this change (https://github.com/fastbuild/fastbuild/commit/53560f6ca39a88e2ad79b121cf551589a5fcd142) I've made redundant calls to to OwnsData be caught by an ASSERT. This change will be part of the next release (v0.95). I'll leave this pull request open to track that.

ffulin:
v0.95 has now been released and the source includes protection for misuse of this API. Thanks for reporting this issue!

PR #361, opened 2018-01-05
By default show progress bar only if stdout goes to the terminal

dummyunit:
This allows to redirect `FBuild` output to file or pager (like `less`) without needing to specify `-noprogress` to get a readable result.

Also added option `-progress` to force the old behaviour – to show progress bar unconditionally.

Windows implementation of redirect detection supports both native console windows and Cygwin/MSYS terminals. I was unable to test it with MSVS 2013 (don't have it installed right now), but it works fine when `FBuild.exe` is compiled with MSVS 2017.

There is no implementation for MacOS. In theory it should be the same as for Linux, but since I can't check that (don't have a Mac) there is a stub currently.

### DIFF
--- a/Code/Core/Env/Env.cpp
+++ b/Code/Core/Env/Env.cpp
@@ -228,6 +228,54 @@ void Env::GetExePath( AString & output )
     #endif
 }
 
+// IsStdOutRedirected
+//------------------------------------------------------------------------------
+/*static*/ bool Env::IsStdOutRedirected()
+{
+    #if defined( __WINDOWS__ )
+        HANDLE h = GetStdHandle( STD_OUTPUT_HANDLE );
+        ASSERT( h != INVALID_HANDLE_VALUE );
+        if ( h == NULL )
+            return true; // There is no handle associated with stdout, this is essentially a redirection to NUL
+
+        // Check if the handle belongs to a console window
+        DWORD unused;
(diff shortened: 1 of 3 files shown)


ffulin:
Nice one. I can extend/test this on OSX, since as you say it should be the same as the Linux implementation.

Is there a reason we can't use isatty on Windows? (or more specifically _isatty) It seems to do what we want as described here: https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/isatty

dummyunit:
> Is there a reason we can't use isatty on Windows? (or more specifically _isatty)

We _can_, but there are several reasons not to:
1. It does [parameter validation](https://docs.microsoft.com/en-us/cpp/c-runtime-library/parameter-validation), which by default crashes application when something goes wrong. That possibility isn't appealing for our purely cosmetic feature.
1. It documentation says that it detects if "a file descriptor is associated with a **character device**", and indeed it returns non-zero when stdout is redirected to `NUL`, `CON`, `COM1`, etc. This is not not big problem (in case of `CON` it maybe even beneficial), just feels inconsistent (on Linux `isatty` returns 0 on redirections to `/dev/null` and `/dev/ttyS*` (COM ports)).
1. We still need to get a Windows handle via `GetStdHandle` for Cygwin/MSYS checks, so we can't save on amount of system calls that way.

PR #362, opened 2018-01-05
Update bash completion script

dummyunit:
* Added new FBuild options.
   * Disabled default completion source that was adding files from current directory to the list (which are not very useful as an argument to FBuild).
   * Disabled completion after `-cachetrim` to indicate that an argument is required.


ffulin:
Thanks for the update. I've integrated this into the Perforce mainline. This will be in the next release (v0.95)

Issue #146, opened 2016-09-15
Config paths with spaces are not passed with quotes to the subprocess


liamkf:
Hi there! I've been stumbling my way through getting [msfastbuild](https://github.com/liamkf/msfastbuild) working, which attempts to build Visual Studio projects and solutions with FASTBuild, with various degrees of success getting it to work for some unusual projects... :D

It was noticed in [this bug](https://github.com/liamkf/msfastbuild/issues/3) that spaces in the -config bff path cause the args to not be passed correctly to the spawned process when used with -ide or -wrapper:

Works:
` $>fbuild -config "C:/space bees/fbuild.bff"`
Can't find "C:/space":
` $>fbuild -ide -config "C:/space bees/fbuild.bff"`

It looks like it's down to the configFile not being quoted when it gets passed along. Putting quotes around the file seems reasonable enough, which is what I've done [here](https://github.com/liamkf/fastbuild/commit/ccca403fb01c457c91105772192b15e966d56569). It's obviously not a big change, but I can also make it a pull request after we merge the visualizer changes back into the branch if you'd prefer!

ffulin:
Your diagnosis and fix makes total sense. I'd be happy to integrate that fix as-is immediately if you like. I think it's nice to keep changes as self-contained as possible to keep track of things.

ffulin:
Your https://github.com/liamkf/msfastbuild is pretty intriguing - it's an often requested feature and one of the few areas than Incredibuild outshines FASTBuild currently. I'd pretty much abandoned any hope of creating such a system given the crazy behaviour I've seen in Solution/Project files (things like conditionals tied to the selected project config that are only evaluated on load and not when you switch configs for example). I guess I was thinking of a perfect system, rather than one that might be able to handle enough cases to be useful. It looks like you've made some great progress!

liamkf:
Sounds good! I caught a few places where I wasn't escaping arguments myself along the way.

Re: msfastbuild, yeah I think going for complete coverage of bizarre VS/MSBuild projects that are out there seems a ways off, but even now I think it's usable for most projects. @jairbubbles is doing a good job of kicking the tires. :) We definitely really like FASTBuild (thanks for making it!), and I think getting more people in the door with their existing projects could really help its adoption.

[...]

PR #366, opened 2018-01-16
Surround config file path with "

liamkf:
To avoid problems with spaces in the path. As mentioned here https://github.com/fastbuild/fastbuild/issues/146


ffulin:
Thanks for the fix. I've integrated this into the Perforce mainline. This fix will be in the next release (v0.95)

PR #369, opened 2018-01-31
Fix MS linker error when command line length is exactly equal to 32767

belkiss:
Fix MS linker error when command line length is exactly equal to 32767, hence not using response file.

The documentation indicates that this case should not require a response file, but it doesn't work.
We also verified that one character less in the command line fixes the issue.

Note that the code hasn't been made MS specific, so that same case on other platform will use response files as well, even though it might not need it.


belkiss:
Np! Thanks for integrating :)

PR #370, opened 2018-01-31
FastBuild doesn't have to suppress the broker file when a build is finished.

belkiss:


ffulin:
The existing behavior was deliberate. The idea is that when exiting the FBuildWorker, this is our last chance to try and delete the token, which might have failed to be cleaned up previously (due to a temporary network outage for example). Was this causing a problem of some kind?

I'm also a little confused by the title of this issue. The code you changed should only execute when you close the FBuildWorker. It should not be related to builds finishing. Can you provide a bit more detail about what you mean so I can be sure I understand? Thanks :)

belkiss:
This has been found during investigations of timeout.
I'm not the one who fixed the bug, bug looking at the code, it seems that the workerbrokerage is also run on the client.
Meaning that if you run a worker at the same time on the machine, the build completion will remove your worker file.

ffulin:
Thanks for the clarification. The local build deleting the token is definitely something that should not happen.

In case it helps with your other investigation, it's worth noting that the list of tokens is only read from the brokerage by a remote machine once at the start of a build, after which time the brokerage is not used at all. So this bug should only affect the discovery of workers, not their ongoing use.

ffulin:
Thinking about this some more, I think the worker trying to delete the tokenon exit isn't actually that important, so I'm just going to integrate this fix as-is. This fix will be in the next release (v0.96). Thanks as always!

belkiss:
Thank you for the info, and for integrating this, I'll keep you posted if we do further investigations.

PR #372, opened 2018-02-02
FBuildWorker: Don't show notification when minimizing and on start-up

v1ne:
Please consider this or a different mechanism to solve two issues with the "FBuildWorker minimized to tray." notification. On applies to all versions of Windows (AFAIK), the other only to recent Windows 10:

- Showing and hiding the window on start-up steals the focus. This is quite
  annoying because it prevents me from using my system until every background
  process like FBuildWorker has started up. I firmly believe that software should not
  steal a user's focus unless it comes to the foreground and stays there.

- Since Windows 10 1709, notifications for tray icons are added to the
  notification area. They will stay there until the user dismisses them.
  This is distracting since Windows shows that I have one new notification that 
  after /every/ log-in + start of FBuildWorker. It reads:
  "FBuildWorker minimized to tray."

### DIFF
--- a/Code/Tools/FBuild/FBuildWorker/Worker/WorkerWindow.cpp
+++ b/Code/Tools/FBuild/FBuildWorker/Worker/WorkerWindow.cpp
@@ -206,14 +206,17 @@ void WorkerWindow::UIUpdateThread()
         m_Menu = CreatePopupMenu();
         AppendMenu( m_Menu, MF_STRING, ID_TRAY_EXIT_CONTEXT_MENU_ITEM, TEXT( "Exit" ) );
 
-        // Display the window, and minimize it
-        // - we do this so the user can see the application has run
-        ShowWindow( (HWND)GetHandle(), SW_SHOW );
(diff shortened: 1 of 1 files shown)


ffulin:
I agree that stealing the focus is not desirable, so avoiding that seems like a good change.

The notification behavior I'm a little less sure about. I find applications that minimize with no obvious indication that they do so annoying. Similarly, launching FASTBuild and having it start without any obvious indication it is running is potentially problematic also. Perhaps the behavior could be changed. Some ideas:
 - not showing a balloon tip on startup, unless launched directly from Explorer (i.e. scripted or on startup launches are silent, but double clicking the icon or manually minimizing would not be silent)
 - seeing if there is a way to not have the message log in the notification center

I'm open to other ideas as well.

ffulin:
I decided to not complicate the behavior and simply integrate your fix as is. These changes will be in the next release (v0.96). Thanks :)

v1ne:
Thanks a lot! 💐

PR #377, opened 2018-02-05
Fix name of ExecExecutable property

dummyunit:
I was updating Vim syntax highlight for FASTBuild 0.95 and [my script](https://github.com/dummyunit/vim-fastbuild/blob/master/check_properties.sh) reported this string as a potential new property: `ExecExcecutable`. This is suspiciously similar to existing property `ExecExecutable` and is probably a typo.


ffulin:
Thanks for the fix - this is indeed an error (that string is only used in the error message if the property is missing/bad). I've integrated this fix into the Perforce mainline. It will be in the next release (v0.96)

PR #378, opened 2018-02-06
Fix path simplification unit-test on non-windows platforms

belkiss:
Fix dual forward slash leftover in the expected simplified result for non-windows platforms.

Sorry about that...


ffulin:
No worries - thanks for the quick fix ;)

PR #379, opened 2018-02-06
Fix missing include uio.h causing compile error on linux

belkiss:
The line `#include <sys/uio.h>` is also needed on linux, otherwise we get:

`error: ‘writev’ was not declared in this scope`

Thanks!


ffulin:
Cheers!

PR #380, opened 2018-02-08
Reset FLAG_STATIC_ANALYSIS_MSVC when /analyze- is specified

dummyunit:
Previously presence of option `/analyze-` in `.CompilerOptions` was enabling `FLAG_STATIC_ANALYSIS_MSVC`. This resulted in failing cache writes because FASTBuild was trying to open `.nativecodeanalysis.xml` files which weren't produced in that case.

This change resets `FLAG_STATIC_ANALYSIS_MSVC` when `/analyze-` is encountered.


ffulin:
Thanks! This fix will be in the next release (v0.96)

PR #382, opened 2018-02-10
GUID generation based on full path.

rickvanprim:


Code/Tools/FBuild/FBuildCore/Helpers/VSProjectGenerator.h
```diff
@@ -113,7 +112,6 @@ class VSProjectGenerator
                                         AString & outRelativeFileName );
 
     // project details
-    AString m_ProjectName;
```
rickvanprim: This wasn't used for anything other than GUID generation, if the GUID wasn't set.

Code/Tools/FBuild/FBuildCore/Graph/VCXProjectNode.cpp
```diff
@@ -63,6 +63,12 @@ VCXProjectNode::VCXProjectNode( const AString & projectOutput,
 
     // depend on the input nodes
     m_StaticDependencies.Append( paths );
+
+    // generate GUID if not specified
+    if ( m_ProjectGuid.IsEmpty() )
+    {
+        VSProjectGenerator::FormatDeterministicProjectGUID( m_ProjectGuid, m_Name );
```
ffulin: I think we need to truncate the path to be relative to the working directory so that the same GUID is generated for two different users who might have different paths.

ffulin:
I just realized that this pull request is against the master branch. Pull requests are only accepted against the dev branch. Would you mid re-creating the pull request please?

ffulin:
Thanks for the fix. This will be in the next release (v0.96)

PR #386, opened 2018-02-14
Not using LIB from Settings when looking for libs

fran6co:
`LinkerNode` should request the `LIB` variable from FBuild directly instead of going to the environment.


ffulin:

PR #387, opened 2018-02-14
Non MSVC library flags are ignored

fran6co:
They should start with `-`

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
@@ -824,6 +824,29 @@ void LinkerNode::GetAssemblyResourceFiles( Args & fullArgs, const AString & pre,
     return ( AString::StrNCmpI( token.Get() + 1, arg, argLen ) == 0 );
 }
 
+// IsStartOfLinkerArg
+//------------------------------------------------------------------------------
+/*static*/ bool LinkerNode::IsStartOfLinkerArg( const AString & token, const char * arg )
+{
+    ASSERT( token.IsEmpty() == false );
+
+    // Args start with -
+    if ( token[0] != '-' )
+    {
+        return false;
+    }
+
+    // Length check to early out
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for the fix. I think it would be better to fix the GetOtherLibsArg function instead (so that it behaves consistently). What do you think?

fran6co:
That's probably better, I just tried to find a minimal fix for this.

fran6co:
@ffulin do you think it's better like this?

ffulin:
I took a look at your new version, but unfortunately that code doesn't compiler (cannot add two pointers).

I think the correct fix would be to implement a new IsStartOfLinkerArg which would be a slightly altered version of IsStartOfLinkerArg_MSVC.

IsStartOfLinkerArg would look something like this:
```C++
/*static*/ bool LinkerNode::IsStartOfLinkerArg_MSVC( const AString & token, const char * arg )
{
    ASSERT( token.IsEmpty() == false );

    // Args can start with -
    if ( token[0] != '-' )
    {
        return false;
    }

    // Length check to early out
    const size_t argLen = AString::StrLen( arg );
    if ( ( token.GetLength() - 1 ) < argLen )
    {
        return false; // token is too short
    }

    // Args are case-sensitive
    return ( AString::StrNCmp( token.Get() + 1, arg, argLen ) == 0 );
}
```
(the difference being that its case-sensitive and only checks for '-').

Then we could simply call that in the else clause of GetOtherLibsArg:
{code}
    else
    {
        if ( IsStartOfLinkerArg( *it, arg ) == false )
        {
            return false; // not our arg, not consumed
        }
    }
{code}

fran6co:
@ffulin fixed

ffulin:
Thanks for the fix. I've merged the pull request and the fix will be in the next release (v0.96). I went ahead and added some regression tests as well (https://github.com/fastbuild/fastbuild/commit/9e240dcbe668abf9afe1737d6e1cab3df9ac64ab).

PR #388, opened 2018-02-16
CopyDirNode flags SourceExcludePaths as MetaPath.

rickvanprim:


ffulin:
Thanks for the fix. This will be in the next release (v0.96)

PR #389, opened 2018-02-16
Fixes OSX compilation

fran6co:
Getting error that `uintptr_t` doesn't exist.


ffulin:
Thanks for reporting this. I've applied your fix, but in a slightly different way: https://github.com/fastbuild/fastbuild/commit/108a131dd9bc319032acab1965039e09810b4394

PR #392, opened 2018-02-22
BFF fixes for Linux

dummyunit:
First commit fixes build in configuration `x64Linux-ASan` (with GCC) by removing compiler options related to fuzzer, which GCC don't support.
Second commit applies BFF cleanup changes from recent commits to `BFFFuzzer.bff`.


ffulin:
Thanks for the fix!

PR #393, opened 2018-02-26
Store generated unity files in different directories on different platforms

dummyunit:
Generated unity files have different contents on different platforms (`\` on Windows vs `/` on Linux/OSX and likely a different absolute paths on Linux and OSX). So it seems logical to use different places to store them.

Main goal of this change is to allow to use the same repository checkout (by using network share or virtual machine shared folder) on different platforms without triggering rebuild on one platform because previous build on the other platform modified shared unity files.

### DIFF
--- a/Code/Core/Core.bff
+++ b/Code/Core/Core.bff
@@ -9,7 +9,7 @@
     Unity( '$ProjectName$-Unity' )
     {
         .UnityInputPath             = '$ProjectPath$\'
-        .UnityOutputPath            = '$OutputBase$\Unity\$ProjectPath$\'
+        .UnityOutputPath            = '$UnityOutputBase$\$ProjectPath$\'
         .UnityInputExcludePath      = '$ProjectPath$\CoreTest\' // Exclude Tests
     }
 
--- a/Code/Core/CoreTest/CoreTest.bff
+++ b/Code/Core/CoreTest/CoreTest.bff
@@ -9,7 +9,7 @@
     Unity( '$ProjectName$-Unity' )
     {
         .UnityInputPath             = '$ProjectPath$\'
-        .UnityOutputPath            = '$OutputBase$\Unity\$ProjectPath$\'
+        .UnityOutputPath            = '$UnityOutputBase$\$ProjectPath$\'
     }
 
     // Executable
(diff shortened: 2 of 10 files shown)


ffulin:
It's not really possible to run multiple builds from different machines in the same shared directory structure since the database (.fdb) is not portable between operating systems or machines. With that in mind, is this still a useful change?

dummyunit:
Technically you are right, an attempt to use the same .fdb file from different platform results in error message about database being moved. But it can be easily worked around by using proxy .bff as root on different platforms.

In our company we use the following setup for over a year without any issues (related to FASTBuild at least):
* Host system is Windows (7 or 10), most of development tasks (editing, git, debugging system-independent bugs) are done on it.
* Some Linux disto (people use what they know/like more) is running inside VirtualBox. Main repo which is on host's file system is mounted via VBox shared folder inside this Linux guest.
* On the Linux guest people run fbuild with an extra argument: `-config fbuild.linux.bff`. And `fbuild.linux.bff` contains one line: `#include "fbuild.bff"`

About usefulness:
This setup allows us to check if the code builds correctly and run tests on other platform just by Alt-Tab'ing to VM window, no need to copy files back and forth or do dozens of temporary commits (and push/pull them around). Overall this is a major time saver.

Of course, for this technique to work correctly, the build configuration should not modify same files when building on different platforms. For most build steps this is guarantied by the fact that results are placed in directories named after build configurations. The two other places where attention is required are Unity and Exec nodes.

ffulin:
Thanks for the extra context. Somewhat related to this, I was considering adding a built-in variable for the host platform (e.g. _HOST_PLATFORM_ which would be a string defined to "windows", "osx" or "linux"). That could be used directly in the path construction avoiding the #ifdefs. What do you think?

dummyunit:
> `_HOST_PLATFORM_`

I don't think it would be very useful. Because the value is fixed this variable can't cover all similar cases -- sometimes you may need/want a different value e.g `"win32"` instead of `"windows"`. Plus people already can construct such variable if they really need it.
And in our case `#if` blocks are big and contain all platform dependent parts together, so using this variable wouldn't make a difference in complexity.

ffulin:
This will be part of the next source release (v0.96). Thanks for the pull request!

PR #394, opened 2018-02-26
Fix running FBuildTest from Code directory

dummyunit:
After FBuildTest cleanup (02e2979f) it became possible to successfully run FBuildTest from any subdirectory of the Code directory, but not from Code directory itself.
This change fixes corner case in `FBuildTest::GetCodeDir()` that prevents that.


ffulin:
Nice one. Thanks!

ffulin:
This fix will be in the next source release (v0.96). Thanks again!

PR #395, opened 2018-02-26
Reuse base paths to VS an WinSDK in fbuild.bff for TestZW

dummyunit:
This allows to override paths to VS and WinSDK installation places once in the corresponding .bff in `External/SDK/` folder.
Without this change it has to be done here too, otherwise TestZw fails with this error message:
> fatal error C1107: could not find assembly 'platform.winmd': please specify the assembly search path using /AI or by setting the LIBPATH environment variable


ffulin:
Good stuff - thanks. I'll probably move the actual path definitions into the new SDK bffs at some point to avoid the need for the ifdefs, but this is great to make it centralized as intended.

ffulin:
This will be in the next source code release (v0.96). Thanks again!

PR #396, opened 2018-02-26
Fix pthread leaks

dummyunit:
POSIX threads that are created as joinable (which is the default) have to be joined (or marked as detached) before all resources used by them are released, see https://man7.org/linux/man-pages/man3/pthread_create.3.html#NOTES , second paragraph.
This issue was found by [ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html)

### DIFF
--- a/Code/Core/Network/Network.cpp
+++ b/Code/Core/Network/Network.cpp
@@ -101,6 +101,10 @@
 
         break; // success!
     }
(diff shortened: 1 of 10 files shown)


ffulin:
Thanks for the bug fix! This will be in the next release (v0.96)

PR #397, opened 2018-02-26
Call ShutdownAllConnections in dtor of all TCPConnectionPool descendants

dummyunit:
This fixes [data race on vtable pointer](https://github.com/google/sanitizers/wiki/ThreadSanitizerPopularDataRaces#data-race-on-vptr) caused by destruction of `TCPConnectionPool`-based class in one thread while there are connection threads that execute virtual functions (`OnReceive`) of that class.
This issue was found by [ThreadSanitizer](https://clang.llvm.org/docs/ThreadSanitizer.html).


ffulin:
Nice one! This will be in the next source release (v0.96). Thanks as always for all your fixes!

PR #399, opened 2018-02-28
Protect build output collection in tests with a Mutex

dummyunit:
Output messages are produced by multiple worker threads, but they were concatenated into a single `AString` in `FBuildTest::LoggingCallback` without any synchronization.
This issue was found by ThreadSanitizer.


ffulin:
Good find. I'm kind of surprised this hasn't been crashing fairly frequently.

ffulin:
Thanks for the fix. This will be in the next source code release (v0.96)

PR #400, opened 2018-02-28
Fix crash in Client::OnDisconnected when connection closes before UserData is set

dummyunit:
`Client::LookForWorkers` tries to open a connection by calling `TCPConnectionPool::Connect` and then, if it succeeds, sets pointer to `ServerState` object as `ConnectionInfo::m_UserData`.

The problem is that doing so after `Connect` is finished is too late. All code in `Client` that calls `GetUserData` assumes that returned value is correct, including callbacks `OnReceive` and `OnDisconnected` which are called in a thread created by `Connect`.
It is possible that these callbacks will be called in a new thread before `Connect` returns in a thread executing `LookForWorkers`.

For `OnReceive` this won't result in a crash because both `OnReceive` and `LookForWorkers` acquire `m_ServerListMutex`.
But crash in `OnDisconnected` is easily reproducible in AddressSanitizer and ThreadSanitizer builds of FBuildTest in TestWith1RemoteWorkerThread test.

To fix this problem we set UserData before connection thread is started.

### DIFF
--- a/Code/Core/Network/TCPConnectionPool.cpp
+++ b/Code/Core/Network/TCPConnectionPool.cpp
@@ -190,7 +190,7 @@ bool TCPConnectionPool::Listen( uint16_t port )
 
 // Connect
 //------------------------------------------------------------------------------
-const ConnectionInfo * TCPConnectionPool::Connect( const AString & host, uint16_t port, uint32_t timeout )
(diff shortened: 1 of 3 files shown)


ffulin:
Thanks for the fix. I've integrated the change - this will be in the next release (v0.96)

Issue #107, opened 2016-06-28
removing a .cpp from a DirectoryListNode does not cause a rebuild


lucasmeijer:
The DirectoryListNode.DoBuild() does get called, and it actually searches the filesystem, however the ExeNode does not realize that the there's one fewer ObjectNode to link, and thinks everything is just dandy,  instead of relinking without the objectfile belonging to the removed cpp file.

ffulin:
Hello

I've reproduced this bug. The linking will be correctly performed (i.e. no stale objects are ever used) if something else causes it to trigger, but if the only action taken is to delete a cpp file, it indeed doesn't trigger a re-link as it should.

It's almost hard to believe it took this long for this bug to be reported, but it's because most users use Unity to accelerate compilation, in which case deleting a cpp file causes a change in at least one unity cpp file and therefore rebuilds/relinks.

Thanks for reporting it!  I'll look into creating a fix.

lucasmeijer:
no worries, I'm not blocked by this, so if you don't need a fix, don't make one for me, just reporting what I found :)

ffulin:
It's definitely a bug that needs fixing. I'll keep this report open until it's resolved.  Thanks again :)

ffulin:
This dependency graph has been changed in a way that fixes this issue: https://github.com/fastbuild/fastbuild/commit/735ba92dfa2d37f0d53dce285a120af091903808). This fix will be in thenext release (v1.00). I'll leave this issue open until that version is released.

ffulin:
v1.00 has now been released. It includes the overhauled dependency engine which fixes this issue (and others).

PR #401, opened 2018-03-03
TestNode supports Paths.

rickvanprim:


Code/Tools/FBuild/FBuildCore/Graph/TestNode.cpp
```diff
@@ -52,11 +61,43 @@ bool TestNode::Initialize( NodeGraph & nodeGraph, const BFFIterator & iter, cons
...
+    {
+        return false; // GetDirectoryListNodeList will have emitted an error
+    }
+    ASSERT( testInputPaths.GetSize() == m_TestInputPath.GetSize() ); // No need to store count since they should be the same
+
+    // Store Static Dependencies
+    m_StaticDependencies.SetCapacity( 1 + testInputFiles.GetSize() + testInputPaths.GetSize() );
+    m_StaticDependencies.Append( executable );
+    m_StaticDependencies.Append( testInputFiles );
+    m_StaticDependencies.Append( testInputPaths );
```
ffulin: We're depending on the directory listings nodes here, but not the actual files found by the directory listing. A DoDynamicDependencies pass is needed to add dynamic dependencies on the results of the DriListNodes. Without this, changes to the contents of those directories will not cause a recompile. CSNode is probably the closest example of what's needed.

Note too that due to a limitation in the dependency engine, deleting a file inside the specified directories won't trigger a rebuild as noted here: https://github.com/fastbuild/fastbuild/issues/107 That's not something you can fix (it will be fixed by a core change to dependency management in the future), but I just wanted to call it out.

ffulin:
Thanks for the enhancements! I've integrated this into the mainline. It'll be part of the next release (v0.96)

PR #408, opened 2018-04-03
Add new compiler: orbis-wave-psslc

gkoreman:
I was previously using simple distribution mode to support the orbis wave shader compiler, however adding it as a real compiler seems like a better choice.


ffulin:
If this compiler flavor is fully compatible, it's great to handle it as a first class citizen. Thanks for the pull request.

ffulin:
I've integrated this into the mainline. This will be in the next release (v0.96). I also updated the compiler detection to detect the exe if the .exe extensions is omitted. Thanks again for the pull request!

PR #409, opened 2018-04-03
SimpleDistributionAllowCache

gkoreman:
The simple distribution mode compilers did not have the ability to cache their results. This pull request adds that ability.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
@@ -29,6 +29,7 @@ distribution, caching and more.
   .AllowDistribution            // (optional) Allow distributed compilation (if available) (default: true)
   .ExecutableRootPath           // (optional) Override default path for executable distribution
   .SimpleDistributionMode       // (optional) Allow distribution of otherwise unsuported "compilers" (default: false)
+  .SimpleDistributionAllowCache // (optional) Allow simple distribution compilers to cache their results (default: false)
   .CustomEnvironmentVariables   // (optional) Environment variables to set on remote host
   .ClangRewriteIncludes         // (optional) Use Clang's -frewrite-includes option when preprocessing (default: true)
   .VS2012EnumBugFix             // (optional) Enable work-around for bug in VS2012 compiler (default: false)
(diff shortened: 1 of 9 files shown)


ffulin:
Supporting caching for the Simple compilers is a good change, but I don't think this change will work correctly as is.  Specifically, there is nothing in place to hash the input from the simple compiler. With supported compilers, the preprocessor is run, and the output of that is hashed to generate the cache key. For the simple compiler case, we'll need to manually read in the input and hash it ourselves. 

It's possible that this might seem to work, but would likely be caching different objects with the same key. It'd be good to make sure that the caching code asserts or otherwise fails in the future to catch this case. (I suspect that we can assert on the part of the cache key representing the input being non-zero)

gkoreman:
The simple compiler has a fairly major gotcha - which is that it doesn't have a preprocessor, so if the resource depends on anything except for the single input file you will not get correctly hashed results. This is an issue, and one that can be solved by representing multiple input files as dependencies. 

If the SimpleCompiler only uses a single input file and has a single output file (eg, texture compressor) then the SimpleDistributionMode reads the input into an ObjectNode in preparation for distributing it to other clients. When we GetCacheName it uses the job->GetData() to calculate the input data hash as it normally would the preprocessed data.

KeyB and KeyC are hashes of the args and the compiler (as they are with regular compilers) so I don't think we have an issue there either. The test I added seems to be caching correctly, is there another test I can add that shows the behavior you describe?

gkoreman:
In practice I have been using SimpleDistributionMode as a compiler that only ever works on a single file, so caching based on source file hash is safe in my situation. I've added #461 to allow object nodes to specify their dependencies and looking back at this PR perhaps a good addition would be to allow the compiler to specify itself as a PreBuildDependencyPreprocessor? I could extend this PR to include something like that if you think that is the best route, or we can treat that as another task.

ffulin:
I think it would be nice if we could decouple this change from .SimpleDistributionMode. i.e. If we rename your new flag to .SimpleCacheMode and ensure it works whether .SimpleDistributionMode is set or not. I imagine that some further modifications would be required if this is relying on code paths used by .SimpleDistributionMode.

Issue #162, opened 2016-11-06
Using a built executable as compiler


Leandros:
I'm trying to use an executable, which is build as part of the project, as a compiler for other objects in the project.

FASTBuild, unfortunately, refuses this:

```
BFF file 'C:\perforce\thor\res\models\fbuild.bff' has changed (reparsing will occur).
C:\perforce\thor\res\models\fbuild.bff(22,1): FASTBuild Error #1102 - ObjectList() - 'Compiler' ('C:\Users\Arvid\workspace\perforce\thor\_out\bin\objbin.ex e') is of unexpected type 'Exe'. Expected 'Compiler'.
ObjectList('res-models-obj')
^
\--here
```

Said executable is compiled via `Executable` and I want to use it as `.Compiler` in `ObjectList`.

Is there any way to make this work?

ffulin:
FASTBuild expects the .Compiler to point something declared with Compiler(), but for convenience it automatically upgrades a single file to a Compiler() declaration for you.  If you provide anything more complex, it can't do this automatic conversion, and so you get this error.

It should be possible to make this work as follows:
```C++
Executable( 'CustomExecutable' ) // Your executable
{
    ...

Compiler( 'CustomCompiler' )
{
    .Executable = 'CustomExecutable'
    ...

.Objectlist( 'ObjectList' )
{
    .Compiler = 'CustomCompiler'
```

Leandros:
That was also one of my first ideas to try, sadly FASTBuild complains the compiler I try to define has already been defined:

```
BFF file 'C:\Users\Arvid\workspace\perforce\thor\res\models\fbuild.bff' has changed (reparsing will occur).
C:\Users\Arvid\workspace\perforce\thor\res\models\fbuild.bff(10,1): FASTBuild Error #1100 - Compiler() - Target 'C:\Users\Arvid\workspace\perforce\thor\_out\bin\objbin.exe' already defined.
Compiler('objbin-compiler')
^
\--here
```

ffulin:
I'm afraid you are right: this won't currently work :(

To make this work, a change is required to Compiler() to allow it to reference the primary executable from another target, instead of requiring it to be the entry that declares that file. This would have the added benefit of allowing multiple Compiler() instances to reference the same primary executable, which would be a useful improvement to the flexibility of FASTBuild.

I'll investigate further to see how easy/difficult this change might be.

Leandros:
Merci. 🙇

gkoreman:
[...]

PR #410, opened 2018-04-05
CompilerNode can now be a build dependency

gkoreman:
Fixes issue #162. This is a rather large change so please let me know if there is anything you would like changed or if you see issues.


Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
```diff
@@ -524,6 +525,51 @@ bool Function::GetDirectoryListNodeList( NodeGraph & nodeGraph,
     return true;
 }
 
+// GetCompilerNode
+//------------------------------------------------------------------------------
+bool Function::GetCompilerNode(NodeGraph & nodeGraph, const BFFIterator & iter, const AString & compiler, CompilerNode * & compilerNode) const
```
gkoreman: Moved GetCompilerNode to Function base class as it is now used in CSNode, ObjectNode, and ObjectListNode.

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionCompiler.cpp
```diff
@@ -41,19 +41,7 @@ FunctionCompiler::FunctionCompiler()
 //------------------------------------------------------------------------------
 /*virtual*/ bool FunctionCompiler::Commit( NodeGraph & nodeGraph, const BFFIterator & funcStartIter ) const
 {
-    AStackString<> name;
-    if ( GetNameForNode( nodeGraph, funcStartIter, CompilerNode::GetReflectionInfoS(), name ) == false )
-    {
-        return false;
-    }
-
-    if ( nodeGraph.FindNode( name ) )
```
gkoreman: If compilers use the m_AliasForFunction name does it still need to be checked for uniqueness?

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionCompiler.cpp
```diff
@@ -65,8 +53,7 @@ FunctionCompiler::FunctionCompiler()
         return false;
     }
 
-    // handle alias creation
-    return ProcessAlias( nodeGraph, funcStartIter, compilerNode );
+    return true;
```
gkoreman: Is not creating an Alias for CompilerNodes the correct thing to do?

Code/Tools/FBuild/FBuildCore/Graph/CSNode.h
```diff
@@ -31,7 +31,7 @@ class CSNode : public FileNode
     virtual bool DoDynamicDependencies( NodeGraph & nodeGraph, bool forceClean ) override;
     virtual BuildResult DoBuild( Job * job ) override;
 
-    inline Node * GetCompiler() const { return m_StaticDependencies[ 0 ].GetNode(); }
+	CompilerNode * GetCompiler() const;
```
gkoreman: GetCompiler functions were changed to return CompilerNodes and implementation moved to the cpp to keep includes low.

Code/Tools/FBuild/FBuildTest/Data/TestCSharp/csharp.bff
```diff
@@ -8,9 +8,10 @@
 Using( .StandardEnvironment )
 Settings {}
 
-Alias( 'CSharpCompiler' )
+Compiler( 'CSharpCompiler' )
```
gkoreman: Changed the CS test to actually use a compiler instead of a FileNode with an alias. This may break people's projects. Any ideas?

Code/Tools/FBuild/FBuildCore/Helpers/SLNGenerator.cpp
```diff
@@ -286,7 +286,7 @@ void SLNGenerator::WriteProjectConfigurationPlatforms( const Array< AString > &
     for( const AString * it = projectGuids.Begin() ; it != projectGuidsEnd ; ++it )
     {
         // only one project active in the solution build
-        const bool projectIsActive = solutionBuildProjectGuids.Find( *it );
+        const bool projectIsActive = solutionBuildProjectGuids.Find( *it ) != nullptr;
```
gkoreman: This wasn't supposed to be in this PR, but I added it because it fixes a warning

ffulin:
Thanks for the pull request. Changing the way Compiler works so that you can build compilers and use them in subsequent steps via Compiler in the same way you can with Exec is a great change. As you noted, it's a non-trivial one, so it'll take a little time to analyse the pull request thoroughly.

ffulin:
I've integrated your changes. There were a few problem which I needed to address (which I've done in https://github.com/fastbuild/fastbuild/commit/27d2f4e5095d9bb20465929ba6dc9d646daea36b). This change will be in the next release (v0.96). Thanks again!

PR #414, opened 2018-04-16
Added support for Android standalone toolchain clang compiler (windows)

cyberkm:
Installing  android standalone toolchain for windows requires to use compiler wrappers with .cmd
extensions, fastbuild didn't recognize them properly as clang and failed with error 1501.


ffulin:
I've integrated your change into the mainline. This will be in the next release (v0.96). Thanks for the pull request!

PR #417, opened 2018-04-18
Handle failure to skip a string in BFFIterator::ParseToMatchingBrace

dummyunit:
When `SkipString` fails to find the end of a string it leaves `m_Pos` pointing at the last character of iterator range. When we are parsing some inner scope that last character might be a closing brace. In that case `ParseToMatchingBrace` will be tricked into thinking that matching closing brace is found, when in reality that brace is a part of unterminated string.

Later this breaks logic in places that expect all strings inside a region parsed by `ParseToMatchingBrace` to be terminated.
For example in `StoreVariableArray` ([BFFParser.cpp:1468](https://github.com/fastbuild/fastbuild/blob/b76a1920b91d9ca65a37becabe6af79581a888ee/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp#L1468)), here iterator is incremented past the end and in Release build it ultimately leads to infinite loop in `FormatError`.

That infinite loop was found by BFFFuzzer, minimal reproducer is:
```
{#ifB'
#endif
.a={'}
```

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFIterator.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFIterator.cpp
@@ -193,7 +193,12 @@ bool BFFIterator::ParseToMatchingBrace( char openBrace, char closeBrace )
         // hit a string?
         if ( ( *m_Pos == '\'' ) || ( *m_Pos == '"' ) )
         {
+            const char quote = *m_Pos;
             SkipString( *m_Pos );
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the fix and the repro case! This fix will be in the next release (v0.96). I went and added a regression test as well (https://github.com/fastbuild/fastbuild/commit/bf18b9a3e9402b8e0212ffb779022652c3ef3942)

PR #418, opened 2018-04-18
Update build instructions for BFFFuzzer and MSan configuration

dummyunit:
* Configuration for MSan build updated to link libc++ statically. This allows to run MSan builds of BFFFuzzer without specifying `LD_LIBRARY_PATH`.
* Instead of using a different name for libFuzzer (and libc++) static library built with MSan it is now expected to be placed in the `External/MSan/` subdirectory inside the repo.
* Fixed instructions for building libc++ with MSan instrumentation: added missing options to enable use of libc++abi as runtime support library.
* Updated and verified compiler and linker options to enable libFuzzer instrumentation in Clang 6.0.0.
* New operators of the If() function introduced in b76a1920 added to the dictionary file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # FASTBuild
 External/SDK
+External/MSan
 *.exe
 *.fdb
 *.log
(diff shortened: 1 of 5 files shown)


ffulin:
Thanks for the updates! I've integrated this into the mainline.

PR #419, opened 2018-04-20
Fix crash when struct variable is replaced by its member in Using()

dummyunit:
When we are importing members of a struct into current scope via function `Using()` we have to ensure that struct won't get deleted while we iterating its members. This may happen if a struct has a member with the same name as the struct itself.

This issue was found by BFFFuzzer, minimal reproducer is:
```
.A=[.A=[]]Using(.A)
```


ffulin:
Thanks for the fix! This fix will be in the next release (v0.96).

I added a regression test in https://github.com/fastbuild/fastbuild/commit/181aa2a3ac477129cec0be596dd8e3beb76f847d.

PR #421, opened 2018-04-21
Allow a variable to be frozen multiple times

dummyunit:
This effectively legalizes nested `ForEach` loops over the same array.
Previously there was an `ASSERT` that prevented such loops, but `ASSERT` works only Debug builds and in Release such loops worked just fine in most cases.

Issues were appearing only if the BFF code was trying to modify the array after the end of a nested loop. In that case the array, which was marked as not frozen at that point, was replaced without producing a visible error. And later `FunctionForEach::ParseFunction` was accessing already deleted variable on the second iteration.

This issue was found by BFFFuzzer, minimal(-ish) reproducer is:
```
.a={'' 'X'}
ForEach(.x in .a) {
    ForEach(.y in .a) {}
    ^a={}
}
```
Unfortunately this reproducer doesn't trigger a crash in Release build (only diagnostic output in ASan build), so it probably not so useful as a test case for FBuildTest.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFVariable.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFVariable.cpp
@@ -29,7 +29,7 @@
 BFFVariable::BFFVariable( const AString & name, VarType type )
 : m_Name( name )
 , m_Type( type )
-, m_Frozen( false )
+, m_FreezeCount( 0 )
 , m_BoolValue( false )
 , m_IntValue( 0 )
 //, m_StringValue() // default construct this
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for the bug report. It was always intended that you could nest the for loops, but being able to modify the variable being looped due to the lock being binary instead of ref counted is definitely a bug.

ffulin:
Thanks for the fix. I've integrated the pull request and the fix will be in the next release (v0.96). I went and added a test (https://github.com/fastbuild/fastbuild/commit/81091f77699a60c771d4ccb05df7c2b2adedc945). I also changed the m_FreezeCount in order to avoid increasing the size of the BFFVariable class (to avoid any performance impact to bff parsing).

PR #423, opened 2018-04-30
 FileIO is symlink aware.

rickvanprim:


Code/Core/FileIO/FileIO.cpp
```diff
@@ -157,10 +159,13 @@
...
-    s = copyfile_state_alloc();
-    bool result = ( copyfile( srcFileName, dstFileName, s, COPYFILE_DATA | COPYFILE_XATTR ) == 0 );
-    copyfile_state_free(s);
+    /*
+     If the state parameter is the
+     return value from copyfile_state_alloc(), then copyfile() and fcopyfile() will use the information from
+     the state object; if it is NULL, then both functions will work normally, but less control will be
+     available to the caller.
+    */
+    bool result = ( copyfile( srcFileName, dstFileName, nullptr, COPYFILE_DATA | COPYFILE_XATTR | COPYFILE_NOFOLLOW ) == 0 );
```
rickvanprim: If you want this change (or the quoted documentation) removed, let me know.

Code/Core/FileIO/FileIO.cpp
```diff
@@ -785,6 +803,14 @@
         FindClose( hFind );
 
     #elif defined( __LINUX__ ) || defined( __APPLE__ )
+        // Special case symlinks.
+        struct stat stat_source;
+        VERIFY( lstat( pathCopy.Get(), &stat_source ) == 0 );
+        if ( S_ISLNK( stat_source.st_mode ) )
+        {
+            return;
```
rickvanprim: Per the Windows behavior, `FindFileEx()` won't traverse a symlink, where as I'm 99% sure `opendir()` will, so we want to early out in that case for consistent behavior.

ffulin:
I've integrated this into the mainline. This will be in the next release (v0.96). Thanks for the pull request!

PR #424, opened 2018-05-01
Correctly handle closing brace right after the end of iterator range

dummyunit:
Recent fix for brace parsing (c253336d) addressed only one source of incorrect parsing: unterminated strings, but the same incorrect behaviour might be triggered by a comment at the end of iterator range.
Such case was found by BFFFuzzer:
```
{#ifB'
#endif
{;'}
```

The real problem in all cases is that #if directive can change meaning of a text by conditionally wrapping it in strings an that `ParseToMatchingBrace` ignores `#if` directives and parses bodies unconditionally.
While this problem is hard to fix, `ParseToMatchingBrace` can be fixed to at least not return true when iterator has reached the end of its range.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFIterator.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFIterator.cpp
@@ -199,12 +199,13 @@ bool BFFIterator::ParseToMatchingBrace( char openBrace, char closeBrace )
             {
                 return false;
             }
+            continue;
         }
 
         // hit the close brace?
         if ( *m_Pos == closeBrace )
         {
-            return true;
+            return !IsAtEnd();
         }
 
         // a regular charater.... keep searching


ffulin:
Thanks for the fix. This will be in the next release (v0.96). I added a regression test as well (https://github.com/fastbuild/fastbuild/commit/292469aadda2e9e2960aa3dd153e643d66a09eea).

As you noted, the parsing is tricky, and in retrospect having a one-pass parser was probably a mistake as it significantly complicates parsing to mix directives and other logic together (as well as creating some unintuitive behaviour in some cases). In the future, I might re-do the parsing which would allow for some more consistent handling of things like this as well as allowing me to implement some global improvements to error reporting and also to improve performance.  For the short to medium term, I think patching the holes is the right way to go. All the edge cases you've caught and the tests we've added will be great if/when tidying up the parsing in the future.

PR #426, opened 2018-05-01
Build only Debug and Release configurations in TestBuildFBuild

dummyunit:
This change speeds up `BuildFBuild` tests by minimizing amount of build configurations while maintaining almost the same amount of code exposed to compilers. Resulting speedup is especially significant on Linux (from 459 seconds down to 109 seconds in my case).

1. Sanitizer configurations are excluded because they have the same defines as Release configuration. Also just building code with sanitizers doesn't uncover any bugs by itself.
2. Profile configuration has a different set of defines than Debug and Release configurations but almost all code is already covered by them. (There are only trivial code behind `defined( RELEASE ) &&
   defined( PROFILING_ENABLED )` condition.)

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestBuildFBuild.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestBuildFBuild.cpp
@@ -52,9 +52,25 @@ FBuildStats TestBuildFBuild::BuildInternal( FBuildTestOptions options, bool useD
     FBuild fBuild( options );
     TEST_ASSERT( fBuild.Initialize( useDB ? GetDBFile() : nullptr ) );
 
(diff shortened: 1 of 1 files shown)


ffulin:
This change makes sense - thanks!

I realize that if you are pulling the repo from git, then because all the files are writeable all the files will be isolated from Unity, so you'd be getting *really* slow compile times:

| | Isolated | Unity |
| Before | 812s | 93s |
| With This Pull Request | 564s | 68s |

I've made a change to the bff (https://github.com/fastbuild/fastbuild/commit/8c792e558730c55e04250f47dfe45214c0210882) so that when FASTBuild is compiling itself, it will be done entirely in Unity, so now we get the faster time (68s above) in all cases

PR #427, opened 2018-05-01
Don't allow to replace nodes with CopyDir/RemoveDir nodes

dummyunit:
During conversion of CopyDir and RemoveDir to reflection system (9e6c46e3 and bde40a70) checks for uniqueness of the node name were lost.


ffulin:

PR #428, opened 2018-05-06
Support colored output from Clang and GCC 

dummyunit:
Both Clang and GCC (since 4.9) have support for colored diagnostic output. Also both by default enable colored output when their stderr is not redirected. This works well with other build drivers like `make` and `ninja` because they don't intercept command output.

To make colored output work with FASTBuild we need to force it by passing `-fdiagnostics-color=always` option to Clang/GCC when FASTBuild output is not redirected.
Because not all GCC versions support `-fdiagnostics-color` options we add that option only if `-fdiagnostics-color=auto` (which is the default) is explicitly specified in `.CompilerOptions`.
Also we don't add that option if colored output is explicitly forced to disabled or enabled in `.CompilerOptions`.

The second commit adds caching for the result of `Env::IsStdOutRedirected` because now it can potentially be called 1-2 times for each ObjectNode and that check is expensive on Windows.

### DIFF
--- a/Code/Core/Env/Env.cpp
+++ b/Code/Core/Env/Env.cpp
@@ -229,9 +229,9 @@ void Env::GetExePath( AString & output )
     #endif
 }
 
-// IsStdOutRedirected
+// IsStdOutRedirectedInternal
 //------------------------------------------------------------------------------
-/*static*/ bool Env::IsStdOutRedirected()
+static bool IsStdOutRedirectedInternal()
 {
     #if defined( __WINDOWS__ )
         HANDLE h = GetStdHandle( STD_OUTPUT_HANDLE );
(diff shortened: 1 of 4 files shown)


ffulin:
Thanks for the changes.

I wonder if the change to add the -fdiagnostics-color=always is really necessary. i.e. since auto it on by default, and the user has to specify auto manually to make it work, why not simply let the user specify always and have FASTBuild not modify the args at all?

dummyunit:
> user has to specify auto manually to make it work

Only for GCC, this is only because `-fdiagnostics-color` family of options is supported only since GCC 4.9. And if FASTBuild would add `-fdiagnostics-color=always` when GCC <= 4.8 is used then user would get a compilation error about unrecognized command line option.
For Clang, which supported `-fdiagnostics-color` since the first release, colored output will work without any modifications to `.CompilerOptions`.

> why not simply let the user specify always and have FASTBuild not modify the args at all?

Then the user will get color output even when FASTBuild's output is redirected, if that is what the user wants then adding `-fdiagnostics-color=always` to `.CompilerOptions` is the right move (for the user).

But, if user wants behavior of `-fdiagnostics-color=auto` (colored output only if output goes to the terminal), then FASTBuild is have to help the compiler (because from compiler point of view output is always redirected, so no colors) by forcing colored output when it is needed.

ffulin:
Thanks for the fix/improvement! This has been integrated and will be in the next release (v0.98).

PR #430, opened 2018-05-09
Linux and gcc adaptations

szubersk:
Few small adaptations for Linux:

ability to compile the project with g++ 7
FASTBUILD_TEMP_PATH is respected on Linux systems
gcc/g++ requires preprocessed files to have *.i extension to operate correctly
gcc compiler type will be detected if binary name is in a form "gcc-${version}" e. g. gcc-7

### DIFF
--- a/Code/Core/Mem/MemDebug.cpp
+++ b/Code/Core/Mem/MemDebug.cpp
@@ -9,10 +9,6 @@
 
 #if defined( MEM_DEBUG_ENABLED )
 
-#if defined( __GNUC__ ) && __GNUC__ >= 7
-    #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
-#endif
-
 // Core
 #include "Core/Env/Assert.h"
 #include "Core/Env/Types.h"
@@ -44,12 +40,20 @@ void MemDebug::FillMem( void * ptr, const size_t size, const uint32_t pattern )
                             (char)( ( pattern & 0x00FF0000 ) >> 16 ) };
         const char * b = bytes;
         char * cit = static_cast< char * >( static_cast< void * >( it ) );
+
+        #if defined( __GNUC__ ) && __GNUC__ >= 7
+            #pragma GCC diagnostic push
+            #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+        #endif
         switch( remainder )
         {
             case 3: *cit = *b; ++cit; ++b;
             case 2: *cit = *b; ++cit; ++b;
             case 1: *cit = *b;
         }
+        #if defined( __GNUC__ ) && __GNUC__ >= 7
+            #pragma GCC diagnostic pop
+        #endif
     }
 }
 
--- a/Code/Core/Mem/MemDebug.cpp
+++ b/Code/Core/Mem/MemDebug.cpp
@@ -41,19 +41,12 @@ void MemDebug::FillMem( void * ptr, const size_t size, const uint32_t pattern )
         const char * b = bytes;
         char * cit = static_cast< char * >( static_cast< void * >( it ) );
 
-        #if defined( __GNUC__ ) && __GNUC__ >= 7
-            #pragma GCC diagnostic push
-            #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
-        #endif
         switch( remainder )
         {
             case 3: *cit = *b; ++cit; ++b;
             case 2: *cit = *b; ++cit; ++b;
             case 1: *cit = *b;
         }
-        #if defined( __GNUC__ ) && __GNUC__ >= 7
-            #pragma GCC diagnostic pop
-        #endif
     }
 }
 
--- a/Code/Core/Mem/MemDebug.cpp
+++ b/Code/Core/Mem/MemDebug.cpp
@@ -40,7 +40,6 @@ void MemDebug::FillMem( void * ptr, const size_t size, const uint32_t pattern )
                             (char)( ( pattern & 0x00FF0000 ) >> 16 ) };
         const char * b = bytes;
         char * cit = static_cast< char * >( static_cast< void * >( it ) );
-
         switch( remainder )
         {
             case 3: *cit = *b; ++cit; ++b;
--- a/Code/Core/Mem/MemDebug.cpp
+++ b/Code/Core/Mem/MemDebug.cpp
@@ -9,6 +9,10 @@
 
 #if defined( MEM_DEBUG_ENABLED )
 
+#if defined( __GNUC__ ) && __GNUC__ >= 7
+    #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+#endif
+
 // Core
 #include "Core/Env/Assert.h"
 #include "Core/Env/Types.h"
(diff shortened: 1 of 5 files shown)


szubersk:
@ffulin Please let me know what you think about changes above

Code/Core/Mem/MemDebug.cpp
```diff
@@ -9,6 +9,10 @@
 
 #if defined( MEM_DEBUG_ENABLED )
 
+#if defined( __GNUC__ ) && __GNUC__ >= 7
+    #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+#endif
+
```
ffulin: I think we might be better off just changing the one switch statement the flags this so that we don't need to turn it off, potentially covering up other instances of this warning in this cpp and other cpp files in the library when it is built in Unity.

Code/Tools/FBuild/FBuildCore/FBuild.cpp
```diff
@@ -742,28 +742,32 @@ bool FBuild::DisplayDependencyDB( const Array< AString > & targets ) const
 //------------------------------------------------------------------------------
 /*static*/ bool FBuild::GetTempDir( AString & outTempDir )
 {
-    #if defined( __WINDOWS__ )
+    #if defined( __WINDOWS__ ) || defined( __LINUX__ )
```
ffulin: This is a good fix. It looks like we can actually apply the same logic on OSX as well.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -2021,6 +2021,10 @@ bool ObjectNode::WriteTmpFile( Job * job, AString & tmpDirectory, AString & tmpF
 
     FileStream tmpFile;
     AStackString<> fileName( sourceFile->GetName().FindLast( NATIVE_SLASH ) + 1 );
+    if (GetFlag( FLAG_GCC ))
+    {
+       fileName += ".i";
+    }
```
ffulin: Can you tell me a bit about what problems you had with this?  FASTBuild is built with gcc on Linux and I've not encountered this problem (though i have seen this in the past). It would be good to ensure we have test coverage for any case that requires this.

Code/fbuild.bff
```diff
@@ -38,11 +38,11 @@ Compiler( 'Compiler-x64Clang' )
 }
 Compiler( 'Compiler-x64-Linux' )
 {
```
ffulin: In the future, I'd like to extend the config files for Linux so that you can easily switch between versions, but for now, given that it's not going to be the right version for everyone, I think I'd prefer to keep this at v4.9 since that is the version I use to make official releases.

ffulin:
Thanks for the changes. I've added some comments to the various files.

Code/Core/Mem/MemDebug.cpp
szubersk: Corrected. I'll upload next iteration soon

Code/Tools/FBuild/FBuildCore/FBuild.cpp
szubersk: Applied for __APPLE__ too. I think __APPLE__ is more reliable define than __OSX__.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
szubersk: The problem here is strictly connected to g++ being unable to distinguish between cpp file and preprocessed cpp file when file extension is .cpp. If it's the case then some defines automagically inserted by g++ at the beginning of the file (compiler defaults) are mistreated as regular macro definitions causing compilation to fail. The fix is to help compiler detect file content by adding proper extension (*.i). Seems like g++ 7 on my laptop is unaffected by this but on my friends machine it was a clear error (g++ 5.3.x if I remember correctly). I don't think this can be covered by unit testing though. The change is not harmful in any way and it works perfectly on g++ 7 and 5.3.x.

Code/fbuild.bff
szubersk: Noted on g++ version. The binary name needs to be changed though because invocation of gcc implies good-old C code in the file, not C++ code which causes compilation errors. As with *.i extension - some gcc version might or might not detect the proper language based on file content. It's safer to clearly force C++ mode - saves nerves.

szubersk:
@ffulin Thank you very much for reviewing my changes. I uploaded newest revision addressing your concerns.

Code/fbuild.bff
```diff
@@ -38,7 +38,7 @@ Compiler( 'Compiler-x64Clang' )
...
-    .Executable     = '/usr/bin/gcc'
+    .Executable     = '/usr/bin/g++'
```
dummyunit: This change will result in .c files being compiled as C++. In case of LZ4/xxHash (which are bundled with FASTBuild) this doesn't lead to new errors or warnings, but it might change code generation (for better or worse).

dummyunit:
> Failed due to missing file: /home/szubersk/projects/fastbuild/Code/BFFFuzzer-x64Linux-Debug

BFFFuzzer can be compiled only in 2 configurations, but there in no way (in current configs) to exclude a specific project from `All-*` aliases in some configurations.

This error is my fault essentially, I've added BFFFuzzer, but still didn't find a nice solution for this error. In my personal branch I currently use an ugly workaround for it: https://github.com/dummyunit/fastbuild/commit/b3116df72c9ea9b0359d010a4793d5670df2359e#diff-dcd73127757265333744a09bdb3ef30bR57 I don't think that it is acceptable for inclusion if official sources though.

Code/fbuild.bff
szubersk: Is there a way to define 2 different compilers, one for C and the other one for C++ code?

ffulin: It is possible to define two compilers by having two Compiler() sections in the bff. However, a simpler way is to keep the same compiler and to specify the language on the command line with the -x option.

szubersk: I reverted this change and gcc-7 and gcc-8 seem to work OK. No point in forcing g++.

szubersk:
@dummyunit 
@ffulin 
I addressed all comments and either corrected or reverted changes.

Code/Core/Mem/MemDebug.cpp
```diff
@@ -40,12 +40,20 @@ void MemDebug::FillMem( void * ptr, const size_t size, const uint32_t pattern )
                             (char)( ( pattern & 0x00FF0000 ) >> 16 ) };
         const char * b = bytes;
         char * cit = static_cast< char * >( static_cast< void * >( it ) );
+
+        #if defined( __GNUC__ ) && __GNUC__ >= 7
+            #pragma GCC diagnostic push
+            #pragma GCC diagnostic ignored "-Wimplicit-fallthrough"
+        #endif
```
ffulin: This switch statement has now been changed so this warning disabling is no longer required: https://github.com/fastbuild/fastbuild/commit/68e5e7dc88774d63dd6c81f67b2ae64b7570f565

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
ffulin: I am using gcc 4.9.2 and am not seeing this problem. I wonder if it's how gcc is being invoked that is the problem - in all my cases, i'm compiling with: ```/<path>/gcc -std=c++11``` Can you tell me what args you are using, and if possible if this makes a difference on your version(s)?

If this is the issue, it would be nice to make this just work with a better preprocessed file extension along the lines of what you have proposed, but I think we need to detect .c vs .cpp and use .i or .ii respectively to be correct.

ffulin:
I've integrated these changes. They will be part of the next release (v0.98). Thanks for the fixes!

Issue #351, opened 2017-12-15
Create compilation database JSON


Leandros:
To integrate clang tooling into projects, you need to pass all preprocessor definition flags and include directories to each tool. Having a generated compilation database from your build system can dramatically reduce the effort required. 

The compilation database is, essentially, a json file which contains the path to the file compiled and the compiler flags. The full specification can be found here:

https://clang.llvm.org/docs/JSONCompilationDatabase.html

ffulin:
I'm not familiar with the use-case for this file, but I took a look at the documentation you linked.

The file itself looks like something which FASTBuild could generate without too much trouble. I imagine we'd add a new Function, such that this could work similarly to Visual Studio or XCode project generation.

It should be noted that if any external tooling wants to invoke any tooling by using these files, it would need to do so in a way that does not modify the actual built files that FASTBuild has generated to ensure it does not cause any problems like unnecessary rebuilds.

Leandros:
The compilation database is mainly used for code completion tools (like [YouCompleteMe](https://github.com/Valloric/YouCompleteMe) to give you code suggestions inside Vim, or the same for Emacs) or clangs tooling (like clang-tidy, to run static analysis). 
The chances that any of those tools is changing any generated files is very low.

ffulin:
A compilation database was added some time ago (in v0.98, Apr 2019) but I missed resolving this issue.

Issue #193, opened 2017-01-13
Output compilation commands in compile_commands.json


Ryp:
There are tools to get the compile commands out of fastbuild, like the `-showcmds` or external tools that substitute as a compiler to get the flags used by each source file, but nothing is really built in to do that.

I think this could be a great addition, being used to CMake's `set(CMAKE_EXPORT_COMPILE_COMMANDS ON)` that does exactly that, and is compatible with some clang-based tools.

ffulin:
Hello.  This would certainly be possible to add as a feature, although some build steps (those managed internally by FASTBuild) would not be captured. For example, the generation of project files or unity files does not invoke any external tools. Perhaps that's fine, depending on your use-case.

Ryp:
Usually this is useful for knowing what command was used to compile which source file, I don't think generation of sln's or unity objects really fall into that use case, but I could be wrong (or at least this wouldn't be very useful to me). That's what I'm interpreting form http://clang.llvm.org/docs/JSONCompilationDatabase.html

ffulin:
Support for the compile_commands.json was added in https://github.com/fastbuild/fastbuild/pull/436 and integrated. Version 0.98 contains this new feature (I forgot to close this issue when that version was released).

PR #436, opened 2018-05-27
Add option to generate JSON Compilation Database

dummyunit:
This PR adds new option `-compdb` that instructs FBuild to generate [JSON Compilation Database](https://clang.llvm.org/docs/JSONCompilationDatabase.html) for the targets specified on the command line instead of building them.

This was requested in #193 and #351. Some points regarding the discussion in these threads:
* Compilation database is a (semi-)standard way to provide tools with compiler options that are required to _correctly parse_ source files. It is not required that commands in the compilation database should match the exact commands used to build actual object files. I don't know of any tools that will actually try to execute commands listed in the database. Given that, it is OK to include only compile commands for the files compiled via ObjectList and Library and in the most basic form (so no separate preprocesing step and no entries for generated unity files).
* I decided to implement generation as a command line option instead of a new BFF Function, for the following reasons:
  1. With an option approach the new version of FASTBuild can be used with old config files that predate support for generation of compilation databases (and thus can't have the new function).
  1. Including new BFF function to generate compilation database in config file would automatically make it incompatible with the old versions of FASTBuild, I could imagine that in some cases that would be undesirable.
  1. With a BFF function approach list of targets would have to be hard coded, with an option approach desired targets can be specified dynamically on the command line. This is a nice feature because some tools (e.g. `run-clang-tidy.py`) can perform action for each file included in the database and in that case it is useful to have only interesting files in the database.

Also here is a nice list of current consumers of compilation database files: https://sarcasm.github.io/notes/dev/compilation-database.html#what-is-it-good-for

Generation is performed by recursively traversing all dependencies of the specified targets and:
* Building all DirectoryList nodes to generate lists of input files.
* Emitting one entry for each input file that is included in ObjectList and Library nodes. When Unity node is used as an input, DB entries are generated for all its input files instead of generated unity files.

Generated DB uses newer format (split arguments in the `"arguments"` array instead of raw command line in the `"command"` string) to avoid issues with different rules for argument quoting and escaping in Clang and FASTBuild (especially on Windows). The downside is that generated compilation database will not work with tools from clang 3.9 and below.


Code/Core/Strings/AString.h
```diff
@@ -92,6 +92,8 @@ class AString
     uint32_t Replace( const char * from, const char * to, uint32_t maxReplaces = 0 );
     void ToLower();
     void ToUpper();
+    void JSONEscape();
+    void Unquote();
```
dummyunit: Are these functions generic enough to be in the `AString` header, or should I move them to `CompilationDatabase` class?

dummyunit:
Rebased PR onto current dev branch to accomodate changes made to the CompilerNode (executable name is accessed via `GetExecutable()` instead of `GetName()`).

Code/Tools/FBuild/FBuildCore/Helpers/CompilationDatabase.cpp
```diff
@@ -0,0 +1,215 @@
...
+        VisitNodes( node->GetPreBuildDependencies(), visited );
+        VisitNodes( node->GetStaticDependencies(), visited );
+        VisitNodes( node->GetDynamicDependencies(), visited );
+
+        switch ( node->GetType() )
+        {
+            case Node::DIRECTORY_LIST_NODE:
+            {
+                // Build directory list node to populate its file list
+                node->CastTo< DirectoryListNode >()->DoBuild( nullptr );
```
ffulin: DirectoryListNodes can have PreBuildDependencies, which won't be correctly traversed by this manual invocation of the build.

Code/Tools/FBuild/FBuildCore/Helpers/CompilationDatabase.cpp
```diff
@@ -0,0 +1,215 @@
...
+    m_Output += m_DirectoryEscaped;
+    m_Output += "\",\n    \"file\": \"";
+    m_Output += inputFileEscaped;
+    m_Output += "\",\n    \"output\": \"";
+    m_Output += outputFileEscaped;
+    m_Output += "\",\n    \"arguments\": [\"";
+    m_Output += ctx->m_CompilerEscaped;
+    m_Output += "\"";
+    for ( const AString & argument : ctx->m_ArgumentsEscaped )
+    {
```
ffulin: This code seems fragile. Any changes to how args are built for the compiler won't be reflected here. It might be that anything not currently handled here isn't important, but this is sure to get broken if the %1 syntax is expanded in the future.

Code/Core/Strings/AString.cpp
```diff
@@ -580,6 +580,215 @@ void AString::ToUpper()
...
+    if ( extraSpace == 0 )
+    {
+        return; // No characters need escaping
+    }
+
+    if ( m_Length + extraSpace > GetReserved() )
+    {
+        Grow( m_Length + extraSpace );
+    }
+
```
ffulin: I think this function would be much simpler if it built a temporary buffer for the output and swapped the strings. That would allow the loop to go forwards. As is, the grow already causes a temporary copy to occur, so it wouldn't be any more costly.

Code/Core/Strings/AString.h
ffulin: As they only have one user, I think moving them into the compilation database would make sense.
That would also allow the JSONEscape to be greatly simplified, since it would no longer need to handle most of what it currently does, because we know the use is contstrained.

ffulin:
This looks like a good feature, and the inclusion of tests is much appreciated. My apologies for taking so long to go through the PR.

I'm a little worried about a couple of things:

*a) It bypasses the dependency graph and re-implements a partial traversal.* This means the output file will be incorrect in some situations (where there is codegen for example). This also seems like it will also increase the cost of any changes to the graph representation, or the addition of new node types as it will require this new traversal to be considered and updated.

*b) It bypasses command line manipulation* This looks like something that will either be broken by future chanegs to command line building, or will need updating if those things change (requiring things to be implemented twice). As an example, a more flexible or new new build-time replacement(s) will require this code be considerd and updated.

I think the graph traversal problem is a tricky one, but some ideas we could talk about inlude:
 - generating this file post-build, after normal dependency traversal has occurred (walk again after the build, but can assume graph is complete/correct so avoid DoBuild issues)
 - generating this file during the traversal of a normal build (avoid having to implement another walk and avoid DoBuild issues)

For the args, I think perhaps the BuildArgs could be used to retrieve the needed arguments, including all special manipulation. It might currently have some dependencies that are problematic (like on "job") but they are probably fixable.

Let me know what you think.

Code/Core/Strings/AString.h
dummyunit: OK, I will move them.
But I don't think that `JSONEscape` can be simplified much, because:
1. It still has to be able to escape all characters that need escaping (they can appear inside defines in `.CompilerOptions`), so the giant switch won't get any smaller.
2. It is still beneficial to calculate `extraSpace` before modifying anything because most of the time there will be no characters to escape. Here are some statistics for the FASTBuild repo:
On Linux running `fbuild -compdb All-x64Linux-Debug` results in 615 calls to `JSONEscape` and all of them are no-op (exit via `extraSpace == 0` condition).
On Windows running `FBuild -compdb All-x64-Debug` results in 794 calls to `JSONEscape`, 371 of them are no-op and only 10 call `Grow`. And if `/` would have been used as a path separator in Windows parts of bff files then there would be even more no-op cases.

One thing that can be changed is the direction of the second loop. This, I guess, will make it simpler but it will also pessimize cases in which there are enough reserved space to hold the modified string (which, as the stats above show, is the majority of cases).
So, should I change the loop direction?

Code/Tools/FBuild/FBuildCore/Helpers/CompilationDatabase.cpp
dummyunit: Yes, this is a problem. But I think it is not a severe problem because:
1. Most source files are written manually and not generated, so actual impact will be low.
2. Due to ways in which compilation databases are used (providing includes/defines to static analysis/refactoring tools and IDEs) they are much less useful for generated files.
3. If the file generated by a pre-build dependency doesn't exist then the corresponding entry in the compilation database is useless, but if the file exists then it will be found by the DirectoryNode even without building pre-build dependencies.
4. If somebody really wants files generated by pre-build dependencies to be present in the generated compilation database all he/she has to do is to as FASTBuild to execute that part of the build just once (not the whole build, and not each time `-compdb` is used).

So to me it seems that there is little value in solving this problem.
I think issuing a warning like "Skipping pre-build dependency NAME, files generated by it may or may not be present in the generated compilation database." when we encounter ObjectListNode, LibraryNode or UnityNode with a pre-build dependency will be enough.

Code/Tools/FBuild/FBuildCore/Helpers/CompilationDatabase.cpp
dummyunit: As I wrote in the original post, compilation database doesn't have to have the same exact arguments that are used during the actual build, it has to have some arguments that allow to correctly parse the sources.

So almost all of the argument manipulations that `ObjectNode::BuildArgs` does are not wanted here, arguments specified explicitly in `.CompilerOptions` will be sufficient.

The only exception is expansion of `%N` tokens, but while the logic for it is the same (replace `%N` with some string) the actual code is completely different due to different argument storage (`AString` vs `Args`). So I don't think that this part can be easily refactored to be used from both places.

dummyunit:
> It bypasses the dependency graph and re-implements a partial traversal.

I must not understand what do you mean by "bypasses the dependency graph", the dependency graph as a data structure is used: `FBuild::GetTarget` builds list of initial vertices (nodes) by finding them in the graph (`FindNodeExact`), and `CompilationDatabase::VisitNodes` does a depth first scan by getting a list of edges from each vertex (`GetPreBuildDependencies`, `GetStaticDependencies` and `GetDynamicDependencies`).

The traversal is re-implemented (or rather just copied from `NodeGraph::DisplayRecurse`) because:
1. The way in which the graph is traversed during the build is convoluted and heavily tied to the build process so it can't reused.
2. The traversal logic in other places (`NodeGraph::DisplayRecurse`, `NodeGraph::SaveRecurse`) is trivial but has many customization points.

> it will also increase the cost of any changes to the graph representation

As long as there will be a way to get all edges (dependencies) of a vertex (node) the required changes should be trivial.

> addition of new node types as it will require this new traversal to be considered and updated

Considered - yes. Updated - most likely no, unless the new node will be involved in compiling C/C++ code.

> It bypasses command line manipulation

Yes, it does. Intentionally. The user has already supplied the correct arguments in `.CompilerOptions` so no manipulation besides expanding %1 and %2 is required. In theory expansion on `%N` tokens could be unified by extracting it into a separate function but I don't see how it can be done right now.

> I think the graph traversal problem is a tricky one, but some ideas we could talk about inlude:

A I wrote above I don't think the problem with pre-build dependencies and generated files is worth solving.
Also both these solutions would force a full build (of specified targets) upon everybody, whether their case need it or not. This will be very detrimental in many cases, for example: A dedicated CI task for running static analysis on the code will now have to build the code too.

ffulin:
This change has been integrated into the mainline and will be in the next release (v0.98).

arbv:
Thank you!

PR #439, opened 2018-06-18
Limit number of files in TestProjectGenerator.TestFunction_Speed to 5000

dummyunit:
Logic for excluding duplicate files in `VSProjectGenerator` has O(n^2) complexity. This leads to apparent hanging of `TestFunction_Speed` if the user has large amount of extra files in the `Code` directory. In my case I had several copies of corpus for BFFFuzzer inside `Code` directory, about 90000 files in total.

To prevent confusion (it's not obvious what is the cause of hang) limit number of files passed to `VSProjectGenerator::AddFiles` to 5000, that many files can be handled in a reasonable time.

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestProjectGeneration.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestProjectGeneration.cpp
@@ -261,6 +261,10 @@ void TestProjectGeneration::TestFunction_Speed() const
     // files (about 5,000)
     Array< AString > files;
     FileIO::GetFiles( baseDir, AStackString<>( "*" ), true, &files );
+    if ( files.GetSize() > 5000 )
+    {
+        files.SetSize( 5000 );
+    }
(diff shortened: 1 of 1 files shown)


ffulin:
I went ahead and fixed the performance regression (https://github.com/fastbuild/fastbuild/commit/03a7d87362dad5f69f2fffde8461faf7888fa28f). The performance is now almost the same as before in Release, and is significantly faster in Debug builds. I stressed it out with 90,000 files:

| | Old | Current | New |
| --- | --- | --- | --- |
| Debug | 66.6s | > 10 minutes* | 7.1s |
| Release | 3.75s | 448s | 4.0s |

\* I didn't bother waiting for it to finish :)

Thanks for letting me know about the regression. I'll leave this pull request open until the next version is released, so v0.96 will remain overall about the same as before, but with the de-duplication in-tact.

ffulin:
A version containing this fix has now been released (v0.96).

PR #441, opened 2018-06-18
Fix assertion when FileIO::GetFiles functions are called on non-existent directory

dummyunit:
After symlink-related changes in e12a8d2a `GetFiles` functions started to fail with assertion when called on non-existent directory. But calling them on non-existent directory should be allowed (at least from looking at the older version of code).
To fix that don't expect `lstat` call on original `pathCopy` to succeed.

This fixes `TestBuildFBuild` and `TestDistributed` tests in the case when `../tmp/Test` directory is missing.


ffulin:
It should not assert when calling GetFiles, so this is a regression. Thanks for the fix!

ffulin:
Thanks! This fix will be in the next source code release (v0.96)

PR #442, opened 2018-06-18
Fix assertion when referencing variable with dynamic name from parent scope

dummyunit:
An if block at the end of `BFFParser::ParseVariableName` that sets first symbol to `.` is moved (with both related assertions) into block for handling variables with a fixed name.
Block that handles variables with dynamic names already produces names starting with `.` regardless of `parentScope` value, so assertion in the moved block was triggering when it shouldn't be.


ffulin:
Thanks!

PR #444, opened 2018-06-22
Refactor generation of All-* aliases

dummyunit:
This change allows a project to exclude itself from `All-$Platform$-$Config$` alias for a configuration in which it can't be built. This is used by BFFFuzzer which builds only in `x64ClangLinux-ASan` and `x64ClangLinux-MSan`.
Without this change building (for example) alias `All-x64Linux-Debug` fails with an error that says that file `Code/BFFFuzzer-x64Linux-Debug` is missing. Also because TestBuildFBuild uses such aliases it was also broken.

Additionally this change removes `All-$Platform$-$Config$` aliases which can't be built on the current platform, e.g. `All-x64Linux-Debug` won't be defined on Windows anymore.


ffulin:
Thanks!

PR #445, opened 2018-06-23
Fix TestPCH with clang on windows

belkiss:
Hello!

Here is a fix for an error I got when running the tests on the dev branch:

```
 - Test 'TestPCH'
E:\RD\FastBuild\Code\Tools\FBuild\FBuildTest\Data\TestPrecompiledHeaders\fbuild.bff(64,5): FASTBuild Error #1102 - ObjectList() - 'Compiler' ('C:\Program Files\LLVM\bin\clang++.exe') is of unexpected type 'File'. Expected 'Compiler'.
```

It seems that this was missed by 27d2f4e.

(Note that I locally changed ClangPath in testcommon.bff to the path on my machine, but that should be of no consequence in that case.)

Thanks!

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Data/TestPrecompiledHeaders/fbuild.bff
+++ b/Code/Tools/FBuild/FBuildTest/Data/TestPrecompiledHeaders/fbuild.bff
@@ -68,7 +68,7 @@ Executable( "PCHTest" )
 
         .PCHOptions         = ' -c -x c++-header $PCHInputFile$ -o"$PCHOutputFile$"'
 
-        .Compiler           = '$ClangPath$\bin\clang++.exe'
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for that. Your local change to the ClangPath is actually what caused your local problem, but it was actually only working before by luck. I was conincidentally fixing an issue where nested Aliases in .Compiler on ObjectList and Library were not handled correctly (https://github.com/fastbuild/fastbuild/commit/6f3cbd349165916f9710bed894274fc0b83da8b2) and that also uncovered this problem. That fix will be in the next release (v0.96)

Thanks!

ffulin:
(Eventually I hope to give Clang and GCC the same cleanup as I did for VS so that the tests don't duplicate their setup - i.e. it pulls from the main FASTBuild bff automatically)

PR #449, opened 2018-06-24
Fix alloc/dealloc mismatch in ToolManifest

dummyunit:
This was found by AddressSanitizer.
In two places a `FileStream` object is allocated with `new` (obviously) and deallocated with `::Free` by `AutoPtr`. This case of alloc/dealloc mismatch is actually bad because `FileStream` has a non-trivial destructor and in `ToolManifest::Deserialize` it could be deleted while file is open which leads to a handle leak.

To fix this I added a new deleter class that uses `FDELETE`.
I also renamed `DefaultDeletor` to `FreeDeletor` to expose the fact that it uses `free` instead of `delete` operator (which people might expect given that the name is similar to `std::default_delete`). A hope this rename won't cause many problems in other (non-FASTBuild) code that uses `AutoPtr`.

### DIFF
--- a/Code/Core/Containers/AutoPtr.h
+++ b/Code/Core/Containers/AutoPtr.h
@@ -6,17 +6,26 @@
 //------------------------------------------------------------------------------
 #include "Core/Mem/Mem.h"
 
-// DefaultDeletor - free using "Free"
+// FreeDeletor - free using FREE
 //------------------------------------------------------------------------------
-class DefaultDeletor
+class FreeDeletor
 {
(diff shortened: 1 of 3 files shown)


ffulin:
Thanks for the fix. This is definitely a bug, but thankfully it's unlikely to leak any handles since in normal case that the FileStream is transferred and freed explcitily at shutdown.

ffulin:
Thanks for the fix. This has been integrated and will be in the next release (v0.97).

Issue #447, opened 2018-06-23
Possible data race between OnReturnRemoteJob and FinalizeCompletedJobs in JobQueue


dummyunit:
A have got a MemorySanitizer report for RegressionTest_RemoteCrashOnErrorFormatting. It states that uninitialized value is used on [line 306 in JobQueue.cpp](https://github.com/fastbuild/fastbuild/blob/1c42593db528d4552d6896fabcb57e1e1b120993/Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp#L368) and that this uninitialized value is a result of deletion of the `Job` object on [line 484 in JobQueue.cpp](https://github.com/fastbuild/fastbuild/blob/1c42593db528d4552d6896fabcb57e1e1b120993/Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp#L484). Full report is available [here](https://travis-ci.com/dummyunit/fastbuild/jobs/130966620#L2545).

From looking at state transitions for `Job::m_DistributionState` this seems possible if the following happens:
1. Thread 1 (NetworkThread?) executes `OnReturnRemoteJob`, sees that there is a race with local job (line 358), cancels local job and enters the `Sleep(1)` loop.
2. Thread 2 (WorkerThread?) executes `FinishedProcessingJob`, sees that cancellation of local job was requested (line 585), sets `m_DistributionState` to `DIST_COMPLETED_LOCALLY` and puts the job into `m_CompletedJobs`.
3. Thread 3 (MainThread?) executes `FinalizeCompletedJobs`, grabs this jobs from `m_CompletedJobs`, sees that it is in state `DIST_COMPLETED_LOCALLY` and deletes it.
4. Thread 1 wakes up from sleep and accesses the deleted `Job` object.

I am not familiar with this part of code so I am not sure that this chain of events can really happen, maybe there are other conditions that will prevent that.

ffulin:
This looks like it might be related to this: https://github.com/fastbuild/fastbuild/issues/437

dummyunit:
> FYI, I have tried to address this race condition in the above commit (though it’s not very elegant) and it does not solve #437. It may solve this issue, though.

Sorry for the delay.
I have tested your patch and with a small tweak it fixes this issue. I had to add
```cpp
m_DistributableJobs_InProgress.Erase( it );
```
[...]

Issue #437, opened 2018-06-07
Retrying failed remote jobs does not work


samhocevar:
The mechanism that blacklists hosts and retries jobs up to a certain limit does not appear to work properly. I injected random failures in `Client::Process` to force retries (`result = false; systemError = true;`) and noticed several things happen:

  - `ASSERT( Job::GetTotalLocalDataMemoryUsage() == 0 );` fails in `JobQueue::~JobQueue`
  - `job->GetDistributionState() == Job::DIST_COMPLETED_REMOTELY` should probably account for `Job::DIST_RACE_WON_REMOTELY` too in `JobQueue::ReturnUnfinishedDistributableJob`
  - `ASSERT( distState == Job::DIST_RACING );` fails in `JobQueue::FinalizeCompletedJobs` because `distState == Job::DIST_BUILDING_LOCALLY` sometimes.
  - and of course, builds fail even if the job is retried and succeeds locally

I am attaching a [build log](https://github.com/fastbuild/fastbuild/files/2080321/build003.log) of FASTBuild with `-dist -verbose -distverbose -summary`. The relevant parts are probably:

    -> Obj: C:\Users\s.hocevar\fastbuild\tmp\x64-Release\FBuildCore\Graph\ObjectListNode.obj <REMOTE: farmagent13>
    10>Obj: C:\Users\s.hocevar\fastbuild\tmp\x64-Release\FBuildCore\Graph\ObjectListNode.obj <LOCAL RACE>
    Got Result: farmagent13 - C:\Users\s.hocevar\fastbuild\tmp\x64-Release\FBuildCore\Graph\ObjectListNode.obj (Won Race)
    Remote System Failure!
     - Blacklisted Worker: farmagent13
     - Node              : C:\Users\s.hocevar\fastbuild\tmp\x64-Release\FBuildCore\Graph\ObjectListNode.obj
     - Job Error Count   : 1 / 3
    7> Obj: C:\Users\s.hocevar\fastbuild\tmp\x64-Release\FBuildCore\Graph\ObjectListNode.obj <LOCAL>

ffulin:
That snippet of the log certainly seems suspect:
 - we seem to consider the remote race "won" even though it failed.  Could be just a mistake in the log, or a legit problem
 - it looks while we were already racing the job, we started compiling it again locally. That should never happen

samhocevar:
[...]

PR #450, opened 2018-06-24
Add support for running tests on Travis CI

dummyunit:
Hi Franta.
This is my attempt at bringing continuous integration to FASTBuild, this time for Linux builds using Travis CI.

I hope that this attempt won't fail (efbced9b) as the last one (#151) because, as I understand, the only thing that is required to set this up (aside from merging this PR of course) is to "install" [TravisCI GitHub app](https://github.com/marketplace/travis-ci) to fastbuild/fastbuild repo. This should enable automatic builds on new commits for branches that have a `.travis.yml` file and for pull requests to such branches.
Also you can look how integration with GitHub will look in the FASTBuild case [in my fork](https://github.com/dummyunit/fastbuild/pull/1).

Currently there are build jobs for these configurations:
1. Linux, default compiler versions for Ubuntu 14.04 (GCC 4.8 and Clang 5.0), running default set of tests (`Tests` target).
2. Linux, latest compiler versions supported by FASTBuild (GCC 6 and Clang 6.0), running default set of tests (`Tests` target).
3. Linux, Clang, running tests in AddresSanitizer build.
4. Linux, Clang, running tests in MemorySanitizer build.

AddresSanitizer and MemorySanitizer are currently allowed to fail because they often break on flaky tests that currently exist (these tests need to fixed first).

There is also a disabled (commented out) configuration for runnig tests on OSX. It is disabled because some tests are failing consistently on Travis OSX VMs, so somebody has to fix these tests first.

Now the bad parts:
1. Some tests in FASTBuild are currently broken: #441, #448, #449. So merging this PR before those three is pointless because Travis builds will fail every time.
2. Some tests for build distribution (at least RegressionTest_RemoteCrashOnErrorFormatting and TestLocalRace) fail _sometimes_, this is probably related to #447 and #437. So until those issues are fixed Travis builds wouldn't be a reliable indicator for problems in new PRs (although build for a PR can be restarted even without leaving GitHub from the Checks tab of the PR).
3. Test TestSharedMemory.CreateAccessDestroy also sometimes fails (Assert: t.GetElapsed() < 10.0f), but only on Travis. It never failed for me locally, so I don't know what to with it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,172 @@
+language: cpp
+
+matrix:
+  include:
+    # Linux, default compilers
+    - os: linux
+    # OSX, default compilers
+    - os: osx
+      compiler: clang
+    # Windows, we expect MSVC 14.16.27023 to be installed.
+    - os: windows
+    # Linux, latest supported compilers
+    - os: linux
+      env: GCC=gcc-8 CLANG=clang-8
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-8
+          packages:
+            - gcc-8
+            - g++-8
+            - clang-8
+    # Linux, AddressSanitizer
+    - os: linux
+      env: CFG=asan CLANG=clang-8
+      # Use latest clang when building with sanitizers.
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-8
+          packages:
+            - clang-8
+            - llvm-8 # for llvm-symbolizer
+      sudo: true # LeakSanitizer doesn't work inside Travis containers, see https://github.com/travis-ci/travis-ci/issues/9033.
+    # Linux, MemorySanitizer
+    - os: linux
+      env: CFG=msan CLANG=clang-8
+      # Use latest clang when building with sanitizers.
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+            - llvm-toolchain-trusty-8
+          packages:
+            - clang-8
+            - llvm-8 # for llvm-symbolizer
+  allow_failures:
+    - env: CFG=asan CLANG=clang-8
+    - env: CFG=msan CLANG=clang-8
+  fast_finish: true
+
+install:
+  # Download and unpack archive for a release version.
+  - DL_VER=0.98
+  - |
+    case "${TRAVIS_OS_NAME}" in
+    linux)   FN=FASTBuild-Linux-x64-v${DL_VER}.zip   FBUILD=fbuild ;;
+    osx)     FN=FASTBuild-OSX-x64-v${DL_VER}.zip     FBUILD=FBuild ;;
+    windows) FN=FASTBuild-Windows-x64-v${DL_VER}.zip FBUILD=FBuild.exe ;;
+    esac
+  - wget http://fastbuild.org/downloads/v${DL_VER}/${FN}
+  - unzip ${FN}
+  - chmod +x ./${FBUILD}
+  # Set up some variables for convenience.
+  - |
+    case "${TRAVIS_OS_NAME}" in
+    linux)   NCPU=$(nproc) ;;
+    osx)     NCPU=$(sysctl -n hw.ncpu) ;;
+    windows) NCPU=${NUMBER_OF_PROCESSORS} ;;
+    esac
+  - FBUILD_CMD="../${FBUILD} -j${NCPU} -noprogress"
+  - GCC=${GCC:-gcc}
+  - GXX=${GCC/gcc/g++}
+  - CLANG=${CLANG:-clang}
+  - CLANGXX=${CLANG/clang/clang++}
+  # On Linux/OSX print paths to binaries for the installed compilers.
+  # On Windows check that all tools are present at paths we expect, if not then try to find them on the disk and fail the build.
+  - |
+    case "${TRAVIS_OS_NAME}" in
+    linux)
+      for x in ${GCC} ${GXX} $(${GCC} -print-prog-name=cc1) $(${GCC} -print-prog-name=cc1plus) ${CLANG} ${CLANGXX}; do
+        echo $x; which $x && readlink -f $(which $x)
+        echo
+      done
+      true # These commands are purely informational, don't fail the build if they fail.
+      ;;
+    osx)
+      for x in ${CLANG} ${CLANGXX}; do
+        echo $x; which $x
+        echo
+      done
+      true # These commands are purely informational, don't fail the build if they fail.
+      ;;
+    windows)
+      SUCCESS=true
+      if [[ ! -f 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools\VC\Tools\MSVC\14.16.27023\bin\Hostx64\x64\cl.exe' ]]; then
+        find 'C:\Program Files (x86)' 'C:\Program Files' -iname cl.exe
+        SUCCESS=false
+      fi
+      if [[ ! -f 'C:\Program Files (x86)\Windows Kits\10\Bin\10.0.17134.0\x64\RC.exe' ]]; then
+        find 'C:\Program Files (x86)' 'C:\Program Files' -iname rc.exe
+        SUCCESS=false
+      fi
+      if [[ ! -f 'C:\Program Files\LLVM\bin\clang++.exe' ]]; then
+        find 'C:\Program Files' -iname clang++.exe
+        SUCCESS=false
+      fi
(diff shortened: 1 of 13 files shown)


.travis.yml
```diff
@@ -0,0 +1,70 @@
...
+  # Bootstrap fbuild using Makefile
+  - ${MAKE_CMD} -s -f Makefile.bootstrap fbuild
+  # Run tests in the selected configuration
+  - |
+    case "${TRAVIS_OS_NAME}:${CFG}" in
+      linux:)
+        ${FBUILD_CMD} {CoreTest,FBuildTest}-RunTest-x64Linux-Debug CoreTest-RunTest-x64Linux-Release
+        ;;
+      osx:)
+        ${FBUILD_CMD} {CoreTest,FBuildTest}-RunTest-x64OSX-Debug CoreTest-RunTest-x64OSX-Release
```
ffulin: I think the FBUILD_CMD is incorrect for OS X (should be "FBuild")

.travis.yml
```diff
@@ -0,0 +1,70 @@
...
+  # Set up some variables for convenience
+  - export NPROC=$(if [[ "${TRAVIS_OS_NAME}" == osx ]]; then sysctl -n hw.ncpu; else nproc; fi)
+  - export MAKE_CMD="make -j${NPROC}"
+  - export FBUILD_CMD="./fbuild -j${NPROC} -noprogress"
+  - export GCC=${GCC:-gcc} CLANG=${CLANG:-clang}
+  - export GXX=${GCC/gcc/g++} CLANGXX=${CLANG/clang/clang++}
+  # Patch configs to use installed compilers
+  - |
+    sed -ie "1i\\
+    #define CI_BUILD
```
ffulin: This CI_BUILD define doesn't seem to be visible to FASTBuild. I also don't understand why the bff changes would be needed.

.travis.yml
```diff
@@ -0,0 +1,70 @@
...
+    s:CI_GCC_BINARY:$(which ${GCC}):
+    s:CI_GXX_BINARY:$(which ${GXX}):
+    s:CI_CC1_BINARY:$(${GCC} -print-prog-name=cc1):
+    s:CI_CC1PLUS_BINARY:$(${GCC} -print-prog-name=cc1plus):
+    s:CI_CLANG_BINARY:$(which ${CLANG}):
+    s:CI_CLANGXX_BINARY:$(which ${CLANGXX}):" fbuild.bff Tools/FBuild/FBuildTest/Data/testcommon.bff
+  # Bootstrap fbuild using Makefile
+  - ${MAKE_CMD} -s -f Makefile.bootstrap fbuild
```
ffulin: Is it possible to submit the binary for fbuild to the repo instead of bootstrapping like this, so that we can avoid the maintenance cost of maintaining the makefile?

dummyunit: I think it would be the worst option, let me expand on this.

We have three ways of obtaining `fbuild` binary in the CI environment:
1. Download the binary from fastbuild.org.
    **Pros**:
    * Easy to implement.
    * Low (but continuous) maintenance: Version in the URL will need to be updated after each release, or when .bff files will actually start using features of a new release.

    **Cons**:
    * Builds will fail if CI will fail to download the file for some reason.
2. Store the binary to the repo.
    **Pros**:
    * Even easier to implement.
    * Low (but continuous) maintenance: Committed binary will need to be updated after each release, or when .bff files will actually start using features of a new release.

    **Cons**:
    * Doing so will significantly increase size of the repo over time. This is bad because everybody will now have to download and store these files that nobody besides CI need.
    For example: current size of the repo (`.git/objects` directory) is about 9Mb, size of the Linux binary is 4.2Mb, OSX binary - 1.4Mb, so after 2 releases the repo size will double.
3. Build the binary from source using other build system (make in this cake).
    **Pros**:
    * Already implemented :)
    * Because the makefile does a non-unity build we are testing non-unity builds for free.
    * Zero maintenance (most of the time): Proposed makefile can build any tagged version of FASTBuild starting from v0.80.

    **Cons**:
    * We are not testing that current sources can be built with a release version of FASTBuild. So if some new .bff feature (not included in any released FASTBuild version) will be used in the build configuration CI will not catch that.
    * Higher upper bound on maintenance costs: If structure of the source tree will change (more directory layers, new libraries, etc) or if some new mandatory compiler flag (like requiring newer C++ standard) will appear then the makefile will need to updated.
    This is definitely harder than incrementing a number in the URL or commiting a new file. But so far changes that would require an update happened only once (between 0.77 and 0.80).

Also having a way to build the `fbuild` binary that doesn't require a `fbuild` binary has some benefits not related to CI:
1. It enables initial build in environments where pre-built binary doen't fork (e.g. different libc implementation: uclibc or musl).
2. It enables initial build in situations when running binaries downloaded from the Internet is prohibited (e.g. company security policy).

Let me know what you think. If you still don't want the makefile, I think option 1 would be better.

.travis.yml
dummyunit: This `sed` command inserts this define at the top of `fbuild.bff` and later when this define is present hard-coded paths to compiler binaries are replaced with paths that exist in the current environment.  
But since we now have dedicated files for compiler configuration this logic can be done much cleaner.

As to why this is needed at all - it is because in each Linux distribution compilers have different versioned names and stored in different places (especially `cc1` and `cc1plus` binaries).

.travis.yml
dummyunit: Bootstrap makefile produces binaries with same names as on Linux (all lower case), so it works. But if we replace makefile with downloaded binaries then this will need to be adjusted.

dummyunit:
Rebased on top current of the current dev branch:
1. Now at least half of CI specific logic is in dedicated compiler files in `External/SDK/`.
2. Updated to GCC 8 and Clang 8 since FASTBuild can now be built by them.
3. Enabled OSX builds, problems that they had before have disappeared, somehow.

Also currently builds are failing die to these bugs: #551, #553.

.travis.yml
ffulin: I think using pre-built binaries (either pulling them from fastbuild.org or checking them into the repo) would greatly reduce the complexity and maintenance of this change. Pulling them from fastbuild.org would only have to be updated 3-4 times a year at the current release rate.

For what it's worth, until now, the Linux and OSX executables have had no effort put into their size (so the Linux ones contained full debug info for example). As we approach v1.0, I've extracted the debug info from the Linux exe and will look into the size of the OSX exe soon. Your point about binaries bloating the gitrepor still stands of course, but this will be less of an issue with the Linux exe now comparable to the Windows one in size.

ffulin:
I just wanted to say that I really appreciate all the effort you have put into this so far - it will be very valuable once we get it going.

.travis.yml
dummyunit: > It does sound like option 1) is the way to go to me.

Changed build script to download the archive for the latest release from fastbuild.org

.travis.yml
dummyunit: Now `FBUILD_CMD` should use correct name for the FASTBuild binary on each operating system.

dummyunit:
Updated Travis config to download pre-built binaries and use them.

To compensate for removed indirect check of non-unity builds a have added additional build pass with `-nounity -clear` flags.

Also I found the cause of sporadic failures of TestSharedMemory in OSX and Linux+MSan configurations, opened a different PR for it: #559. With that fix OSX builds are consistently succeeding so I removed them from `allow_failures`.

While doing all of that I noticed that, since I initially started adding CI support, Travis added the Windows build environment. So I tried it and it worked pretty good, so now Windows build are tested too.

Instead of computing paths for compilers, etc dynamically (like in Linux and OSX builds) Windows build uses hard-coded paths to MSVC, WinSDK, LLVM and C# compiler (apparently it is not a part of VS). This is not ideal because when any of these will be updated on Travis our Windows build will break. But I think I can be fixed later.

Now there are only two sources of failures remain:
1. TestLocalRace sometimes fails on Linux, like [here](https://travis-ci.com/dummyunit/fastbuild/jobs/198940555#L963). Log shows that `/usr/bin/ar` can open an object file that was (supposedly) successfully built remotely and just written to disk. I could not reproduce this localy yet.
2. Tests on Windows ans OSX sometimes fail to connect via loopback interface, like [here](https://travis-ci.com/dummyunit/fastbuild/jobs/199317960#L402) and [here](https://travis-ci.com/dummyunit/fastbuild/jobs/199317959#L449). I don't know what we could do about it besides not touching network at all.

ffulin:
This change has now been integrated! I've activated/authorized Travis CI on the fastbuild repo, but no builds are triggering yet.

ffulin:
I've integrated a small changed into the dev branch and a build has been triggered - hurray!

PR #453, opened 2018-06-25
New -debugprocessor command line argument to help debugging preprocessor problems

belkiss:
Hello!

A change quite old from JS. I don't remember if we sent it to you before...


Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -552,6 +552,14 @@ Node::BuildResult ObjectNode::DoBuildWithPreProcessor2( Job * job, bool useDeopt
     Args fullArgs;
     AStackString<> tmpDirectoryName;
     AStackString<> tmpFileName;
+
```
ffulin: Forcing the use of the preprocessor will cause the build to fail for various things (some things are completely incompatible with the preprocessor)

jspelletier:
This is very old code that I think I might have added while debugging the VS2012 enum bug. This is an experimental option that of course can have side effects. Same as when running with nolocalrace that can result in infinite wait of build completion...
This is a debugging option so if it is incompatible with some setup is probably not so bad in my opinion.

PR #454, opened 2018-06-25
Add brokerage path to the log on build startup with -dist.

belkiss:
Small improvement :)


ffulin:
Thanks for the improvement! This will be in the next release (v0.97)

PR #455, opened 2018-06-25
Make sure the directory exist before writing the monitor file

belkiss:
I also added some logs in case something goes wrong.

This fixes the issue I had opened here:
https://github.com/yass007/FASTBuildMonitor/issues/18


ffulin:
Thanks for the fix! This will be in the next release (v0.97)

PR #456, opened 2018-06-25
Fix incremental link when at least one /OPT:NO* option is present in the command line

belkiss:
Hello!

A fix from a fellow developer at Ubisoft.

This occurred when we had /OPT:NOREF in the linker command line for instance.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
@@ -711,17 +711,17 @@ void LinkerNode::GetAssemblyResourceFiles( Args & fullArgs, const AString & pre,
 
             if ( IsStartOfLinkerArg_MSVC( token, "OPT" ) )
             {
-                if ( token.FindI( "REF" ) )
+                if ( token.FindI( ":REF" ) )
                 {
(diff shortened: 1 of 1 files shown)


Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
```diff
@@ -711,17 +711,17 @@ void LinkerNode::GetAssemblyResourceFiles( Args & fullArgs, const AString & pre,
 
             if ( IsStartOfLinkerArg_MSVC( token, "OPT" ) )
             {
-                if ( token.FindI( "REF" ) )
+                if ( token.FindI( ":REF" ) )
```
ffulin: Options can also be comma delimited, i.e. /OPT:REF,ICF so this would break those cases.

ffulin:
I've implemented a fix for this that handles the comma delimited case: https://github.com/fastbuild/fastbuild/commit/f50f66fd0c77c9efd412c99647c22dcc3410eb38

This fix will be in the next release (v0.97). I'll leave this pull request open until that version is released.

ffulin:
v0.97 has been released which includes a fix for the comma delimited case so we can close this PR now.

PR #457, opened 2018-06-25
Fix assert when race occur and local and remote finish at the same time

belkiss:
What could happen is that the OnReceive would enter the job with RACING status, and try to cancel the local job when that one just finishes
Because access to job status is protected by a mutex (m_DistributedJobsMutex), that mutex was released by the receiving code after setting the status to cancel, allowing the locally won race updating code to continue.
That one expected to be still in RACING, and was asserting on the value.

Now we handle the cancel request and set the status to RACE_WON_LOCALLY, as we want to give precedence to local job (they have less performance impacts).
Note that the job is deleted by the remote handling thread.


ffulin:
Thanks for the fix. We ended up with 2 resolutions for this issue, and I've integrated the other one (https://github.com/fastbuild/fastbuild/pull/517) so I'm going to close this one. This fix will be in the next release (v0.98).

PR #458, opened 2018-06-25
Fix worker tray icon tooltip

belkiss:
With this change the tooltip is back (it's been broken for a long while!) :)


ffulin:

PR #459, opened 2018-07-02
Added distribution of tests, worker tags, and sandbox features

bburgin:
Added distribution of tests, worker tags, and sandbox features
1. Distribution of tests can help speed up test exe running
2. Worker tags allow a compiler or test to use specific workers
3. The sandbox feature allows workers to run compilers and tests in a sandbox, to improve security
4. Added tests and docs for these new features
Also:
5. Added refreshing of the worker list, in case remote workers are rebooting
6. Migrated worker .settings file to bff format, so the user can modify fields ( including the new WorkerTags field )
7. Added compressing and uncompressing of  ToolManifest data
8. Improved error code printing, codes are consistently printed as hexadecimal now
9. Fixed some crash and memory leak issues

### DIFF
--- a/Code/Core/Containers/Array.h
+++ b/Code/Core/Containers/Array.h
@@ -20,7 +20,7 @@ public:
     explicit Array( const Array< T > & other );
     explicit Array( const T * otherBegin, const T * otherEnd );
     explicit Array( size_t initialCapacity, bool resizeable = false );
-    ~Array();
+    virtual ~Array();
 
     void Destruct();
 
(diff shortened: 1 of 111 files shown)


ffulin:
Thanks for the pull request. It looks like there are some interesting changes here. I won't be able to look into your changes thoroughly for a few weeks, but I'll provide some initial rough thoughts (because it's already been a while since you created the PR).

I think we're going to have difficulty integrating this change because it touches so many core systems in so many ways that it's difficult to thoroughly understand the changes and check for regressions. There are new features, bug fixes, improvements and white space changes all mixed together. I have questions about many of these that we could probably address individually, but as a whole set in one pull request, it's hard to tear apart that changes and thoroughly review them. If we go forward, trying to integrate this as one monolith, then it's likely we'll be continually "reset" as other changes cause merge conflicts.

I think to make some headway on this, we need to break this pull request apart, dealing with each of the changes one by one (roughly along the lines of your numbered list). I would suggest we start with the simpler, non-controversial, and lowest level bug fixes/improvements:
 - memory leak fixes
 - crash fixes
 - compression of ToolManifest data
 - worker list refreshing

Then we could extract/resolve some of the changes I'm not sure about:
 - Worker using bff format (conceptually not sure we want this. It also loses backwards compatibility with existing settings)
 - Error formatting changes (it is fairly windows-centric way print errors as hex. Need some thought on if we should differentiate per platform, or print both and how to manage this cleanly without lots of code duplication)
 - Virtual function added to the Array<> class can't be kept (will bloat memory usage and runtime cost of all systems)

Then we could tackle the new features individually:
 - Sandbox
 - Remote execution of Tests
 - Tagging
(I realize that you probably need all of these features together for your use-case so may have considered the features to be intertwined)

Let me know what you think.

bburgin:
Thanks for your help. I will try to split this up into multiple pull requests.

bburgin:
Closing this pull request, to add smaller ones in its place.

Issue #365, opened 2018-01-12
Custom dependency checker


awefers:
Hi, I was wondering what you think about adding support for custom dependency checkers. This would be useful for building assets. As an example, scan a 3D model asset for referenced textures and use the list of textures as input to an ObjectList() that then uses a custom "compiler" to compress the textures. The dependency checker could be a simple command line tool that writes the list of file paths to stdout.

gkoreman:
Added PR #461

PR #461, opened 2018-07-07
New functionality to allow PreBuildDependencyPreprocessor.

gkoreman:
Hi Franta, this pull request adds functionality to allow a user to specify a preprocessor that can scan the input file and list it's dependencies which will be built before being queued for build. This change allows for more complicated asset builds to be compiled using FASTBuild while respecting their dependency chain.

I know it's a rather large and structural pull request and as always I'm happy to make changes and refactor if there is another way you would like this implemented. I think I took the approach you suggested from your comments in issue #365.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/Node.h
+++ b/Code/Tools/FBuild/FBuildCore/Graph/Node.h
@@ -110,8 +110,11 @@ public:
         NOT_PROCESSED,      // no work done (either not part of this build, or waiting on static dependencies )
         PRE_DEPS_READY,     // pre-build deps processed
         STATIC_DEPS_READY,  // static dependencies are uptodate - we are ready to DoDynamicDeps
+		PRE_DYN_DEPS_READY, // pre-build dynamic deps list updated
+		PRE_DYN_DEPS_DONE,  // pre-build dynamic deps list has been updated, wait for pre build dynamic deps to be ready
         DYNAMIC_DEPS_DONE,  // dynamic deps updated, waiting for dynamic deps to be ready
         BUILDING,           // in the queue for building
+		PRE_DYN_DEPS_BUILDING, // use the job queue to build dynamic dependencies
         FAILED,             // failed to build
         UP_TO_DATE,         // built, or confirmed as not needing building
     };
@@ -187,7 +190,8 @@ protected:
     friend class WorkerThread;
 
     inline const Dependencies & GetPreBuildDependencies() const { return m_PreBuildDependencies; }
-    inline const Dependencies & GetStaticDependencies() const { return m_StaticDependencies; }
+	inline const Dependencies & GetPreBuildDynamicDependencies() const { return m_PreBuildDynamicDependencies; }
+	inline const Dependencies & GetStaticDependencies() const { return m_StaticDependencies; }
     inline const Dependencies & GetDynamicDependencies() const { return m_DynamicDependencies; }
 
     void SetName( const AString & name );
@@ -204,10 +208,13 @@ protected:
 
     // each node must implement these core functions
     virtual bool DoDynamicDependencies( NodeGraph & nodeGraph, bool forceClean );
-    virtual bool DetermineNeedToBuild( bool forceClean ) const;
+	virtual bool DeterminePreBuildDynamicDependenciesNeedToBuild(bool forceClean) const;
+	virtual bool DetermineNeedToBuild(bool forceClean) const;
+	virtual BuildResult DoPreBuildDynamicDependencies( Job * job );
     virtual BuildResult DoBuild( Job * job );
     virtual BuildResult DoBuild2( Job * job, bool racingRemoteJob );
-    virtual bool Finalize( NodeGraph & nodeGraph );
+	virtual bool PreBuildDynamicDependenciesFinalize(NodeGraph & nodeGraph);
+	virtual bool Finalize( NodeGraph & nodeGraph );
 
     inline void     SetLastBuildTime( uint32_t ms ) { m_LastBuildTimeMs = ms; }
     inline void     AddProcessingTime( uint32_t ms ){ m_ProcessingTime += ms; }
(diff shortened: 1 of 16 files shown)


Code/Tools/FBuild/FBuildCore/Graph/Node.h
```diff
@@ -110,8 +110,11 @@ class Node : public Struct
         NOT_PROCESSED,      // no work done (either not part of this build, or waiting on static dependencies )
         PRE_DEPS_READY,     // pre-build deps processed
         STATIC_DEPS_READY,  // static dependencies are uptodate - we are ready to DoDynamicDeps
+		PRE_DYN_DEPS_READY, // pre-build dynamic deps list updated
+		PRE_DYN_DEPS_DONE,  // pre-build dynamic deps list has been updated, wait for pre build dynamic deps to be ready
         DYNAMIC_DEPS_DONE,  // dynamic deps updated, waiting for dynamic deps to be ready
         BUILDING,           // in the queue for building
+		PRE_DYN_DEPS_BUILDING, // use the job queue to build dynamic dependencies
```
gkoreman: This stage actually happens between PRE_DYN_DEPS_READY and PRE_DYN_DEPS_DONE, however I put it after BUILDING so that all of the logic that checks to see if this node is currently in a job queue can stay as-is. There are multiple spots where we check if state > BUILDING.

Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -1050,10 +1056,57 @@ void NodeGraph::BuildRecurse( Node * nodeToBuild, uint32_t cost )
...
+		}
+
+		nodeToBuild->SetState(Node::PRE_DYN_DEPS_DONE);
+	}
+
+    ASSERT( ( nodeToBuild->GetState() == Node::PRE_DYN_DEPS_DONE ) ||
             ( nodeToBuild->GetState() == Node::DYNAMIC_DEPS_DONE ) );
 
-    if ( nodeToBuild->GetState() != Node::DYNAMIC_DEPS_DONE )
+    if ( nodeToBuild->GetState() == Node::PRE_DYN_DEPS_DONE )
```
gkoreman: I changed this from != to == to match the pattern in the rest of this state machine. Logic should be identical.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -205,6 +222,114 @@ ObjectNode::~ObjectNode()
...
+// DoPreBuildDynamicDependencies
+//------------------------------------------------------------------------------
+/*virtual*/ Node::BuildResult ObjectNode::DoPreBuildDynamicDependencies(Job * job)
+{
+	m_PreBuildDynamicDependencyNames.Clear();
+
+	Args fullArgs;
+
+	{
+		Array< AString > tokens(1024, true);
```
gkoreman: This could be combined into BuildArgs. Originally I broke it out so that there would be less impact on established code, but looking at it now it may make more sense to use BuildArgs and take advantage of other compiler specific substitutions if needed. At least it would be consistent. Thoughts?

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -205,6 +222,114 @@ ObjectNode::~ObjectNode()
...
 }
 
+// DeterminePreBuildDynamicDependenciesNeedToBuild
+//------------------------------------------------------------------------------
+/*virtual*/ bool ObjectNode::DeterminePreBuildDynamicDependenciesNeedToBuild(bool forceClean) const
+{
+	if( GetPreBuildPreprocessor() == nullptr )
+		return false;
+
+	if ( forceClean ||
```
gkoreman: Currently we only re-build dynamic dependencies if the source file or the preprocessor changes, since the assumption is that the preprocessor only gathers dependencies from the single input file. If this is not the case perhaps we should be checking if any of the static dependencies have changed?

ffulin:
It'll be a while before I can provide detailed feedback on this change or integrate anything, but I'll provide some very brief initial thoughts for your consideration as this PR has been up for a while.

I'm a little concerned about adding another set of dependencies to the node graph. As well as adding overhead to all node graph traversal, we'll need to be sure all the places in the code dealing with dependencies are updated.

I'll need to look into your changes in more depth to be sure I understand your goals and to see if there is an alternative way to do what you want.

gkoreman:
Hi @ffulin, thanks for taking a look at this PR. I know it's a large change and I appreciate the comments.

Perhaps I can add some context around this PR and a set of other PR's I have been experimenting with. Essentially I am looking at FASTBuild as a platform for building all game assets, similar to what your former colleagues are doing at Ubisoft with their PPD tool.

I see this as a reasonable and logical extension of FASTBuild, but agree that this extra functionality must not compromise the primary purpose of FASTBuild as a code build system. 

From the experimentation I have done so far I see a few features that I believe will enable asset building using FASTBuild:

1. Allow build tools to expose dynamic dependencies at build time. For example, a material will typically require one or more shaders and multiple texture assets to be compiled and compressed. These additional dependencies cannot be represented statically based on file type or file name but are revealed after inspecting the contents of the source. These may be build time dependencies or runtime dependencies, but I'm not sure that matters for the system. This PR is meant to provide this feature.

2. Allow chaining of build steps based on per-file dependencies. For example a mesh may depend on a material which depends on a shader. Currently this requires the use of static PreBuildDependencies on an entire build step.

3. Allow multiple static outputs from a single node. A processor frequently outputs more than one file. Representing this in FASTBuild will allow chaining build steps on all outputs, not just the single "primary" output from an ObjectNode.

4. Allow multiple dynamic outputs from a node. This is an extension of point number 3. Consider a 'level' processor that takes a level source file as input. This level may be broken into streaming chunks and output as dozens of sublevels, each of which may require additional build steps.

I have been prototyping solutions to 2 and 3 by extending ObjectListNode and ObjectNode, however your concerns about the dependency graph are making me rethink this proposal. Perhaps creating dedicated nodes for asset processing would be more appropriate. This would allow the core functionality of FASTBuild to remain relatively unaffected.

Is this type of functionality something that you are interested in for FASTBuild?

PR #464, opened 2018-07-17
Fixed cacheinfo output for Mac/Linux

bkslasher:
Cacheinfo output displays incorrect age for Mac/Linux. Windows file time uses a number of 100-nanosecond intervals, whereas for Mac/Linux it's a number of nanoseconds.


ffulin:
This fix has been integrated and will be in the next release (v0.97). Thanks again!

PR #472, opened 2018-08-22
Compress manifest

bburgin:
Added compressing and uncompressing of ToolManifest data


bburgin:
Thanks for the suggestions. I posted an update.

PR #474, opened 2018-08-22
Crash fixes

bburgin:
1. Call InitBuckets() in SmallBlockAllocator::Alloc() even if size too big, to avoid later crash if only too big allocs exist in a run.
2. Added some nullptr checks to avoid nullptr dereference crashes.
3. Set some ptrs to nullptr during cleanup steps, to avoid later deference crashes.
4. Changed some methods to pass NodeGraph by pointer instead of by reference, since the DoBuild part of the logic in the methods still needs to run. So, I have it check for nullptr in lower methods.

### DIFF
--- a/Code/Core/Mem/SmallBlockAllocator.cpp
+++ b/Code/Core/Mem/SmallBlockAllocator.cpp
@@ -108,16 +108,16 @@ NO_INLINE void SmallBlockAllocator::InitBuckets()
 //------------------------------------------------------------------------------
 void * SmallBlockAllocator::Alloc( size_t size, size_t align )
 {
-    // Check we can support allocation of this size
-    if ( size > BUCKET_MAX_ALLOC_SIZE )
-    {
-        return nullptr; // Too big
-    }
-
     // Lazy initialization of buckets to support static allocations
     if ( s_BucketMemoryStart == nullptr )
     {
         InitBuckets();
+    }
+
+    // Check we can support allocation of this size
+    if ( size > BUCKET_MAX_ALLOC_SIZE )
+    {
+        return nullptr; // Too big
     }
 
     // Handle 0 byte allocations
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
@@ -790,7 +790,12 @@ bool BFFParser::ParseIncludeDirective( BFFIterator & iter )
     AStackString<> includeToUse;
     if (PathUtils::IsFullPath(include) == false)
     {
-        AStackString<> fileName( iter.GetFileName() );
+        AStackString<> fileName;
+        const char * iterName = iter.GetFileName();
+        if ( iterName )
+        {
+            fileName += iterName;
+        }
         const char * lastSlash = fileName.FindLast( NATIVE_SLASH );
         lastSlash = lastSlash ? lastSlash : fileName.FindLast( OTHER_SLASH );
         lastSlash = lastSlash ? ( lastSlash + 1 ): fileName.Get(); // file only, truncate to empty
--- a/Code/Tools/FBuild/FBuildCore/FBuild.cpp
+++ b/Code/Tools/FBuild/FBuildCore/FBuild.cpp
@@ -237,6 +237,11 @@ bool FBuild::GetTargets( const Array< AString > & targets, Dependencies & outDep
 {
     ASSERT( !targets.IsEmpty() );
 
+    if ( m_DependencyGraph == nullptr )
+    {
+        return false;
+    }
+
     // Get the nodes for all the targets
     const size_t numTargets = targets.GetSize();
     for ( size_t i=0; i<numTargets; ++i )
@@ -311,6 +316,11 @@ bool FBuild::SaveDependencyGraph( const char * nodeGraphDBFile ) const
 
     PROFILE_FUNCTION
 
+    if ( m_DependencyGraph == nullptr )
+    {
+        return false;
+    }
+
     FLOG_INFO( "Saving DepGraph '%s'", nodeGraphDBFile );
 
     Timer t;
@@ -368,7 +378,10 @@ bool FBuild::SaveDependencyGraph( const char * nodeGraphDBFile ) const
 //------------------------------------------------------------------------------
 void FBuild::SaveDependencyGraph( IOStream & stream, const char* nodeGraphDBFile ) const
 {
-    m_DependencyGraph->Save( stream, nodeGraphDBFile );
+    if ( m_DependencyGraph != nullptr )
+    {
+        m_DependencyGraph->Save( stream, nodeGraphDBFile );
+    }
 }
 
 // Build
@@ -388,6 +401,7 @@ bool FBuild::Build( Node * nodeToBuild )
     m_LastProgressCalcTime = 0.0f;
     m_SmoothedProgressCurrent = 0.0f;
     m_SmoothedProgressTarget = 0.0f;
+
     FLog::StartBuild();
 
     // create worker dir for main thread build case
@@ -402,12 +416,12 @@ bool FBuild::Build( Node * nodeToBuild )
     for ( ;; )
     {
         // process completed jobs
-        m_JobQueue->FinalizeCompletedJobs( *m_DependencyGraph );
+        m_JobQueue->FinalizeCompletedJobs( m_DependencyGraph );
 
         if ( !stopping )
         {
             // do a sweep of the graph to create more jobs
-            m_DependencyGraph->DoBuildPass( nodeToBuild );
+            NodeGraph::DoBuildPass( m_DependencyGraph, nodeToBuild );
         }
 
         if ( m_Options.m_NumWorkerThreads == 0 )
@@ -467,7 +481,7 @@ bool FBuild::Build( Node * nodeToBuild )
     }
 
     // wrap up/free any jobs that come from the last build pass
-    m_JobQueue->FinalizeCompletedJobs( *m_DependencyGraph );
+    m_JobQueue->FinalizeCompletedJobs( m_DependencyGraph );
 
     FDELETE m_JobQueue;
     m_JobQueue = nullptr;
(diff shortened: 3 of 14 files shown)


Code/Core/Mem/SmallBlockAllocator.cpp
```diff
@@ -108,18 +108,18 @@ NO_INLINE void SmallBlockAllocator::InitBuckets()
...
     if ( s_BucketMemoryStart == nullptr )
     {
         InitBuckets();
     }
 
+    // Check we can support allocation of this size
+    if ( size > BUCKET_MAX_ALLOC_SIZE )
+    {
+        return nullptr; // Too big
+    }
```
ffulin: It makes sense to robustify this allocator in case behaviour changes such that it's possible to only perform large allocations and exit, but I think it's much better to do that by making the shutdown code that was crashing not crash (rather than forcing creation of buckets we don't need). Can you describe where it was crashing?

Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
```diff
@@ -790,7 +790,12 @@ bool BFFParser::ParseIncludeDirective( BFFIterator & iter )
     AStackString<> includeToUse;
     if (PathUtils::IsFullPath(include) == false)
     {
-        AStackString<> fileName( iter.GetFileName() );
+        AStackString<> fileName;
+        const char * iterName = iter.GetFileName();
+        if ( iterName )
+        {
+            fileName += iterName;
+        }
```
ffulin: a BFFIterator should never have a nullptr for a filename. In what situation did you see this?

Code/Tools/FBuild/FBuildCore/FBuild.cpp
```diff
@@ -237,6 +237,11 @@ bool FBuild::GetTargets( const Array< AString > & targets, Dependencies & outDep
 {
     ASSERT( !targets.IsEmpty() );
 
+    if ( m_DependencyGraph == nullptr )
+    {
+        return false;
+    }
```
ffulin: It's not valid to call this function (or any other function that derefs the m_DependencyGraph) without first successfully initializing the FBuild class (which creates the m_DependencyGraph. I think none of these nullptr guards make sense.

Was there a particular crash you saw?

Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
```diff
@@ -98,26 +98,28 @@ Job * JobSubQueue::RemoveJob()
         return nullptr;
     }
```
ffulin: If this assert ever fires, there is a serious problem somewhere else in the code. We should not silently try to cope with this.

Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
```diff
@@ -546,7 +548,10 @@ void JobQueue::MainThreadWait( uint32_t maxWaitMS )
 void JobQueue::WorkerThreadWait( uint32_t maxWaitMS )
 {
     ASSERT( Thread::IsMainThread() == false );
-    ASSERT( FBuild::Get().GetOptions().m_NumWorkerThreads > 0 );
+    if (FBuild::IsValid())
+    {
+        ASSERT( FBuild::Get().GetOptions().m_NumWorkerThreads > 0 );
+    }
```
ffulin: JobQueue should never exist without the FBuild singleton existing. Did you experience a crash here? If so, that would point to a problem elsewhere that needs fixing?.

Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerThread.cpp
```diff
@@ -176,46 +176,52 @@ void WorkerThread::WaitForStop()
     }
 
     // no local job, see if we can do one from the remote queue
```
ffulin: FBuild should always exist at this point. As with the other cases, this singleton not existing points to a problem elsewhere.

Code/Core/Mem/SmallBlockAllocator.cpp
bburgin: Here is the crash stack:
 	ntdll.dll!RtlEnterCriticalSection()	Unknown
 	dummysandbox.exe!Mutex::Lock() Line 64	C++
>	dummysandbox.exe!SmallBlockAllocator::Free(void * ptr) Line 198	C++
 	dummysandbox.exe!Free(void * ptr) Line 82	C++
 	dummysandbox.exe!DefaultDeletor::Delete(void * ptr) Line 14	C++
 	dummysandbox.exe!AutoPtr<char,DefaultDeletor>::~AutoPtr<char,DefaultDeletor>() Line 25	C++
 	dummysandbox.exe!main(int argc, char * * argv) Line 93	C++
 	[External Code]	

Notice that the SmallBlockAllocator is being called but was never initialized, resulting in a crash when trying to acquire the bucket.m_Mutex.Lock() in SmallBlockAllocator::Free(). The lock object is invalid because SmallBlockAllocator Init() was never called. It was never called because all allocs in this process were too big and so returned from Alloc() early, before Init() could be called.

The fix I made here is to always Init(). If you think the Free() could be fixed up, instead, then that may work. I am unclear how, though. We could have a bool that says whether the SmallBlockAllocator was initialized or not, but wouldn't that need a lock around it?

Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
bburgin: Not sure where I saw this. Doesn't seem to be a problem now. Disregard?

Code/Tools/FBuild/FBuildCore/FBuild.cpp
bburgin: Numerous tests in the FBuildTest.exe initialize an FBuild object but don't pass in a const char * nodeGraphDBFile. Since there is no db file, m_DependencyGraph is null, resulting in a crash later in FBuild Initialize(). For these cases, FBuild Initialize() needs to return before it tries to access the graph.
Example crash stack:

>	FBuildTest.exe!NodeGraph::FindNodeInternal(const AString & fullPath) Line 1269	C++
 	FBuildTest.exe!NodeGraph::FindNode(const AString & nodeName) Line 613	C++
 	FBuildTest.exe!FBuild::Initialize(const char * nodeGraphDBFile) Line 171	C++
 	FBuildTest.exe!TestAlias::ReflectionAliasResolution() Line 62	C++
 	FBuildTest.exe!TestAlias::RunTests() Line 31	C++
 	FBuildTest.exe!UnitTestManager::RunTests(const char * testGroup) Line 128	C++
 	FBuildTest.exe!main(int __formal, char * * __formal) Line 51	C++
 	[External Code]

ffulin: nodeGraphDBFile is allowed to be null because it overrides the default.

Is it possible you have other local changes that cause this? All the tests are currently passing.

Code/Core/Mem/SmallBlockAllocator.cpp
ffulin: Ok, so the crash is in a new executable you have created, and not in some existing part of FASTBuild which explains why I have never seen it.

The SmallBlockAllocator already knows if it has been initilalized (s_BucketMemoryStart is nullptr). We could add an additional guard for that in Free. 

It's a shame however to add another branch to the free, but we can avoid that by changing the initial value s_BucketMemoryStart to a pointer outside of the valid user mode address space. This way, no large allocation can be within the memory range that would result in us calculating a valid pageIndex.

I've implemented that fix here: https://github.com/fastbuild/fastbuild/commit/4ac7570be393b9d11bad2dc953065f5bfc8e97dc

Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
ffulin: Works for me :)

Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
bburgin: Doesn't seem to be a problem now. Disregard?

Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerThread.cpp
bburgin: Doesn't seem to be a problem now. Disregard?

Code/Tools/FBuild/FBuildCore/FBuild.cpp
bburgin: This crash happens when nodeGraphDBFile is null, then FBuild Initialize() parses the .bff file from m_Options.m_ConfigFile. But in many of the tests, the bff file contains errors (to test error cases), yielding a null m_DependencyGraph and the resulting crash.

Code/Core/Mem/SmallBlockAllocator.cpp
bburgin: Your fix works great. Thanks!

Code/Tools/FBuild/FBuildCore/FBuild.cpp
ffulin: Do you have a specific example of where you are seeing this problem? All the tests that expect failure during parsing should not be calling any functions that require the DependencyGraph post failed Initialize. It is not valid to call any on functions that interact with the dependency graph if Initialize() has failed, so silently returning would be covering up a problem and is not something we want to do.

As I say, the tests are currently all passing, so I suspect any problems are more likely to do with local changes you have, perhaps violating the expected usage pattern.

bburgin: Thanks. You're right, I pulled out these changes and all seems okay. Disregard?

ffulin:
It seems the legit bug with the BlockAllocator has been fixed, and the other changes are not needed as per our discussions. I'm going to close this pull request. Let me know if we missed anything!

PR #475, opened 2018-08-23
Worker refresh

bburgin:
Added refreshing of the worker list, in case remote workers are rebooting

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/settings.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/settings.html
@@ -29,8 +29,20 @@ Defines global FASTBuild settings.
   
   // Distribution
   .Workers                          // (optional) Fixed list of workers if not using automatic discovery
-  .WorkerConnectionLimit            // (optional) Limit number of connected workers (default: 15)
-  .DistributableJobMemoryLimitMiB   // (optional) Limit memory used locally to prep jobs (default: 2048 (1024 for X86))
+  .WorkerConnectionLimit            // (optional) Limit number of connected workers, default: 15
+  .DistributableJobMemoryLimitMiB   // (optional) Limit memory used locally to prep jobs
+                                    //            default: 2048 (1024 for X86)
+  .WorkerListRefreshLimit           // (optional) When not using a static worker list,
+                                    //            max number of seconds to refresh the network share worker list.
+                                    //            default: 300 (5 minutes, in case workers are rebooting)
+                                    //            0  means no retries
+                                    //            -1 means infinite retries
+
+  .WorkerConnectionRetryLimit       // (optional) When a worker does not respond to a connection request,
+                                    //            max number of seconds to retry a worker connection.
+                                    //            default: 300 (5 minutes, in case workers are rebooting)
+                                    //            0  means no retries
+                                    //            -1 means infinite retries
 }
 </div>
     </div>
--- a/Code/Tools/FBuild/FBuildCore/Graph/SettingsNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/SettingsNode.cpp
@@ -34,13 +34,19 @@ REFLECT_NODE_BEGIN( SettingsNode, Node, MetaNone() )
     REFLECT_ARRAY(  m_Workers,                  "Workers",                  MetaOptional() )
     REFLECT(        m_WorkerConnectionLimit,    "WorkerConnectionLimit",    MetaOptional() )
     REFLECT(        m_DistributableJobMemoryLimitMiB, "DistributableJobMemoryLimitMiB", MetaOptional() + MetaRange( DIST_MEMORY_LIMIT_MIN, DIST_MEMORY_LIMIT_MAX ) )
+    REFLECT(        m_WorkerListRefreshLimitSec,      "WorkerListRefreshLimit",    MetaOptional() )
+    REFLECT(        m_WorkerConnectionRetryLimitSec,  "WorkerConnectionRetryLimit",    MetaOptional() )
 REFLECT_END( SettingsNode )
 
 // CONSTRUCTOR
 //------------------------------------------------------------------------------
 SettingsNode::SettingsNode()
 : Node( AString::GetEmpty(), Node::SETTINGS_NODE, Node::FLAG_NONE )
-, m_WorkerConnectionLimit( 15 )
+, m_WorkerListRefreshLimitSec( 300 )  // default: a maximum of 5 minutes refreshing the worker list,
+                                      // in case workers are rebooting
+, m_WorkerConnectionRetryLimitSec( 300 )  // default: a maximum of 5 minutes retrying connections,
+                                          // in case workers are rebooting
+, m_WorkerConnectionLimit( 15 )  // default: a maximum of 15 workers simultaneously connected
 , m_DistributableJobMemoryLimitMiB( DIST_MEMORY_LIMIT_DEFAULT )
 {
     // Cache path from environment
--- a/Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
@@ -24,29 +24,36 @@
 #include "Core/FileIO/MemoryStream.h"
 #include "Core/Math/Random.h"
 #include "Core/Profile/Profile.h"
+#include "Core/Time/Time.h"
+#include "Core/Tracing/Tracing.h"
 
 // Defines
 //------------------------------------------------------------------------------
 #define CLIENT_STATUS_UPDATE_FREQUENCY_SECONDS ( 0.1f )
-#define CONNECTION_REATTEMPT_DELAY_TIME ( 10.0f )
 #define SYSTEM_ERROR_ATTEMPT_COUNT ( 3 )
 #define DIST_INFO( ... ) if ( m_DetailedLogging ) { FLOG_BUILD( __VA_ARGS__ ); }
 
 // CONSTRUCTOR
 //------------------------------------------------------------------------------
-Client::Client( const Array< AString > & workerList,
+Client::Client( const Array< AString > & settingsWorkers,
                 uint16_t port,
+                int32_t workerListRefreshLimitSec,
+                int32_t workerConnectionRetryLimitSec,
                 uint32_t workerConnectionLimit,
                 bool detailedLogging )
-    : m_WorkerList( workerList )
+    : m_SettingsWorkers( settingsWorkers )
+    , m_WorkerList( 256, true )  // start with 256 capacity
+    , m_RetryingWorkers( true )  // begin with retrying of workers
+    , m_WorkerListRefreshIntervalSec( 0 )  // try immediately the first time
+    , m_WorkerListRefreshElapsedTimeSec( 0 )
+    , m_WorkerListRefreshLimitSec( workerListRefreshLimitSec )
+    , m_WorkerConnectionRetryLimitSec( workerConnectionRetryLimitSec )
     , m_ShouldExit( false )
+    , m_Exited( false )
     , m_DetailedLogging( detailedLogging )
     , m_WorkerConnectionLimit( workerConnectionLimit )
     , m_Port( port )
 {
-    // allocate space for server states
-    m_ServerList.SetSize( workerList.GetSize() );
-
     m_Thread = Thread::CreateThread( ThreadFuncStatic,
                                      "Client",
                                      ( 64 * KILOBYTE ),
@@ -73,30 +80,10 @@ Client::~Client()
 {
     ASSERT( connection );
     ServerState * ss = (ServerState *)connection->GetUserData();
-    ASSERT( ss );
-
-    MutexHolder mh( ss->m_Mutex );
-    DIST_INFO( "Disconnected: %s\n", ss->m_RemoteName.Get() );
-    if ( ss->m_Jobs.IsEmpty() == false )
+    if (ss != nullptr)
     {
-        Job ** it = ss->m_Jobs.Begin();
-        const Job * const * end = ss->m_Jobs.End();
-        while ( it != end )
-        {
-            FLOG_MONITOR( "FINISH_JOB TIMEOUT %s \"%s\" \n", ss->m_RemoteName.Get(), (*it)->GetNode()->GetName().Get() );
-            JobQueue::Get().ReturnUnfinishedDistributableJob( *it );
-            ++it;
-        }
-        ss->m_Jobs.Clear();
+        ss->Disconnect();
     }
-
-    // This is usually null here, but might need to be freed if
-    // we had the connection drop between message and payload
-    FREE( (void *)( ss->m_CurrentMessage ) );
-
-    ss->m_RemoteName.Clear();
-    ss->m_Connection = nullptr;
-    ss->m_CurrentMessage = nullptr;
 }
 
 // ThreadFuncStatic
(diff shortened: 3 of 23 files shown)


ffulin:
This seems like a good improvement, but I'm finding it difficult to integrate because:
 - it's a very large change for this functionality: I think refreshing the workers could be implemented with far fewer changes
 - it makes sweeping changes to some of the most sensitive code in the codebase
 - a bunch of refactoring is mixed in: the refactoring seems fine, but mixed in with this already difficult to review change makes it hard to be confident it is safe
 - a bunch of whitespace, indentation changes are mixed in: as with the refactoring, this makes the code hard to review

Given the incredibly sensitive nature of the worker management/network code, I think we need to break this down into smaller pieces. Perhaps a good way forward would be to extract the tidying/refactoring/non-function altering changes into other PRs and get those merged first to try to bring down the complexity of this change.

ffulin:
Thanks for updating this change to resolve conflicts and compile cleanly. I appreciate the effort you've put in.

Unfortunately, I'm still having trouble finding a safe way to integrate this change because it contains many changes that are not required to add the key functionality this is trying to address. This makes understanding and verifying the code difficult, and as mentioned before, given the particularly sensitive nature of the code being impacted this is especially scary. The large amount of files changed unnecessarily means greatly increased chances of conflicts as I integrate other simple changes, further delaying integration.

To re-iterate in more detail, this change contains at least the following unrelated or unnecessary changes:
 - Refactoring of directory listing retrieval
 - Move of time formatting function to core and associated changes
 - Adding a new OUTPUT macro
 - Whitespace changes, around other changes in particular making functional changes difficult to see
 - Other whitespace/indent changes in code making the review larger than needed
 - An almost complete re-write of worker management
 - Removal of asserts that should remain
 - comments and typo fixes unrelated to the change
 - Changes to distributed compilation tests
 - Changes to the VS2015 SDK
 - Changes to behaviour around removal of the worker availability token
 - Deletion of the ServerState message

Several of these changes would be good in isolation, but some of are questionable or need discussion. Having everything lumped in together makes the work intractable.

Code/Core/Network/TCPConnectionPool.cpp
```diff
@@ -456,7 +456,10 @@ bool TCPConnectionPool::SendInternal( const ConnectionInfo * connection, const T
     connection->m_InUse = true;
 #endif
```
ffulin: It should be impossible for  m_Socket to be INVALID_SOCKET while correctly using this class. If you saw the assert firing, it likely indicates a bug elsewhere in the code.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
```diff
@@ -51,6 +51,17 @@ <h1>Settings</h1>
   .AllowDBMigration_Experimental    // (optional) Migrate build state when bff changes to avoid rebuilds (default: false)
                                     // This enables what will eventually be standard behaviour
                                     // WARNING: This feature is experimental.
+  .WorkerListRefreshLimit           // (optional) When not using a static worker list,
+                                    //            max number of seconds to refresh the network share worker list.
+                                    //            default: 300 (5 minutes, in case workers are rebooting)
+                                    //            0  means no retries
+                                    //            -1 means infinite retries
```
ffulin: 0 as a special value for "never refresh" is reasonable (to allow this feature to be disabled), but I don't quite understand what "-1 infitinite retries" means in this context

Code/Tools/FBuild/Documentation/docs/functions/settings.html
```diff
@@ -51,6 +51,17 @@ <h1>Settings</h1>
...
+
+  .WorkerConnectionRetryLimit       // (optional) When a worker does not respond to a connection request,
+                                    //            max number of seconds to retry a worker connection.
+                                    //            default: 300 (5 minutes, in case workers are rebooting)
+                                    //            0  means no retries
+                                    //            -1 means infinite retries
```
ffulin: Does this mean "if a worker can't be connected to, try again after .WorkerConnectionRetryLimit amount of time" or does it mean "for the first .WorkerConnectionRetryLimit of a build, retry connections at some unspecified interval"?

Code/Tools/FBuild/FBuildCore/Graph/SettingsNode.cpp
```diff
@@ -34,15 +34,21 @@ REFLECT_NODE_BEGIN( SettingsNode, Node, MetaNone() )
     REFLECT(        m_WorkerConnectionLimit,    "WorkerConnectionLimit",    MetaOptional() )
     REFLECT(        m_DistributableJobMemoryLimitMiB, "DistributableJobMemoryLimitMiB", MetaOptional() + MetaRange( DIST_MEMORY_LIMIT_MIN, DIST_MEMORY_LIMIT_MAX ) )
     REFLECT(        m_AllowDBMigration_Experimental, "AllowDBMigration_Experimental", MetaOptional() )
+    REFLECT(        m_WorkerListRefreshLimitSec,      "WorkerListRefreshLimit",    MetaOptional() )
+    REFLECT(        m_WorkerConnectionRetryLimitSec,  "WorkerConnectionRetryLimit",    MetaOptional() )
```
ffulin: We can add MetaRange() to these to limit inputs to sensible values (prevent setting -2 for example)

Code/Tools/FBuild/Documentation/docs/functions/settings.html
ffulin: I think I understand. I originally thought this was "refresh every 5 minutes".  Is there a reason we should limit searching for new workers? Seems like for a lengthy build, you want to continually search for new workers if some of your workers become unavailable?

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -64,37 +70,32 @@ Client::~Client()
     ShutdownAllConnections();
 
     Thread::CloseHandle( m_Thread );
+
```
ffulin: This code seems a little odd. Can you please provide some details on why this is needed?

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -137,6 +138,292 @@ void Client::ThreadFunc()
...
+                // elapsed time has not yet met retry interval, so use the existing list
+            }
+        }
+        else  // static workers
+        {
+            // this code block only runs once per run of the app
+            OUTPUT( "Not refreshing worker list, since you specified a .Workers list in your Settings node\n" );
+
+            // clear 2d array
+            m_WorkerList.Clear();
```
ffulin: If this is run only once at startup, this list should already be empty, right?

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -137,6 +138,292 @@ void Client::ThreadFunc()
...
+                AStackString<> timeIntervalString;
+                AStackString<> timeLimitString;
+                if ( ShouldRetry( timeSinceLastRetry,
+                    m_WorkerListRefreshIntervalSec,
+                    m_WorkerListRefreshLimitSec,
+                    m_WorkerListRefreshElapsedTimeSec,
+                    timeIntervalString,
+                    timeLimitString ) )
+                {
+                    OUTPUT( "Will refresh worker list in %s, retry limit %s\n", timeIntervalString.Get(), timeLimitString.Get() );
```
ffulin: I don't think we want this output being emitted by default.

The DIST_INFO macro exists in the client to allow for output that will only be emitted with -distverbose is specified.

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -64,37 +70,32 @@ Client::~Client()
...
+        ServerState * const ss = m_ServerList[ i ];
+        ASSERT( ss );
+
+        if (ss->m_Connection != nullptr)
+        {
+            ss->m_Connection->SetUserData( nullptr );
+        }
+        FDELETE ss;
+    }
+    m_ServerList.Clear();
```
ffulin: This Clear() is unnecessary since the class is being destructed

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -171,63 +468,109 @@ void Client::LookForWorkers()
...
+    for ( size_t i=0; i<numWorkers; ++i )
     {
-        const size_t i( ( j + startIndex ) % numWorkers );
+        const size_t j( ( i + workerStartIndex ) % numWorkers );
+        const AString & worker = m_WorkerList[ j ];
+        ServerState * ss = nullptr;
+        for ( size_t m=0; m<numWorkers; ++m )
+        {
+            ServerState * const stateToCheck = m_ServerList[ m ];
+            if ( stateToCheck && stateToCheck->m_RemoteName == worker )
```
ffulin: It should be impossible for stateToCheck to be nullptr.  All other iteration of m_ServerList asserts that the pointers are non-nullptr.

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -171,63 +468,109 @@ void Client::LookForWorkers()
...
-        if ( ss.m_DelayTimer.GetElapsed() < CONNECTION_REATTEMPT_DELAY_TIME )
+        float timeSinceLastRetry = ss->m_ConnectionDelayTimer.GetElapsed();
+        if ( timeSinceLastRetry <= ss->m_ConnectionRetryIntervalSec )
         {
             continue;
         }
 
-        DIST_INFO( "Connecting to: %s\n", m_WorkerList[ i ].Get() );
-        const ConnectionInfo * ci = Connect( m_WorkerList[ i ], m_Port, 2000, &ss ); // 2000ms connection timeout
+        ss->m_ConnectionDelayTimer.Start();  // reset time
```
ffulin: We now reset the connection timer even if we successfully connect. Is that needed?

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -245,34 +588,45 @@ void Client::CommunicateJobAvailability()
...
-                    SendMessageInternal( it->m_Connection, msg );
-                    it->m_NumJobsAvailable = numJobsAvailable;
-                }
+                PROFILE_SECTION( "UpdateJobAvailability" )
+                Protocol::MsgStatus msg( numJobsAvailableForWorker );
+                SendMessageInternal( ss->m_Connection, msg );
+                ss->m_NumJobsAvailableForWorker = numJobsAvailableForWorker;
+            }
+            if ( numJobsAvailable > 0 &&  // if any job in queue
+                 numJobsAvailableForWorker == 0 )  // worker cannot build any job
```
ffulin: numJobsAvailableForWorker can only ever be less than numJobsAvailable, so it seems we only need to check one

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -500,28 +855,56 @@ void Client::Process( const ConnectionInfo * connection, const Protocol::MsgJobR
...
-        const AString & nodeName = objectNode->GetName();
+        Node* node = job->GetNode();
+        Node::Type nodeType = node->GetType();
+        ObjectNode * on = nullptr;
+        switch ( nodeType )
+        {
+            case Node::OBJECT_NODE:
+                on = node->CastTo< ObjectNode >();
+                break;
+            case Node::TEST_NODE:
```
ffulin: This seems to have snuck in from another change to distribute tests

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -500,28 +855,56 @@ void Client::Process( const ConnectionInfo * connection, const Protocol::MsgJobR
...
+                            data = (const void *)( (const char *)data + sizeof( uint32_t ) + firstFileSize );
+                            ASSERT( ( firstFileSize + secondFileSize + ( sizeof( uint32_t ) * 2 ) ) == size );
+
+                            AStackString<> pdbName;
+                            on->GetPDBName( pdbName );
+                            result = WriteFileToDisk( pdbName, (const char *)data + sizeof( uint32_t ), secondFileSize );
+                        }
+                    }
+                    break;
+                case Node::TEST_NODE:
```
ffulin: More changes from TestNode distribution

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -451,6 +805,7 @@ void Client::Process( const ConnectionInfo * connection, const Protocol::MsgJobR
     ServerState * ss = (ServerState *)connection->GetUserData();
     ASSERT( ss );
 
+    AStackString<> workerName( ss->m_RemoteName );
```
ffulin: We're now making a copy of this name? Is that needed? (I don't see anything in the code that would indicate that, so I feel I may be missing something)

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
```diff
@@ -762,14 +1175,69 @@ bool Client::WriteFileToDisk( const AString & fileName, const char * data, const
...
+    , m_Excluded( false )
+    , m_DetailedLogging( detailedLogging )
+{
+}
+
+// DESTRUCTOR( ServerState )
+//------------------------------------------------------------------------------
+Client::ServerState::~ServerState()
+{
+    MutexHolder mh( m_Mutex );
```
ffulin: holding this mutex in the destructor is suspect. It implies that something else could be trying to access this class while it is being destructed, which is invalid.

Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
```diff
@@ -203,10 +203,22 @@ bool JobQueue::HaveWorkersStopped() const
 
 // GetNumDistributableJobsAvailable
 //------------------------------------------------------------------------------
```
ffulin: This function is called in two places, and numJobsAvailable is not used in one case, and checked redundantly in another.
Additionally, I don't think there should be any nullptr entries in this array.
Maybe I'm missing something about your changes - can you provide some info please?

Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueue.cpp
```diff
@@ -203,10 +203,22 @@ bool JobQueue::HaveWorkersStopped() const
...
-size_t JobQueue::GetNumDistributableJobsAvailable() const
+void JobQueue::GetNumDistributableJobsAvailable(
```
ffulin: The function is called in two places, and in one place, numJobsAvailable is not used, and in the other, it's checked redundantly. Additionally, I don't think there should ever be any nullptrs in the m_DistributableJobs_Available array.
Do you have any more information about this part of your change - maybe I've missed something?

Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
```diff
@@ -70,16 +68,61 @@ void WorkerBrokerage::Init()
...
+    for ( size_t i=0; i<numResults; ++i )
+    {
+        const char * lastSlash = pathResults[ i ].FindLast( NATIVE_SLASH );
+        AStackString<> machineName( lastSlash + 1 );
+        contents.Append( machineName );
+    }
+}
+
+// GetRootWorkers
+//------------------------------------------------------------------------------
```
ffulin: What does it mean for something to be a "root worker"?

Code/Tools/FBuild/FBuildWorker/Worker/Worker.cpp
```diff
@@ -268,8 +268,16 @@ void Worker::UpdateAvailability()
...
-    m_WorkerBrokerage.SetAvailability( numCPUsToUse > 0);
+    
+    if ( numCPUsToUse > 0 )
+    {
+        m_WorkerBrokerage.SetAvailable();
+    }
+    else
+    {
+        // to avoid doing a lot of network file I/O when going idle and not idle,
+        // don't set unavailable here
```
ffulin: It now looks like a worker will never remove its token (until shutdown). This is undesirable - removing the token prevents other workers connecting to a worker that won't do any work for it (and using up a connection slot).

ffulin:
Thanks for trimming down the PR - this is much more manageable. I've gone over it and added some comments and questions to help us get it done. There are quite a few - I hope you understand that the very sensitive nature of this code makes me particularly pedantic :)

Code/Core/Network/TCPConnectionPool.cpp
bburgin: Yes. There's probably a race condition in my code.
This assert is only happening on mac os.
How do I debug race conditions on Travis CI mac machines?
Adding printfs would probably change the timing...

I will try to reproduce it on a mac machine on my side.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
bburgin: Yes. WorkerListRefreshLimit is an upper limit on the total time spent looking for new workers.

It is not the worker poll interval. I have the Client.cpp drive the poll interval; it starts at 0 (poll once immediately), then 5 seconds later, then after that gets doubled after each poll (exponential backoff). This limits network file i/o.

By default, the WorkerListRefreshLimit is 5 minutes, to allow time for scarce workers who are rebooting to appear after a reboot.

I assume that a worker would include its name in the brokerage path for the entire time the worker app is running.  Then clients could use the server status message to know whether a worker is currently available or not. This would minimize network file i/o, and rely on the lighter weight tcp server status messages for the more frequent availability state changes. But based on your other comment, it seems that you are toggling availability more often than the lifetime of the worker app.
Can you explain why this is?
Is it because you limit the number of workers to consider at any given time to 15:
m_WorkerConnectionLimit( 15 )  // default: a maximum of 15 workers simultaneously connected
and want that list to always contain available workers?

Code/Tools/FBuild/Documentation/docs/functions/settings.html
bburgin: WorkerConnectionRetryLimit is not an interval. It is an upper limit similar to WorkerListRefreshLimit above.
So it represents the upper limit of time spent retrying a connection to a given worker.

It allows the client to stop trying a worker that is failing to connect.
Some use cases: a worker is rebooting, a worker app crashed, client or worker was firewalled, etc.
I default the WorkerConnectionRetryLimit to 5 minutes for the reboot case, since it is the largest known value of the use cases I could think of. The other use cases above are unknown time or may never recover & connect. Also, this makes the defaults the same for both the WorkerConnectionRetryLimit and the WorkerListRefreshLimit, which is easier for users to understand.

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
bburgin: What range of lines of code? line 73 is a blank line so you probably meant some other lines?

Code/Tools/FBuild/Documentation/docs/functions/settings.html
ffulin: The backoff strategy of doubling the time between refresh attempts seems reasonable to me. I think maybe we don't need this limit at all (perhaps just a boolean to disable this new behavior, which I think can be on by default.

The current design for workers is that they add themselves to the brokerage when available, and remove themselves when not. They might be unavailable because the FBuildWorker is closed, or they might be unavailable because the machine is under load doing other things (the common case in an office environment where developers run the worked on their machines and the worker works for others when idle.

The aim of removing the token from the brokerage when unavailable is to avoid newly started builds connecting to a worker which will never do any work. The ideal situation is that the workers we connect to are available, so while this doesn't guarantee that, it greatly increases the chances.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
ffulin: Thanks for the clarification. I think we don't need this limit at all. I would expect that for the lifetime of the build, if the FBuildWorker hasn't reached the worker limit (default 15) it will continue to try to connect to available workers in the pool.

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
ffulin: Sorry - the block starting underneath this line: Lines 74 to 86 inclusive.

ShutdownAllConnections is a blocking call, so clearing the userdata after that would seem to serve no purpose. With that in mind, I'm unclear what this loop is trying to do.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
bburgin: I think we still need the WorkerListRefreshLimit, for these cases:

1. A job needs a specific worker and the worker never becomes available. This can happen with the worker tags feature. Without a refresh limit, fbuild would hang forever for this case, but I want fbuild to eventually time out.

2. Having a refresh limit would also help the -forceremote case to eventually time out, rather than hang forever.

Thoughts?


I will put the worker available / not available logic back in so that it adds and removes the file on availability.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
bburgin: I still think we need the WorkerConnectionRetryLimit, for these cases:

1. Worker crashed (or was killed by user) and brokerage file is still present. After some time, we want to exclude this worker from the 15 worker pool and not try to connect to it. Otherwise, we could build up a 15 worker pool list of machines that are all not connectable.

2. Worker is firewalled. Same reasons as above.

Thoughts?

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
bburgin: The FDELETE ss line avoids a memory leak I saw.
I am clearing the userdata before that since other FDELETE ss calls you have also cleared userdata and I was following the same pattern.

Thoughts?

Code/Tools/FBuild/Documentation/docs/functions/settings.html
ffulin: I consider the -forceremote option to be a debug option that should not be used normally. If no workers are available, the build will never finish. There is no mechanism to detect this and time out.

FASTBuild currently assumes that all remote work can be completed locally, so absence of workers would never prevent the build from completing. Do you have a case where the main machine does not have some hardware capability that your workers do that would require some work only ever be done remotely?

If code was added to FASTBuild so the local machine detected when the build was incomplete but some remote-only work was still unfinished, this would require a timeout independent of worker discovery. In-fact, it would have to detect this case even if the Sever was not created (not running in -dist mode). This timeout would in both cases be independent of the period of the build during which we'd attempt to find new workers, so I think this option still isn't needed.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
ffulin: In the current version, workers that cannot be reached (for any reason) or actively refuse the connection, do not count in the worker limit, so neither of those cases should be a problem. We should make sure that behavior remains.

Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
ffulin: ShutdownAllConnections() signals all existing connections to terminate and waits for them to do so. During this termination, the socket is closed, the connection information is removed from the TCPConnecitonPool and the OnDisconnect() is called for each connection. 

This sequence is the same as when a connection is lost during the middle of a build. All cleanup should occur in response to the OnDisconnect() event. At any point that a connection is lost, the same cleanup should happen.

If you get to this point and still have connections to free, after all connections should be freed and cleaned up any related allocations, it implies there is a leak somewhere. This loop shouldn't need to be here at all if everything is working correctly.

Code/Tools/FBuild/Documentation/docs/functions/settings.html
bburgin: Thanks for the explanation.

Yes, I will have future jobs that cannot run locally, examples:
1. A custom compiler that requires a license that is only on one shared machine, not the local machine.
2. Tests that need to be run on a pool of test machines with special hardware; hardware that is not available on the local machine.

I can see how you would prefer the timeout for this to be on the job, rather than the worker refresh. But I could see us needing two different timeouts, the two seem logically and chronologically distinct in the build flow:
a. A worker refresh timeout, with a value for when to stop looking for workers that could meet the needs of these special builds / tests.
b. A job-specific timeout that would apply after successfully connecting to a worker and the job did not complete.

Perhaps you are suggesting that the a. timeout above be moved to some other watchdog thread, not in the worker refresh logic?

Thoughts?

Code/Tools/FBuild/Documentation/docs/functions/settings.html
bburgin: Thanks. I see now that the failed connection workers do not apply to the 15 worker limit.
Thus, I concede that this connection limit is not needed for the case I mentioned.
But it may be useful as a performance optimization:

Say there are 16 machines that are in the brokerage path.
And 15 of those are not connectable (crashed, killed, firewalled, etc).
Then if we could exclude those 15 after some timeout time, then we could quickly connect to the 16th, 17th, etc machines in the next connection attempt.

Granted, this would mean we would never retry the 15 machines we timed out on. This could be bad if they at some point were restored: user relaunched the worker, unfirewalled the machine, etc.)

I guess I am worried that without a timeout we would be slow to get to the 16th, 17th, etc machines, since the current algorithm randomly chooses a machine from the brokerage path.
Will the random number generator eventually get to the 16th and 17th machines?
How long will it take? Perhaps a priority queue would be a better improvement over a timeout here, to solve this performance issue?

Thoughts?

PR #476, opened 2018-08-23
bff worker settings

bburgin:
Migrated worker .settings file to bff format, so the user can modify fields.
Command line arguments may be insufficient in the future, since OSes have a command line char limit. This change allows future lengthy field values to be passed in via .settings file instead of command line.


neo2buha:
This certainly looks interesting.
I wonder if there is a way to specify the number of cpu's a worker can use per job.
Or to put it differently the number of jobs that can be scheduled on a worker.

ffulin:
I'm on the fence about this change. 

It certainly provides a level of extensibility to the worker options that could come in handy, but I'm concerned about:
 - Use of the NodeGraph in the worker is not something it was ever designed for and I think that will get messy
 - The BFF is a very complex parsing system for what is needed in the worker. While the flexibility provides some power, I worry about the implications of the .settings file having loops and conditionals, importing env vars and including other files

Conceptual concerns aside, I'm also worried about some of the implementation specifics:
 - This is a breaking change that will discard existing settings for all users
 - It seems you can define other node types in the settings without error, and you can define WorkerSettingsNode in the bff used by fbuild.exe
 - Some things that used to persist from the ui (like minimization state) are no longer saved
 - The serialization back to textseems to duplicate the functionality of FunctionPrint()
 - The PR includes changes not directly related to this functionality (like documentation/cmd line help for features in other pull requests)

bburgin:
Would an rsp file format be better than bff format?
In other words, the .settings file could contain the command line args that a user would have passed.
Similar to:
https://docs.microsoft.com/en-us/windows/desktop/midl/the-response-file-command
Then the worker options class could parse the file contents the same as it parses the command line args. We would just need to write some logic to serialize the rsp file format.
Thoughts?

ffulin:
Adding support for response files seems reasonable. I think saving back over the top of the rsp if the user changes the settings would probably be fairly unexpected though. I think we'd want to continue to keep only user modified (via the UI) settings saved to the .settings file. If the .rsp file overrides the saved settings, like the existing cmd line options do, would that give you the control you need, or is there some case that doesn't handle well?

bburgin:
I agree. We should not save over the user's input .rsp file. To clarify, we could:

1. Allow the user to pass in command line args to the worker, as they do today; but also allow them to pass in an .rsp file path. Value reads would proceed like this in the worker:
read .settings file -> read .rsp file -> handle other command line args -> handle GUI operations
With the items to the right overriding values from the left

2. When saving the settings, the worker would save them to the .settings file, but in an .rsp file format instead of a binary format. This way, the user could easily view the contents of the .settings file, directly adjust the file contents, resave the file, and double click the worker exe to launch it. They would not need to pass in any command line args or adjust the GUI, to get the settings they wanted.

Thoughts? I am open to input on this proposal.
Perhaps point 2. could be skipped entirely? But then the user would be forced to use command line args or GUI operations to change their settings. They would not get the convenience of editing the .settings file and then double-clicking the worker exe to launch it.

bburgin:
Abandoning this change. For now, we'll just have the user pass in cmd line args to the worker, rather than edit .settings file.

PR #477, opened 2018-08-23
Error formatting improvements

bburgin:
1. Made error formatting more consistent, while preserving decimal representation of the error code
2. Added more context information to some errors


ffulin:
Thanks for the changes - the errors were pretty messy. I took a look at the places emitting errors and realized we need something a bit smarter than just emitting a single number. The range of error values varies widely and differs between platforms, so the desired format needs to be somewhat contextual.

I ended up making a change (https://github.com/fastbuild/fastbuild/commit/7181d4f0296aa48e64ac2dfdd996691cfb6bf7b4) that uses a helper to print errors in both integer and hex formats, so that whichever is most useful can be clearly seen. This change will be in the next release (v0.98).

An example of an NTSTATUS-style code (failing to spawn a process on Windows due to a missing dll) will show:
```C++
3221225781 (0xC0000135)```
While a smaller error code (like Windows ERROR_FILE_NOT_FOUND, or most OSX/Linux errors) will show:
```C++
2 (0x02)```

Now that the formatting is centralized, we can extend or improve this formatting logic if we find a better layout.

I also applied a consistency pass to the messages based on your CL (with some small modifications).

Let me know what you think! (I'll close this PR as I think I've covered everything, but let me know if I missed something).

bburgin:
Looks good. Thanks.

PR #479, opened 2018-08-23
Make the dir creation thread safe

bburgin:
Handle the case where multiple processes or threads are racing
to create the same dir

### DIFF
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -534,7 +534,10 @@
             // create this level
             if ( DirectoryCreate( pathCopy ) == false )
             {
-                return false; // something went wrong
+                // something went wrong, but continue
+                // so we handle the case where
+                // multiple processes or threads are racing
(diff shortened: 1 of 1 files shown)


Code/Core/FileIO/FileIO.cpp
```diff
@@ -534,7 +534,10 @@
             // create this level
             if ( DirectoryCreate( pathCopy ) == false )
             {
-                return false; // something went wrong
+                // something went wrong, but continue
+                // so we handle the case where
+                // multiple processes or threads are racing
+                // to create the same dir
```
ffulin: DirectoryCreate should (unless there is a bug) already handle the case where someone else already created the directory. Ignoring the return value here when caused by another error is also problematic.

What was the case you saw that resulted in this change?

bburgin: Thanks for pointing out the check in DirectoryCreate(). This pull request is not needed. I am closing it.

PR #480, opened 2018-08-23
Path fixes

bburgin:
1. In IsFullPath(), handle the unc path case
2. In ObjectNode's precompiled headers logic, handle the case where a folder name contains a period. So need to FindLast slash and then find first period. Also, included code to handle the else cases of these find operations.

### DIFF
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -26,17 +26,14 @@
     return false;
 }
 
-// IsfullPath
+// IsFullPath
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for the fixes. I've integrated them and they will be in the next release (v0.98). I extended the tests to cover the case with a period in the dirname (https://github.com/fastbuild/fastbuild/commit/3499907e9c8b7c6db684314b6abe44dbacfc5918). While adding that test I saw that a period in the filename could also confuse things, so I reworked the fix a little cover that case as well.

PR #481, opened 2018-08-23
Line ending fix for include parser test

bburgin:
Handle both Windows and non-Windows line endings in the include parser test


ffulin:
Thanks. For completeness, I never noticed this because i use Perforce and it's setup in such a way that the line endings are respected on all platforms when syncing. This is a good fix to ensure we're not reliant on that.

PR #487, opened 2018-09-09
Fix null deference in MetaNone

dummyunit:
This is small refactor to get rid of null pointer derefence in `MetaNone` function. Albeit that dereference is harmless (at least for now) it's still ugly.

`MetaNone` function is replaced with `MetaNone` type, this allows us to keepthe current syntax and specialize functions that take `IMetaData&` for this new type. New overloads produce the same results as the old ones did for a reference-to-null argument.
Despite the comment in the old `MetaNone()` implementation `MetaNone()` wasn't used with `operator+`, so it doesn't need overloads for `MetaNone` type.


ffulin:
It's nice to get rid of this uglyness :) Thanks!

PR #488, opened 2018-09-09
Use offsetof() macro instead of getting offset by dereferencing null pointer

dummyunit:
New code still has a UB - `offsetof()` can be use only with standard layout types and `ReflectionInfo` is not a standard layout type. But there are some technical benefits to this change:
1. This fixes `-Winvalid-offsetof` warnings in old GCC versions.
2. This fixes UBSan warnings about dereferencing a null pointer, which helps to spot other problems in the output.
3. Since C++17 using `offsetof()` with non-standard layout types is "conditionally-supported", so there is a chance that on some compilers this will become legal eventually.

### DIFF
--- a/Code/Core/Reflection/PropertyType.h
+++ b/Code/Core/Reflection/PropertyType.h
@@ -16,6 +16,7 @@ class Mat44;
 class RefObject;
 template< class T > class Ref;
 template< class T > class WeakRef;
+template< class T > class Array;
 
 // PropertyType
 //------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


ffulin:
Thanks for the cleanup! This code will be in the next release (v0.97).

For some context: the original reason this nullptr deref existed, is that it allows the offset and the type to be passed through as a single param to the AddProperty* functions. This reduces the code gen size for every reflected property. In other code bases where this pattern was used, that was a big deal (4 or 8 bytes saved per reflected property added up to hundreds of KiB, or even MiB). In the case of FASTBuild, we don't have that many exposed properties, and the benefit of this change (improved code quality/analysis ability) outweighs the 2 KiB executable size increase.

PR #490, opened 2018-09-16
Fix build in GCC 7

dummyunit:
This fixes places that trigger new `-Wimplicit-fallthrough` warning from GCC 7.

I am opening a new PR for this because similar fixes from #430 doesn't fix all problematic places that are currently in the code.


ffulin:
Thanks for the cleanup. These changes will be in the next release (v0.97)

The missing return is actually a pre-emptive bug fix for the next version too! (Would have emitted and error but then continued parsing the bff in a likely unpredictable manner)

PR #491, opened 2018-09-16
Fix memory leak on node creation error

dummyunit:
This was found with BFFFuzzer and AddresSanitizer.

This leak isn't serious, but it is still sad to see that FASTBuild uses explicit memory management, especially since there is already a smart pointer - `AutoPtr`. With that in mind, would you be interested in patches that will try to fix this by:
1. Enhancing `AutoPtr` to be more "object friendly" (adding `operator->`, etc.).
2. Replacing usage of explicit `FDELETE` with usage of `AutoPtr` to store pointers.
3. Improving ownership transfer semantics by using `AutoPtr` and move semantics. For example `NodeGraph::RegisterNode` would accept `AutoPtr<Node, DeleteDeletor>` as an argument instead of `Node *`.

Or such changes are undesired?

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/Function.cpp
@@ -289,6 +289,7 @@ Function::~Function() = default;
     AStackString<> nameFromMetaData;
     if ( GetNameForNode( nodeGraph, funcStartIter, node->GetReflectionInfoV(), nameFromMetaData ) == false )
     {
+        FDELETE node;
         return false; // GetNameForNode will have emitted an error
     }
     const bool aliasUsedForName = nameFromMetaData.IsEmpty();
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the fix. This will be in the next release (v0.97). As these were both during failed bff parsing, no user-facing harm would come from these leaks, but they are great to fix nonetheless.

With respect to AutoPtr (or smart pointers in general), I generally chose to use them when when several of the following are true:
 - lifetime is complex (changes of ownership throughout lifetime of object for example)
 - performance is not critical (considering both Debug and Release builds)
 - impact of leaks would be high
 - code clarity would be significantly improved (allowing for better structure)
 - code clarity would not be degraded (not having to pass more complex types in many places)

In this case, I don't think any of those criteria are met, so I think it's preferable to stay with the manual pointer management, despite the potential for future errors.

PR #495, opened 2018-10-01
Fix hang during Ctrl+C of FBuild when Windows Clang exits

bburgin:
Issue: Intermittent hang in ReadFile() that happens when doing Ctrl+C on FBuild during a Windows Clang build.
Solution: Duplicate the pipe handles and close FBuild's copy of the handles, that only the child should have open. This prevents the FBuild hang on exit of Windows Clang. By closing FBuild's side of the pipe handles, ReadFile() can properly return after the child process goes away and its handles are closed. I followed this code to get the correct implementation: https://github.com/kovidgoyal/pywin32/blob/master/win32/Demos/pipes/runproc.py


Code/Core/Process/Process.cpp
```diff
@@ -435,14 +458,6 @@ int Process::WaitForExit()
         VERIFY( CloseHandle( GetProcessInfo().hProcess ) );
         VERIFY( CloseHandle( GetProcessInfo().hThread ) );
 
-        if ( !m_SharingHandles && m_RedirectHandles )
-        {
```
ffulin: We still need to call CloseHandle on the duplicate handles, or this will result in handle leaks.

Code/Core/Process/Process.cpp
```diff
@@ -499,14 +514,6 @@ void Process::Detach()
         // cleanup
         VERIFY( CloseHandle( GetProcessInfo().hProcess ) );
         VERIFY( CloseHandle( GetProcessInfo().hThread ) );
-
-        if ( !m_SharingHandles && m_RedirectHandles )
-        {
```
ffulin: As above, we still need to close the handles we've duplicated.

Code/Core/Process/Process.cpp
```diff
@@ -183,23 +186,35 @@ bool Process::Spawn( const char * executable,
...
             else
             {
-                if ( ! CreatePipe( &m_StdOutRead, &m_StdOutWrite, &sa, MEGABYTE ) )
+                if ( ! CreatePipe( &m_StdOutRead, &stdOutWrite, &sa, MEGABYTE ) )
                 {
                     return false;
                 }
-                SetHandleInformation( m_StdOutRead, HANDLE_FLAG_INHERIT, 0 );
+                HANDLE tmp = INVALID_HANDLE_VALUE;
+                DuplicateHandle(GetCurrentProcess(),m_StdOutRead,GetCurrentProcess(),&tmp,0,FALSE,DUPLICATE_SAME_ACCESS);
```
ffulin: It seems like the core of this fix can be kept without the use of DuplicateHandle - i.e. unless I'm missing something, the important part is closing the "other side" of the pipes.  SetHandleInformation can still be used to prevent inheriting the side we keep open.

Re-strucuturing the code a little bit, I think the following would be much clearer. What do you think?
```
    HANDLE stdOutWrite = INVALID_HANDLE_VALUE;
    HANDLE stdErrWrite = INVALID_HANDLE_VALUE;
    HANDLE stdInRead = INVALID_HANDLE_VALUE;

    // Create pipes
    bool ok = true;
    ok &= CreatePipe( &m_StdOutRead, stdOutWrite, &sa, MEGABYTE ) );
    ok &= CreatePipe( &m_StdErrRead, stdErrWrite, &sa, MEGABYTE ) );
    ok &= CreatePipe( &stdInRead, &m_StdInWrite, &sa, MEGABYTE ) );

    // Close the "other" end of the pipes to avoid deadlocks
    if ( stdOutWrite != INVALID_HANDLE_VALUE ) { CloseHandle( stdOutWrite ); }
    if ( stdErrWrite != INVALID_HANDLE_VALUE ) { CloseHandle( stdErrWrite ); }
    if ( stdInRead != INVALID_HANDLE_VALUE ) { CloseHandle( stdInRead ); }

    // Handle failure
    if ( !ok )
    {
        if ( m_StdOutRead != INVALID_HANDLE_VALUE ) { CloseHandle( m_StdOutRead ); }
        if ( m_StdErrRead != INVALID_HANDLE_VALUE ) { CloseHandle( m_StdErrRead ); }
        if ( m_StdInWrite != INVALID_HANDLE_VALUE ) { CloseHandle( m_StdInWrite ); }
        return false;
    }

    // Prevent child inheriting handles
    VERIFY( SetHandleInformation( m_StdOutRead, HANDLE_FLAG_INHERIT, 0 ) );
    VERIFY( SetHandleInformation( m_StdErrRead , HANDLE_FLAG_INHERIT, 0 ) );
    VERIFY( SetHandleInformation( m_StdInWrite, HANDLE_FLAG_INHERIT, 0 ) );
```

ffulin:
This seems like a good change. I think it introduces a handle leak however (I've added a comment inline about that). Additionally, I think we can tidy things up a little while still keeping your fix (I've added a comment inline for that too).

Let me know what you think.

bburgin:
Thanks. I implemented your suggestions.

PR #496, opened 2018-10-01
Use GetProcessTimes() in kill process tree, fixes #71

### LINKED ISSUE #71: Parent scope ^ doesn't work with local variables
I'm not sure if I'm doing this wrong or if this is a bug but following yields an error.
```
.ParentScopeArray = {}
{
	.LocalVariableString 	= 'hello'
	^ParentScopeArray 	+ { .LocalVariableString }
}
```
```
Error:
FASTBuild Error #1026 - Variable '.LocalVariableString' not found for concatention.
		^ParentScopeArray 		+ { .LocalVariableString }
		                  		^
		                  		\--here
```

bburgin:
Check process start time of children against their parent's start time, before descending in kill process algorithm. This fixes the issue where we could kill ourselves or other old processes that had the same parent pid as a process in our tree, due to Windows pid reuse.
Fixes #71

### DIFF
--- a/Code/Core/Process/Process.cpp
+++ b/Code/Core/Process/Process.cpp
@@ -18,7 +18,6 @@
 #include "Core/Tracing/Tracing.h"
 
 #if defined( __WINDOWS__ )
-    #include <windows.h>
     #include <TlHelp32.h>
 #endif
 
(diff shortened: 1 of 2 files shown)


bburgin:
Another fix that I could do is send Ctrl+C to the child processes, instead of killing them. Would this be good? If I worked on this improvement, would such a pull request be accepted?

ffulin:
I'm sorry you spent time working on this fix in parallel to me - I didn't make it clear that I was working on a fix. Sorry about that. I have submitted my fix here: https://github.com/fastbuild/fastbuild/commit/2fccad78ef5f7e5c59633c8d3b3e4acff275de90

For reference, the main differences are:
 - I don't do explicit checks against the main fbuild process id, as this is now impossible due to the additional GetProcessTimes recursion logic
 - I don't track seen processes, as GetProcessTimes makes it impossible to see the same pid twice in a recursion
 - I avoid FILETIME transformations (just store them as u64 and compare them directly)
 - I fixed an existing file handle leak (hSnap)
 - I use lower permissions level in OpenProcess when acquiring the handle for use in GetProcessTimes
 - I added a test of process cancellation (doesn't test the pid re-use per-se, since that's not really practical, but at least serves as a regression test)

If you think I missed any edge-case, please let me know. I spent a lot of time trying to be sure that all possible cases are handled, but (as proved by the original bug) this is surprisingly complex!

bburgin:
No worries. I was able to learn a lot even if my changes were moot.

PR #498, opened 2018-10-02
Some improvements to kill process logic

bburgin:
1. Keep process handles open while doing work, so pids don't change between work steps.
2. Pass 0 for CreateToolhelp32Snapshot( TH32CS_SNAPPROCESS, 0 ), since enumerating all processes not a specific one.
3. For the initial call to KillProcessTreeInternal(), use the child process's creation time, instead of FBuild.exe's creation time so it matches the other calls to KillProcessTreeInternal()
4. Remove a stray set of spaces and colon in the // Get process start time  comment
5. Added colon in front of CloseHandle( hSnap ) call, so matches the other calls to CloseHandle()

### DIFF
--- a/Code/Core/Process/Process.cpp
+++ b/Code/Core/Process/Process.cpp
@@ -18,6 +18,7 @@
 #include "Core/Tracing/Tracing.h"
 
 #if defined( __WINDOWS__ )
+    #include <windows.h>
     #include <TlHelp32.h>
 #endif
 
@@ -74,7 +75,7 @@ Process::~Process()
 // KillProcessTreeInternal
 //------------------------------------------------------------------------------
 #if defined( __WINDOWS__ )
-    void Process::KillProcessTreeInternal( const HANDLE hProc, const uint32_t processID, const uint64_t processCreationTime )
+    void Process::KillProcessTreeInternal( const void * hProc, const uint32_t processID, const uint64_t processCreationTime )
     {
         PROCESSENTRY32 pe;
 
@@ -102,6 +103,7 @@ Process::~Process()
                     const uint64_t childProcessCreationTime = GetProcessCreationTime( hChildProc );
                     if ( childProcessCreationTime < processCreationTime )
                     {
+                        ::CloseHandle( hChildProc );
                         continue; // Cannot be a child because it was created before the parent
                     }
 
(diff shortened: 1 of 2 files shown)


Code/Core/Process/Process.cpp
```diff
@@ -97,48 +96,47 @@ Process::~Process()
...
-                if ( childProcessCreationTime < processCreationTime )
+                const HANDLE hChildProc = ::OpenProcess( PROCESS_ALL_ACCESS, FALSE, childProcessId );
+                if ( hChildProc )
                 {
-                    continue; // Cannot be a child because it was created before the parent
-                }
+                    const uint64_t childProcessCreationTime = GetProcessCreationTime( hChildProc );
+                    if ( childProcessCreationTime < processCreationTime )
+                    {
+                        continue; // Cannot be a child because it was created before the parent
```
ffulin: hChildProc is being leaked here

Code/Core/Process/Process.h
```diff
@@ -7,6 +7,10 @@
 #include "Core/Env/Types.h"
 #include "Core/Containers/AutoPtr.h"
 
+#if defined( __WINDOWS__ )
+    #include <windows.h>
+#endif
```
ffulin: FASTBuild's core libraries go out of their way to keep windows.h our of the headers to avoid compile time impact.

Code/Core/Process/Process.cpp
```diff
@@ -97,48 +96,47 @@ Process::~Process()
...
     {
-        HANDLE processHandle = ::OpenProcess( PROCESS_QUERY_LIMITED_INFORMATION, FALSE, processId );
-        if ( processHandle == nullptr )
+        if ( hProc == 0 )
         {
-            return 0; // Likely due to lack of permissions
+            return 0;
         }
 
-        // Get process start time                                                                           :
```
ffulin: weird - i wonder how that got there :)

ffulin:
Thanks for the pull request.

**> 1. Keep process handles open while doing work, so pids don't change between work steps.**

I was deliberately opening the processes twice, as the GetProcessTimes requires a lower privilege level. However, on reflection, this was unnecessary, as if we ever failed to GetProcessTimes, we could assume the process was not part of our hierarchy anyway, so we might as well only have one call. You are right to point out that keeping the process handle avoids a pid re-use between GetProcessTimes and the Terminate, so this is a good change.

**> 2. Pass 0 for CreateToolhelp32Snapshot( TH32CS_SNAPPROCESS, 0 ), since enumerating all processes not a specific one.**

This makes sense - the processId was ignored anyway

**> 2. For the initial call to KillProcessTreeInternal(), use the child process's creation time, instead of FBuild.exe's creation time so it matches the other calls to KillProcessTreeInternal()**

Good catch. This should have been safe, but it is more correct to use the child process, and this might save some unnecessary traversal.

**> 3. Remove a stray set of spaces and colon in the // Get process start time comment**

Not sure how this got there :)

**> 4. Added colon in front of CloseHandle( hSnap ) call, so matches the other calls to CloseHandle()**

More consistent. Nice.

**Handle Leak**

This change introduces a pretty severe handle leak (which i've highlighted in the comments), so we'll need to fix that up.

**windows.h**

FASTBuild's core libraries try very hard to keep the massive windows.h header out of headers. I'd like to keep that here. We can pass void * for handles and uint32_t for pids as the rest of the class does. Would you mind restoring this?

ffulin:
Thanks for the improvements. I've integrated this pull request. These changes will be in the next release (v0.97)

PR #501, opened 2018-10-22
Added “AllowResponseFile” setting.

awefers:
Added an AllowResponseFile setting to CompilerNode, LibraryNode, LinkerNode and ObjectNode. The existing autodetect behavior for compilers on Windows will continue to override this setting to avoid issues with existing build configurations.

I needed the setting for custom tools on macOS, where I hit a build issue when exceeding the command line limit. It felt only natural that custom tools should be able to opt in the response file support.


ffulin:
I've integrated this change into the dev branch. This will be part of the next release (v1.02). Thankyou!

awefers:
Thank you!

PR #504, opened 2018-11-12
Close handle assert fix

bburgin:
When running debug FBuildWorker.exe, I noticed some asserts when spawning the FBuildWorker subprocess. To avoid the asserts, added if statements around CloseHandle() calls.

### DIFF
--- a/Code/Core/Process/Process.cpp
+++ b/Code/Core/Process/Process.cpp
@@ -540,9 +540,9 @@ void Process::Detach()
 
     #if defined( __WINDOWS__ )
         // cleanup
-        VERIFY( ::CloseHandle( m_StdOutRead ) );
-        VERIFY( ::CloseHandle( m_StdErrRead ) );
-        VERIFY( ::CloseHandle( m_StdInWrite ) );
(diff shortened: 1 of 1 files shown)


ffulin:
I was only able to reproduce this assert if I delete the FBuildWorker.exe.copy before trying to spawn it. i.e. the launch fails was the only way I could get the handles to be invalid.

Do you have any more information about the situation you saw the assert in?

bburgin:
I saw the assert when launching the x64-Debug FBuildWorker.exe (without any command line args) from the Visual Studio Debugger. The FBuildWorker.exe was in the tmp\x64-Debug\Tools\FBuild\FBuildWorker folder. Before launch, only the FBuildWorker.exe and FBuildWorker.pdb files were present in the folder. So no .copy files. I saw the assert in the FBuildWorker.exe process, not the spawned subprocess.

ffulin:
I was able to find a reliable repro case for this assert firing - replacing the exe while running (triggering the self-reload function) would assert reliably. I've integrated the fix which will be part of the next release (v0.97).  Thanks!

PR #513, opened 2018-12-31
Fix inconsistent parsing of function headers

dummyunit:
Empty function headers (without any characters between parenthesis) were incorrectly accepted for all functions, even for functions that require headers.
For example, this code would correctly produce error #1001:
```
    ObjectList( )
    {
        .Compiler = '/usr/bin/gcc'
        .CompilerOptions = '-c %1 -o %2'
        .CompilerOutputPath = '.'
    }
```
And if the whitespace inside the function header would be removed then there would be no error. Also in Debug build this would lead to a failed assertion in `Function::Commit` because function name would be empty.

This change fixes the problem, but as a side effect specifying empty parenthesis for functions that don't require headers (like `Library`) now results in error #1001. This shouldn't be a problem because these empty parenthesis could be just removed.

Also documentation for `Compiler` and `ObjectList` functions is updated to match code behaviour.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
@@ -19,7 +19,7 @@
 Describes a compiler and specifies how to interact with the compiler process for dependency tracking,
 distribution, caching and more.
       </p>
-<div class='code'>Compiler( 'name' )              // (optional) Alias
+<div class='code'>Compiler( 'name' )              // Alias
 {
   .Executable                   // Primary compiler executable
   .ExtraFiles                   // (optional) Additional files (usually dlls) required by the compiler.
--- a/Code/Tools/FBuild/Documentation/docs/functions/objectlist.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/objectlist.html
@@ -18,7 +18,7 @@
       <p>
 Builds a list of objects, typically for later input into a Library, Executable or DLL.
       </p>
-<div class='code'>ObjectList( alias )         ; (optional) Alias
+<div class='code'>ObjectList( alias )         ; Alias
 {
   ; options for compilation
   .Compiler                 ; Compiler to use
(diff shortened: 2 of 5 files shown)


ffulin:
Nice find (as usual :) ). 

This seems like a good fix in general, but I think we need to not break backwards compatibility with existing bffs. While it's easy enough to fixup the few cases in the bff source, we shouldn't make a breaking change unless we have to and unless I'm missing something, I think we can support functions that don't need args in a way that allows empty () or omitting the () entirely.

dummyunit:
> I think we can support functions that don't need args in a way that allows empty () or omitting the () entirely.

Done, now functions for which `NeedHeader() == false` don't emit error 1001 when the header is empty.

Or rather when the header doesn't start with a string, which brings the following problem:
Previously parser accepted constructs like `ObjectList('TheAlias' any properly "quoted" and/or (parenthesized){junk})` which is weird but harmless. But now with the alias string being optional it will silently accept stuff like `Library(TargetName)` (note missing quotes) which may bring some confusion ("why the alias is not recognized?").
I think this weirdness should be fixed and parser should check that there is only whitespace besides the string in the header. What do you think?

Also I noticed that while `BFFIterator` has overloads for operators `<` and `>` there are no overloads for `<=`, `>=`, `==` and `!=` and in a lot of places results of `GetCurrent()` are compared instead of the iterators. Would you be interested in a PR that would fix this? (add missing overloads and put them to use)

dummyunit:
I have added the check for unexpected tokens in the header, now FASTBuild will emit error 1002 for them:
```
test.bff(1,10): FASTBuild Error #1002 - Library() - Matching closing token ) not found.
Library( Name )
         ^
         \--here
```
The message is not great though, it could probably mention that an optional string was expected. But this is what functions `If` and `Error` do in similar cases.

ffulin:
Thanks for the fix. This has been integrated and will be in the next release (v0.98).

PR #514, opened 2018-12-31
Improve Test function logic

dummyunit:
Display and save the test output when test times out. This allows to determine which part of the test is most likely responsible for the timeout.

Show "Test failed" message before trying to save the output because that may fail as well and in that case the former message wouldn't be displayed at all.


ffulin:
Printing the output when the test times out seems like a good improvement.

I think we should keep the "Test failed" message after the general test output however. I think we could print it before trying to save the output so that it will be emitted even if that fails.

dummyunit:
Sorry, didn't pay attention to the order in which messages are displayed. Should be fixed now.

ffulin:
Thanks for the changes. This has been integrated and will be in the next release (v0.98).

PR #515, opened 2018-12-31
Fix array length calculation when adding a string to array of strings

dummyunit:
Hint for the size of the resulting array was taken from the RHS string variable instead of LHS array of strings variable.


ffulin:
Good find.

PR #516, opened 2018-12-31
Call NodeGraph::FindNodeExact instead of FindNode when using clean path

dummyunit:
This is a minor optimization for BFF parsing. `CleanPath` function is somewhat complex and not calling it unnecessarily improves performance slightly (by about 5% in my case).

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
@@ -1178,7 +1178,7 @@ void LinkerNode::GetImportLibName( const AString & args, AString & importLibName
     NodeGraph::CleanPath( potentialNodeName, potentialNodeNameClean );
 
     // see if a node already exists
-    Node * node = nodeGraph.FindNode( potentialNodeNameClean );
+    Node * node = nodeGraph.FindNodeExact( potentialNodeNameClean );
     if ( node )
     {
         // aliases not supported - must point to something that provides a file


ffulin:
This seems like a good little optimization.

I'm surprised to hear you say you saw GetOtherLibraries/GetOtherLibrary showing up as a significant portion of bff parsing though. Do you have a lot of executables or dlls? Or perhaps lots of configurations? Or perhaps a lot of third party libs?  Or maybe all of these things? :)

dummyunit:
Yes, all of these: 8 to 11 `/LIBPATH` options, 6 exe/dll with about 40 `.lib` files in `.LinkerOptions`, 18 configurations ( {VS2013, VS2015, VS2017} * {Win32, Win64, Win32XP} * {Debug, Release} ).

I was investigating why sometimes argument completion in the console was slower than desirable (>1sec). It turned out that it was slow when bff reparsing was required and most of the time was spent in WinAPI's `GetFileAttributes` called from `FileIO::FileExists` called from `LinkerNode::GetOtherLibrary`.
So I wasn't able to fix the slowness, but noticed in the profile that `ClearPath` was often called 2 times in a row.

ffulin:
If on average we check 5 LIBPATHS before finding the lib in question, it does quickly multiply to ~20,000 or so FileExists checks, so that would definitely add up more quickly that one might imagine.

I took a look at the GetOtherLibraries code to see if there are opportunities for quick improvements, and I didn't see anything obvious. In the long-term, it might make sense for this work to happen at build time instead of bff parse time, but we'd need to carefully consider the implications of that.

General bff parsing improvement (outside of this specific bit of code) is something that is on the backlog: there are some huge inefficiencies, but they will need large changes to eliminate.

One thing I did notice though, is that if you use -showtargets, after the bff is parsed if it changed, it doesn't seem to be saved, which means you pay the cost of parsing every time until a build is performed. Similarly, passing an invalid target to fbuild doesn't save the bff after parsing. If you are using the fbuild.bash-completion script, this might be being impacted by this first case at least. I think both of these would be a fairly easily fixed (likely even the same fix).

In the short term, for your specific use-case, I don't know what options you have open to you to change your bff files, but it might be possible to speed things up by:

**a) Switching to working-dir relative paths instead of using /LIBPATHs.** This would give huge reduction in FileExists calls, since you'd only call it once in total for every unique .lib. Of course this might be impractical for some reason (like if your bff files are generated).

**b) Shuffle the order of your /LIBPATHs.** If the number of .lib files contained in certain paths is skewed, listing those /LIBPATHs first might yield some improvment. This could be a good saving and would be a lot less intrusive so potentially workable.

I hope that helps - let me know!

ffulin:
I've integrated this optimization so that it will be in the next release (v0.97). Thanks as always!

PR #517, opened 2018-12-31
Fix data race on a Job object in the case when remote job finishes first

dummyunit:
This is a fix for #447 by @samhocevar, I have tested it and can confirm that it resolves the issue.
I am opening PR myself because there is no activity on the issue so far and it would be nice to get it fixed.


ffulin:
I've integrated this fix which will be in the next release (v0.98). Thankyou both for the fix!

PR #519, opened 2019-01-06
Fix logging output from successful remote jobs

LeoMash:
This patch makes log of compilation which contains warnings more determenistic, not relying on what jobs were completed locally and remotely.
Previously if translation unit has compilation warnings and was compiled remotely - you won't see them on master, but you will if compilation job completed locally.
Sometimes users can't fix or disable all warnings, but need to see them (for example on CI), so I made a patch to fix this.


ffulin:

I made a few additional changes (https://github.com/fastbuild/fastbuild/commit/1dad742d5c385cf7a24ab8523a4629e500c44f99):
 - I added a test to check that remote warnings are shown
 - I modified the logic so remote output is only shown if it contains warnings (rather than always) which makes it consistent with local behavior
 - I added equivalent logic to show local and remote warnings for GCC and Clang (previously only MSVC warnings would be shown locally)

PR #520, opened 2019-01-08
Optional .Environment variables for compiler, librarian, and linker

bburgin:
The variables allow a user to run multiple msvc compilers in one FBuild call, improving build performance.
The variables are needed because the Settings node .Environment is insufficient: it can only work for one msvc compiler PATH at a time.


ffulin:
Nice one - looks like a good enhancement.

bolfran001:
I think it could be useful in the `TestNode` also.

bburgin:
> 
> 
> I think it could be useful in the `TestNode` also.

All of my tests use command line args rather than environment variables. Can you give an example of a test that uses environment variables?

bolfran001:
Some tests are linking with third-party librairies (boost, etc.) and modifying PATH variable could be useful for finding DLLs when those tests are executed.
Currently, we hardlink those DLLs in the folder where the tests will be compiled to able to execute them (or maybe there is a better way to that...) but with this feature, we could avoid that hardlink "pre-step".

bburgin:
> 
> 
> Some tests are linking with third-party librairies (boost, etc.) and modifying PATH variable could be useful for finding DLLs when those tests are executed.
> Currently, we hardlink those DLLs in the folder where the tests will be compiled to able to execute them (or maybe there is a better way to that...) but with this feature, we could avoid that hardlink "pre-step".

Thanks. I agree that environment variables would help you run these tests, for the local case. But how would you run the tests distributed? To remote tests, you must specify .ExtraFiles that are absolute paths or paths relative to the test exe. Would you be able to remove your hard linking and reference your .dlls with absolute paths or paths relative to the test exe?

bolfran001:
Oh yes, you're right, I didn't think about distribution. I guess yes but I won't do that now. Thanks for the hint!

ffulin:
I think it's ok for this .Environment setting to only apply to local compilation. That's already the case for the .Environment variables set "globally" in Settings() that this overrides. We just need to clearly document the behavior.

ffulin:
Thanks for the changes! I've integrated them and they will be in the next release (v0.98).

I made some followup changes (https://github.com/fastbuild/fastbuild/commit/02484a980d35640d17c183d9167fd69a36d32195):
 - moved the code to handle chosing an env var into the base Node class to avoid duplicating logic
 - created a shared mutex to avoid increased per-node memory usage (we actually only this mutex for Compiler(), but for simplicity they all use it)

PR #525, opened 2019-01-21
Use wordexp to tokenize argument string on Linux/OSX. 

### LINKED ISSUE #510: Process argument tokenization on OSX
Hi!

I've encountered an issue in v0.96 where clang will fail to produce objects due to missing files included by the code being compiled. Running clang in verbose mode, it complains about the include paths I'm specifying:

`ignoring nonexistent directory ""/my/include/path""`

I'm adding the include path like this:

`-I"/my/include/path"`

The path does exist, before I go any further :). In fact, I can copy the command printed by -showcmds and run it fine in the terminal. It seems to be that the arguments which clang receives differ to the ones that FASTBuild says that it is sending. In the message that clang prints above, my include path is being enclosed in an additional set of quotes. Also, when running clang in verbose mode, the include path argument is being shown as this:

`-I "\"/my/include/path\"`

If I remove the quotes from around the include path in the .bff, FASTBuild will happily compile my code. However, this is due to the fact that there aren't any spaces in any of my include directory paths. It seems that clang can't handle an include path in quotes being passed as a single token in argv.

So as a temporary "hack" to fix things for myself, I've modified Process::Spawn to find the first and last quotes (if any) in the argument, and removing them (instead of just doing the BeginsWith/EndsWith checks).

However, I'm not sure if this is just one problem I've encountered which I've managed to patch up, and so there may be additional issues with how the arguments string is tokenized compared to how the system does it, which is resulting in the difference in behaviour.

By the way, I have noticed that this seems to be something which FASTBuild's own bff file appears to work around for Linux and OSX? I should note that I'm working with a UE4 project, and am therefore not writing the BFF by hand, but am just translating the build actions to the BFF.

Hope this helps give enough info about the issue. :)

Thanks,
Nick

NEdwards-SumoDigital:
Should perform shell-like expansion of the args string to more reliably produce a correct argv/argc.

Seems to resolve #510, although further testing would be needed to verify this.

### DIFF
--- a/Code/Core/Process/Process.cpp
+++ b/Code/Core/Process/Process.cpp
@@ -31,6 +31,7 @@
     #include <string.h>
     #include <sys/wait.h>
     #include <unistd.h>
+    #include <wordexp.h>
 #endif
 
 // Static Data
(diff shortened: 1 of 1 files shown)


ffulin:
I took a look at your changes and I think that while this fixes the problem, it introduces a couple of new problems:

---

*a) It performs substitutions:*
 - it expands ~ to the home dir
 - it replace environment variables
 - it does command expansion
 - wildcard expansion via directory listings

Many/all of these could change between FASTBuild runs in a way that FASTBuild would not be aware of, resulting in incorrect build behaviour. A build would not trigger if your cmd line referenced an environment variable that changed value for example. The wildcard expansion of directory listings additionally adds the ability to pass files to commands in a way that would at first seem to work, but would lack all dependency tracking.

I see that wordexp() takes some flags, but it's not clear from the documentation if you can disable all these behaviors. It sort of looks like you can only disable some of them.

*b) It will break without escaping characters first*

Many characters need to be escaped otherwise they will not be handled correctly by wordexp. i.e. you can pass a string like "a | b <> c" right now, but wordexp would consider those characters to have special meaning.

Similarly the math expression expansion could cause rather surprising behavior.

---

It might be worth trying to see if the behaviour can be sensibly constrained to just the quote parsing before giving up on this, but I would suspect that's tied to the way escaping in general is handled.

If it can be made to work, I'd also be a little concerned about the performance of this function, so we'd need to check that too.

I think it's quite likely we won't be able to use this change and will probably need to implement custom parsing that handles the " cases as expected without all the functionality we don't want.

ffulin:
I think this tries to address an important issue that does need a fix, but I think this PR can't be used as it stands for the reasons I mention above. With that in mind, I think it's best to close this PR for now.

As noted before, my gut tells me we'll need to smarten up the existing code to correctly handle the " cases that are problematic.

PR #526, opened 2019-01-21
Remove isysroot, which may not exist on a distributed system

NEdwards-SumoDigital:
-isysroot should only be used for include paths, which will have already been processed in the preprocessor pass.

The presence of this option can cause remote builds to fail if the directory doesn't exist on the machine.


ffulin:
Thanks for the fix! This has been integrated into the mainline and will be in the next release (v0.98)

PR #527, opened 2019-01-21
Handle WM_QUIT message

bburgin:
This change allows me to programmatically ask the worker to stop. I plan to use a powershell script to stop and relaunch a worker with new command line args, if my user provides different args from the last time they ran the script.


ffulin:
Thanks for the improvement! This change has been integrated and will be in the next release (v0.98).

PR #528, opened 2019-01-22
Allow specifying brokerage path via bff and command line

NEdwards-SumoDigital:
Eases setup as we don't have to mess around setting environment variables. Command line option is highest priority setting, then bff, then env variable. Possibly not the most elegant implementation, but tries to account for the fact that both FBuild and FBuildWorker need to be able to pass along their command line brokerage paths.


Code/Tools/FBuild/FBuildCore/Graph/SettingsNode.cpp
```diff
@@ -81,30 +82,38 @@ SettingsNode::~SettingsNode() = default;
...
-    // Settings() bff option overrides environment variable
-    if ( m_CachePath.IsEmpty() == false )
+    // Environment variable takes priority
+    if ( m_CachePathFromEnvVar.IsEmpty() == false )
     {
-        return m_CachePath;
+        return m_CachePathFromEnvVar;
     }
-    return m_CachePathFromEnvVar;
+    return m_CachePath;
```
ffulin: The current order of precedence is the documented and expected behavior that many users rely on, so this can't be changed. Any new variables should behave the same way for consistency.

Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
```diff
@@ -45,9 +46,16 @@ void WorkerBrokerage::Init()
...
-    // root folder
+    // root folder - command line, then bff setting, then env var
     AStackString<> root;
-    if ( Env::GetEnvVariable( "FASTBUILD_BROKERAGE_PATH", root ) )
+    if ( !m_CmdlineBrokeragePath.IsEmpty() )
+        root = m_CmdlineBrokeragePath;
+    else if ( !settingsBrokeragePath.IsEmpty() )
+        root = settingsBrokeragePath;
+    else if ( Env::GetEnvVariable( "FASTBUILD_BROKERAGE_PATH", root ) )
+        ((void)0); // Nothing, root is populated
```
ffulin: It would be betterto manage the priority in the SettingsNode, using the same pattern and order of precedence as other options and only pass in the final value to use into this function.

Code/Tools/FBuild/FBuildCore/Graph/SettingsNode.cpp
NEdwards-SumoDigital: This was the result of a bad merge, I think. Oops, sorry!

Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
NEdwards-SumoDigital: I had considering doing that, but decided to go with this approach to eliminate duplicate code for FBuild and FBuildWorker, as the worker won't have a SettingsNode (afaik)

bburgin: Typically, override logic lives in the FBuildApp and FBuildWorker Main.cpp files.
After choosing a value, the logic sets it back on the corresponding Settings node, using set methods.
This pattern works well with the existing SettingsNode and a future WorkerSettingsNode.
I added a WorkerSettingsNode in my pull request here:
https://github.com/fastbuild/fastbuild/pull/476

If you are worried about code duplication in the Main.cpp files, you could make a helper method in Env.h / Env.cpp to do the env fallback logic in a generic way, for any variable.
Then you could call it in the Main.cpp files, for this and other env variables, where needed.

ffulin: I not convinced that using the bff/node serialization infrastructure for the Worker is the right way to go, so I wouldn't go building on top of that for this change.

I think it'd be fine to create a helper function common to FBuildWorker and FBuild if it avoid some duplication. I think FBuildCore/Helpers would be the right place for that (Core/Env should remain FASTBuild agnostic - i.e. no FASTBuild specific functionality)

Most importantly, we need to make sure the order of precedence for any new env vars we add is consistent with those that already exist.

PR #529, opened 2019-01-24
Fixed NodeGraph::CleanPath for paths that end with ".." (no trailing slash)

bkslasher:


ffulin:
Thankyou for the fix! This has been integrated and will be in the next release (v0.98). I've also added a regression test for this case (https://github.com/fastbuild/fastbuild/commit/e93fd0deaa942bc094842b50638eb06f1db58389)

PR #533, opened 2019-02-03
Fix self assignment for arrays

dummyunit:
Previously attempt to assign an array variable to itself in the current scope triggered an assert in `Array::operator=` in Debug builds and cleared the variable in Release builds. This change fixes this surprising behaviour for arrays.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFParser.cpp
@@ -1739,6 +1739,14 @@ bool BFFParser::StoreVariableToVariable( const AString & dstName, BFFIterator & 
             return false;
         }
     }
+    else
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the fix! This has been integrated and will be in the next release (v0.98). Regression tests have also been added (https://github.com/fastbuild/fastbuild/commit/0e8c45bccb634ef97b59107433462350174a2758)

ffulin:
I found out that the existing code was relying on continuing through the function in the case where a variable in a lower scope needed to be created to shadow one at the higher level. I've added a regression tests for that and extended your fix so that it doesn't prevent that case (https://github.com/fastbuild/fastbuild/commit/464c23332e92c3478d6ea24836213775161eac86).

I hope to overhaul the bff parsing in the future to make it less fragile .

PR #534, opened 2019-02-04
Fixed segfault when compiling with Apple LLVM 10.0

bkslasher:
When FASTBuild is built using the latest Xcode 10.1 (Apple LLVM version 10.0.0 (clang-1000.11.45.5)), I encounter segfaults EXC_BAD_ACCESS(code=EXC_I386_GPFLT) when some objects are dynamically allocated. For example at `FileIO::FileInfo * fi = FNEW( FileIO::FileInfo() );` inside UnityNode.cpp.
screen shot 2019-02-04 at 5 02 16 pm
I think it is caused by LLVM [bug 35901](https://bugs.llvm.org/show_bug.cgi?id=35901). Clang generates a `movaps` instruction, which expects 16 byte alignment.
Perhaps these changes should be wrapped by `#ifdef __clang__`, since it seems to be a problem only for Clang.

### DIFF
--- a/Code/Core/Mem/Mem.cpp
+++ b/Code/Core/Mem/Mem.cpp
@@ -13,6 +13,7 @@
 #include "Core/Mem/SmallBlockAllocator.h"
 
 #include <stdlib.h>
+#include <cstddef>
 
 // Defines
 //------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for the report. I agree that this looks like a Clang bug and does appear to be the same issue you linked. I also saw some other similar bugs reporting misuse of movaps incorrectly.

Given that this will increase memory usage, I think we should do as you suggested and wrap the change with ```#if defined( __clang__ )```

ffulin:
I've integrated your fix. I had some strange compile errors with Clang on Windows when including cstddef, so I removed that and explicitly set the alignment to 16. This will be part of the next source code release (v0.98).

Thanks for reporting and fixing this issue!

(follow up commit: https://github.com/fastbuild/fastbuild/commit/8905e8d209848fb1003ff50c73f174c20573d757)

PR #537, opened 2019-02-14
Added .CompilerOutputKeepBaseExtension to ObjectList

bkslasher:
I have a few cases where multiple files in the same directory have the same name but different extensions. Normally when such files are in the same ObjectList there is a conflict. For example `file.ext1` and `file.ext2` both produce `file.o`. This option allows to keeps an original extension in output files, so there is no conflict. For example `file.ext1` produces `file.ext1.o`, `file.ext2` produces `file.ext2.o`.


ffulin:
Thanks for the improvement. I've integrated it, and it will be in the next release (v0.98).

PR #538, opened 2019-02-21
Allow targets to be hidden from -showtargets list

Filippok:
It's useful to hide some targets from -showtargets list. 
For example, here is list of targets:
- Foo-Bar-c
- Foo-Bar-cpp
- Foo-Bar-asm
- Foo-Bar-exe

I don't need information about anything except Foo-Bar-exe, so now I can hide other targets from
default output, but when I need full list of targets, I can specify -showhidden flag and see full list of targets.

This commit changes:
- Added .Hidden option to Alias, Unity and ObjectList. Default == false
- Disabled showing hidden targets in -showtargets output by default
- Added -showhidden option to allow displaying hidden targets.

### DIFF
--- a/Code/Tools/FBuild/FBuildApp/Main.cpp
+++ b/Code/Tools/FBuild/FBuildApp/Main.cpp
@@ -182,7 +182,7 @@ int Main(int argc, char * argv[])
 
     if ( options.m_DisplayTargetList )
     {
-        fBuild.DisplayTargetList();
+        fBuild.DisplayTargetList( options.m_ShowHiddenTargets );
         ctrlCHandler.DeregisterHandler(); // Ensure this happens before FBuild is destroyed
         return FBUILD_OK;
     }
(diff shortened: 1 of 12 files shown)


ffulin:
I was not able to get git to co-operate merging this PR or resolving the conflicts, so I've integrated it manually (https://github.com/fastbuild/fastbuild/commit/9b0d82187276118eb5dbdca66d03934dd84dcd45). This change will be in the next release (v0.99)

I made some small changes:
 - "showhiddentargets" is renamed to "showalltargets" and works by itself (doesn't need to be coupled with -showalltargets)
 - I added documentation for the new command line options
 - I moved the m_Hidden bool in Node to avoid class size increase

ffulin:
Documentation and Integration files updated: https://github.com/fastbuild/fastbuild/commit/8c9efc84392f76be2047a77cd7d0ac2128839959

dummyunit:
It seems that during integration of these changes version of the node graph wasn't bumped. So now builds from the dev branch are crashing when they read an `.fdb` from the release version.

ffulin:
.fdb version has been updated: https://github.com/fastbuild/fastbuild/commit/31c63c45b0c1c67c72f6b2978b24db31d6eeb3ef

PR #546, opened 2019-04-16
Fixed assert firing with orbis-wave-psslc.

krolli:
Hello, while working on a different feature, I ran into an assert triggered when compiling using orbis-wave-psslc. I believe this assert is just missing check for this compiler family, as orbis-wave-psslc uses -E as "run pre-processor only" flag.


ffulin:
Thanks for the fix!

PR #548, opened 2019-05-01
Microsoft Direct3D shader compiler (fxc.exe) support

krolli:
This change adds support for new compiler family based on Microsoft Direct3D shader compiler fxc.exe . Majority of code is similar to other compiler families. One major difference compared to other families is the way dependencies are obtained.

When compiling a shader, compiler writes results of preprocessing to a file and writes include information to stdout. Includes and dependency information are determined based on stdout, as preprocessed file has paths always in form in which they were written in original code, which means they can be always relative to different directory and I couldn't figure out a way to reconstruct correct paths in all cases.

There are 2 outstanding issues I know of, though there might be more after reviewing changes:
1. Preprocessing output path is compilation output path with .i suffix appended and some places assume that is where the file is located when they begin final compilation. I suspect this would run into issues with distributing build, but I haven't been able to test it.
2. Job data is probably not set correctly. Again, this might cause issues with distributed build.

I am primarily looking for feedback whether there is interest in this change. We definitely needed it for our work, and it would simplify things to have it be part of main fastbuild distribution. I welcome any information on what should be fixed/changed/tested.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
@@ -137,6 +137,7 @@ Compiler( 'Test' )
         <tr><td>qt-rcc</td><td>Qt's resource compiler</td></tr>
         <tr><td>vbcc</td><td>vbcc compiler</td></tr>
         <tr><td>orbis-wave-psslc</td><td>orbis wave psslc shader compiler</td></tr>
+        <tr><td>fxc</td><td>Microsoft Direct3D shader compiler</td></tr>
         <tr><td>&nbsp;</td><td></td></tr>
         <tr><th>Value</th><th>Notes</th></tr>
         <tr><td>custom</td><td>Any custom compiler. <font color=red>NOTE:</font> Only primary input and output dependencies will be tracked. No additional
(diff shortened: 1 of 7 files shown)


claimred:
> We definitely needed it for our work

Could you please elaborate on that? We've been using fxc.exe as a generic custom compiler for a while, seemingly no problem so far. Perhaps we are missing something?

krolli:
I assume by "generic custom compiler" you mean setting `.CompilerFamily = 'custom'` which will not track header dependencies. If you're fine with that, then you probably don't need any of this. We were not. I'm not aware of any existing family that matches fxc way of doing things.

That said, I am considering withdrawing this PR and instead writing command line front-end to `d3dcompiler_*.dll` as well as newer `dxcompiler.dll` that will make it use the same interface as one of the existing families, since we've run into problems with `dxc.exe` that make it incompatible with `fxc.exe` anyway. It just seems more sensible to have something that works with any C++-compatible build system than making changes in specific one. Note, I haven't yet searched for existing project that would be doing this already.

claimred:
> which will not track header dependencies. If you're fine with that, then you probably don't need any of this.

Oh, I see, thanks for clarifying! It roughly looks like https://github.com/fastbuild/fastbuild/issues/690 for rc.exe. I guess it's not much of an issue for us but still.

PR #551, opened 2019-05-04
Add missing headers

dummyunit:
* `memory.h` is needed for `memset`, without it non-unity build fails on Linux.
* `fcntl.h` is needed for `open`, without it non-unity build fails on Linux when musl is used.


ffulin:
Thanks :)

PR #552, opened 2019-05-04
Remove definition of DWORD that differs from the one in Windows.h

dummyunit:
This fixes unity build on Windows in the cases when CRC32.cpp is included in the unity before Env.cpp. When that happens compilation fails on line 302 in Env.cpp with error C2664:
```
'BOOL` GetUserNameA(LPSTR, LPDWORD)': cannot convert argument 2 from 'unsigned int *' to 'LPDWORD'
```


ffulin:
Thanks :)

PR #553, opened 2019-05-05
Fix name of the dependency graph file on Linux

dummyunit:
Macro `__LINUX__` was misspelled and as the result node graph on Linux was stored in the file `fbuild` instead of `fbuild.linux.fdb`.

I also would like to propose a way to prevent such problems (or at least reduce probability) in the future by using function-like macros for platform checks. That way if the name of the macro will be misspelled compiler would produce an error about unknown function-like macro.  
In detail I propose:
1. Replacing checks like `#if defined( __WINDOWS__ )` with `#if __WINDOWS__()` (or even better: `#if FBUILD_WINDOWS()` to stop using reserved identifiers).
2. Replacing compiler flags like `-D__WINDOWS__` with `-D__WINDOWS__()=1 -D__LINUX__()=0 -D__OSX__()=0`.

Existing check of the form `#else` `#error Unknown platform` can prevent such errors, but only if it is inserted in each new place where platform defines are checked.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/FBuild.cpp
+++ b/Code/Tools/FBuild/FBuildCore/FBuild.cpp
@@ -155,7 +155,7 @@ bool FBuild::Initialize( const char * nodeGraphDBFile )
             m_DependencyGraphFile += ".windows.fdb";
         #elif defined( __OSX__ )
             m_DependencyGraphFile += ".osx.fdb";
-        #elif defined( __LINUX )
+        #elif defined( __LINUX__ )
             m_DependencyGraphFile += ".linux.fdb";
         #endif
     }


ffulin:
Thanks for the fix. This will be in the next release (v0.98).

While this issue is unfortunate, this type of mistake is also fairly rare so I don't think we want to change the way these includes are managed for the time being.

ffulin:
Related to this issue, if the database was incorrectly parsed, fbuild would exit silently instead of reporting an error. This could happen if fbuild attempted to read itself as the database, if this bug triggered and the database and exe were in the same dir.

I have fixed this issue (https://github.com/fastbuild/fastbuild/commit/103055eac0fc08976069b45f18e7855ee1d2b56d). This fix will also be in the next release (v0.99)

dummyunit:
> This could happen if fbuild attempted to read itself as the database, if this bug triggered and the database and exe were in the same dir.

Yes, this is exactly what happened on Travis. I was planning to investigate why nothing is printed in that case, but you already fixed it, so, great.

PR #554, opened 2019-05-07
[New] Add "Work when available" as continuous idle detection between …

ffulin:
…Idle and Always.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/options.html
+++ b/Code/Tools/FBuild/Documentation/docs/options.html
@@ -172,7 +172,7 @@
 
 <table width=700>
   <tr>
-    <th width=250 align=left>Option</th>
+    <th width=340 align=left>Option</th>
     <th align=left>Summary</th>
   </tr>
   <tr>
@@ -184,7 +184,7 @@
     <td>Control worker CPUs allocation.</td>
   </tr>
   <tr>
-    <td><a href="#mode">-mode=[disabled|idle|dedicated]</a></td>
+    <td><a href="#mode">-mode=[disabled|idle|dedicated|proportional]</a></td>
     <td>Control worker availability.</td>
   </tr>
   <tr>
@@ -469,7 +469,7 @@ runs in UI-less mode)</p>
 </div>
 
 
-    <div class='newsitemheader' id="mode">-mode=[disabled|idle|dedicated]</div>
+    <div class='newsitemheader' id="mode">-mode=[disabled|idle|dedicated|proportional]</div>
     <div class='newsitembody'>
 <p>Control worker availability.</p>
 <p>The FBuildWorker.exe mode is normally controlled through the UI. The "-mode" command line option will override this as follows:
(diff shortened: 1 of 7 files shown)


ffulin:
@Nesquick0 New pull request to integrate"continuous" idle behavior.

ffulin:
Sorry this has taken so long. Re-evaluating the current PR, I think we can get it integrated if we resolve the following minor issues:

*a) Change in behavior for existing users* The PR as it stands will switch all existing users from "Always" to the new mode because of the change in enum value. The enum ordering as you have it makes logical sense, but I think we need to add the new mode at the end the enum to avoid this. I can make this change after integration. (We could decouple the order of the enum from the drop-down, but for now at least I think this would be overkill)

*b) Naming* I think the naming "when available" is a little confusing. The "When Idle" is already described in very similar terms. I think something like "Proportional" would be better (described something like as "A proportional amount of CPUs will be allocated based on system usage"). I'm open to other names/wording. I can make this change after integration

*c) Clamp()* You mentioned in your other PR that you changed the 50 to 10 in the mIdleSmoother because it looked like a bug. This functionality is actually deliberate. It is designed to prevent bouncing rapidly between idle and busy states - i.e. once we reach idle, it takes a period of sustained CPU usage to flip back to "busy". I can restore this after integration. I think the de-bouncing is not needed for the new mode you added, so the clamp there is fine.

Let me know what you think about these points and I'll get the change integrated.

Nesquick0:
Thank you for your response.

**a)** I agree with you, that compatibility is probably better in this case. I've updated my branch.

**b)** "Proportional" sounds good. I don't have anything better. I changed enum, texts and cmd parameters accordingly.

**c) I disagree on this one.** I understand that it should prevent bouncing between states. But based on code I guess that it should change Idle state from available to disabled in 2.5s and from disabled to available in 5s (as said in `if ( m_IdleSmoother == 10 ) // 5 secs (called every ~500ms)`
But with Clamp set to 50, going from state disabled (m_IdleSmoother is 0), it will change to available in 5s (m_IdleSmoother is 10), but keep growing up to 50. But when PC become used and m_IdleSmoother is going down by -2, it will took **25x 500ms iterations to reach 0, which is 12.5s**. And this doesn't make sense to me with `m_IdleSmoother -= 2; // become non-idle more quickly than we become idle`.

ffulin:
The comment in the code ("become not-idle more quickly") and the lack off comments around the clamp are a bit confusing, and I think my previous comment omitted some information about the original intent of this code, which is to handle two situations:
* **a) we're not idle yet**, so any CPU usage should aggressively lower the idle timer (the -2)
* **b) we are idle,** so we want to increase the time taken to leave the idle state, relative to how long we've been idle (the Clamp to 50)

This is why there is the contradiction. In retrospect some more explicit if/else logic would have been better, and certainly some better comments.

With that in mind, the logic is working as originally intended, but thinking about it as part of this discussion makes me think that it's perhaps not really desirable....

The idea behind exiting idle more slowly was that if we're actively compiling, we want some extra time to complete that work. The idea being that FASTBuild would cancel jobs, and we wanted to allow some additional slack in that logic to not terminate too quickly. FASTBuild does not yet cancel jobs however, and conceivably the current logic allows for more jobs to be accepted when exiting idle than is actually desired. Additionally, cancelling (or not) existing jobs in the future could be best managed with explicit logic independent of this timing if needed.

I guess all of that is a long-winded way of saying that I think you've convinced me :)

Nesquick0:
Thank you for explanation. It make sense now. Change it during integration how you like it in current state.

ffulin:
I've integrated this new functionality. "Proportional" mode will be in the next release (v0.99). Thanks for helping to make FASTBuild better!

A small note: I incremented the current version for the FBuildWorker's settings file, so that older workers will know they cannot load the new settings file. (New worker can of course load the old settings just fine as we discussed)

PR #555, opened 2019-05-08
Fix bug where FBuild uses wrong database filename on Linux

BlueMonday:
FBuild.cpp was incorrectly checking if the __LINUX macro was defined. It should
have instead checked for the __LINUX__ macro. This resulted in a file called
"fbuild" being as the database file on Linux.

We happened to have an existing file with that name (the fbuild binary itself).
FBuild ended up reading our existing file and failng to parse it since it was
not the format it expected.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/FBuild.cpp
+++ b/Code/Tools/FBuild/FBuildCore/FBuild.cpp
@@ -155,7 +155,7 @@ bool FBuild::Initialize( const char * nodeGraphDBFile )
             m_DependencyGraphFile += ".windows.fdb";
         #elif defined( __OSX__ )
             m_DependencyGraphFile += ".osx.fdb";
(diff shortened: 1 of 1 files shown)


ffulin:
For completeness, this issue will be fixed in the next version (v0.99) with two fixes:

Use the correct fbuild.linux.fdb name:
 - https://github.com/fastbuild/fastbuild/commit/aa70e04dcd109868d0241e368986806723586557

Emit an error when the .fdb can't be parsed correctly, instead of failing silently:
 - https://github.com/fastbuild/fastbuild/commit/103055eac0fc08976069b45f18e7855ee1d2b56d

BlueMonday:
> Emit an error when the .fdb can't be parsed correctly, instead of failing silently (103055e)

Thank you for following up and adding extra logging!

PR #558, opened 2019-05-10
Added missing support for the CopyDir node in libraries.

Michron:
This should make it possible to have a dependency on another project which has a `CopyDir` in its targets.

For example:
ProjectA.bff
```
DLL( 'Project_A_DLL' )
{
    .Libraries = {
                    // This will now cause "Error #1005 - DLL() - Unsupported node type in 'Libraries'." (Node: 'Copy_Directory', Type: 'CopyDir')
                    'Project_B'
                 }
}
```

ProjectB.bff
```
DLL( 'Project_B_DLL' )
{
    [...]
}

CopyDir( 'Copy_Directory' )
{
    .SourcePaths = 'folder\source\'
    .Dest        = 'folder\destination\'
}

Alias( 'Project_B' )
{
    .Targets =  {
                    'Copy_Directory'
                    'Project_B_DLL'
                }
}
```

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
@@ -1359,6 +1359,14 @@ void LinkerNode::GetImportLibName( const AString & args, AString & importLibName
         return true;
     }
 
+    // a directory copy?
+    if (node->GetType() == Node::COPY_DIR_NODE)
+    {
+        // depend on copy - will use input at build time
+        nodes.Append(Dependency(node));
+        return true;
+    }
+
     // an external executable?
     if ( node->GetType() == Node::EXEC_NODE )
     {


ffulin:
Thanks for the PR. Unfortunately, I don't think this code will work.

This changes means LinkerNode will add a dependency on the CopyDir node, but this will then later be used in GetInputFiles and simply treated as one file, using the name of the CopyDir itself to add to the linker command line:
{code}
    else
    {
        // link anything else directly
        fullArgs += pre;
        fullArgs += n->GetName();
        fullArgs += post;
    }
{code}
instead of expanding all the copied files.

Perhaps in your case you were using the input just to get a side-effect copy of some files that don't actually get used for linking? If so, you can achieve that by using PreBuildDependencies or creating an Alias that bundles up the link and the copy. (If this is the case, let me know and I can elaborate with examples)

If you are actually linking in the copied files, I'd be interested to know more about your setup (as I can't see how that would work currently).

I'm also concerned that there might be some issues around the order of operations: directory listings are tricky to manage in FASTBuild and it's likely the CopyDir would at least need to be treated as a PreBuildDependency to work correctly in all cases. With this (and the other issue) in mind I think we'd need to create some tests around these cases to have confidence in this feature.

Michron:
Oh I see, I was assuming that `CopyDir` would be similar to `Copy` which is why I thought it was missing.

I was indeed using `CopyDir` to move some existing files to the output directory without the need to link them. Without using FastBuild the project is currently copying the dependencies in a custom build step using `xcopy`. I'll see if I can get it working by using `PreBuildDependencies` or creating an Alias that bundles up the link and the copy.
We're using Sharpmake to generate the project and FastBuild files, so I might also be able to solve it by having Sharpmake set up a junction link to the output directory, which saves us from copying some files as well.

PR #559, opened 2019-05-10
Fixes for shared memory test

dummyunit:
This is a series of fixes for TestSharedMemory test:
1. Improvement for failure handling in the test.
2. Fix for sporadic failures of the test on Travis caused by a race between main process and the fork.
3. Support of ASan/MSan and Clang configurations in the chain of ifdefs that produce unique names for global objects in test.

More info in the corresponding commit messages.

### DIFF
--- a/Code/Core/CoreTest/Tests/TestSharedMemory.cpp
+++ b/Code/Core/CoreTest/Tests/TestSharedMemory.cpp
@@ -18,6 +18,10 @@
     #include <unistd.h>
 #endif
 
+#if !defined( __has_feature )
+    #define __has_feature( ... ) 0
+#endif
+
 // TestSharedMemory
 //------------------------------------------------------------------------------
 class TestSharedMemory : public UnitTest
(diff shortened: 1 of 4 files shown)


ffulin:
Thanks for the fixes. I've integrated them and they will be part of the next source release (v0.99)

ffulin:
There was a small problem with this change on OS X. It seems the shared memory name was becoming too long so the test would fail to create the memory. I changed the test to use the parent pid() as a uniquifier so that this would no longer occur. This should have the added benefit of future proofing this code against future config build changes/additions. (https://github.com/fastbuild/fastbuild/commit/f46d2ad9a06516bc7c98f538b925d3cbe15e18be and https://github.com/fastbuild/fastbuild/commit/d0b2332aa761cb76ba3e638d792f37ff550b0470)

I think at least one other test has similar name uniquification logic, so I'll look at making that consistent.

dummyunit:
I see that you removed the try-catch block around the code executed in the forked process. Why?
 
As I described in the comments in the code the try-catch block prevented forked process from leaving `TestSharedMemory::CreateAccessDestroy` via exception, continuing to run subsequent tests, printing the summary at the end and causing much confusion as the result.
Although the are no more `TEST_ASSERT`'s in that block of code, exceptions can be thrown by regular `ASSERT`'s inside `SharedMemory` and `Timer` classes (`ASSERT` calls `AssertHandler::Failure` and it calls a callback set by `UnitTestManager`), so the problem is still relevant.

ffulin:
I removed the try/catch because the test now returns explicit return codes for the places it can fail, which I think makes debugging the calling process more convenient.

You are right to point out that some lower level code could still throw though due to a failed assert. I will change the test to explicitly disable the throw behavior in the forked child, by setting its own SetAssertCallback.

dummyunit:
> I removed the try/catch because the test now returns explicit return codes for the places it can fail, which I think makes debugging the calling process more convenient.

That is OK, just wanted to make sure that it wasn't deleted by accident.

ffulin:
Thanks! I submitted a further change which disables the "throw on assert" behavior for the child process and also improves the output a little to help diagnose any future problems we might have. (https://github.com/fastbuild/fastbuild/commit/a9c8bf47e26cf8329c7425a52ad46bb857ad65a8)

PR #560, opened 2019-05-10
Add missing newline to some messages

dummyunit:
Noticed that some displayed messages are missing a newline at the end.


ffulin:
Thanks for the fix. This has been integrated and will be in the next release (v0.99)

PR #561, opened 2019-05-10
Report error and exit when the intermediate process fails instead of hanging

dummyunit:
Intermediate wrapper process may fail to start the final wrapper process. In that case it exits with an appropriate error code but the main process ignores it and waits (indefinitely) for a signal from the final process.

To fix this problem main process now checks the return code of the intermediate process and proceeds to waiting only if it succeeded.

Main process can still hang if the final process crashes/exits before it can set the `Started` flag. But that problem is not so easy to fix.


ffulin:
Thanks for the fix. This change will be in the next version (v0.99)

Issue #562, opened 2019-05-10
Light cache and #import interaction


dummyunit:
Currently light cache ignores MSVC's `#import` directives and I think this is wrong.

`#import` is equivalent to `#include` of the files generated from the imported type library and they may change when the type library changes and no modifications are done to the source file with the `#import` directive. I think in that case FASTBuild will currently use a cached object file instead of rebuilding the source.

I think the most reasonable solution is to disable light cache for the source file if we encounter an `#import` directive during parsing.

Also it seems that supporting `#import` directives properly would be impossible because:
1. From reading the documentation it looks like one type library may reference another one (and I suppose this affects the generated files).
2. Some syntax form doesn't even mention a file name.
3. I am not sure that content of the generated files is fully determined by the imported library and options in the `#import` directive.

ffulin:
The fix you supplied has been integrated and will be in the next version (v0.99). I'll leave this issue open until that versions is released.

ffulin:
FASTBuild v0.99 has now been released which includes a fix for this issue.

PR #563, opened 2019-05-10
Detect #import directives in light cache

dummyunit:
I went ahead and added a check for `#import` directives to `LightCache`, as proposed in #562.

Note: LightCache.cpp have mixed indentation (tabs+spaces) that is why indentation looks wrong on GitHub.


ffulin:

Note that the LightCache is still very rough, and will likely see a lot of code churn, so you might want to avoid doing too much in there. (Though it's very good you made this change because it is something I had not yet considered!)

PR #564, opened 2019-05-11
Use correct extension for temporary files with preprocessed output

dummyunit:
GCC decides how to interpret input files by looking at file's extensions, in particular it determines the language of the input and the fact whether input was already preprocessed or not. This applies to `gcc` binary itself (not to `g++`, `g++` "upgrades" `.c` files to C++).

The problem was that we always used `.i` extension for the temporary files in which output of the preprocessor was stored, and that extension tell GCC that the file is a preprocessed C program.
So if a C++ source was compiled with separate preprocessing step and `.Compiler` referenced to a `gcc` binary then GCC interpreted that source as C code and failed to compile it.

To fix that `ObjectNode::WriteTmpFile` now selects proper extension for the temporary file based on the extension of the original file.

Change to `.Executable` value in the GCC configuration file allows us to check that we can now use `gcc` binary as the compiler. Before this change it wasn't possible – GCC reported that `-Wno-invalid-offsetof` is not a valid option for compiling C code.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -2148,7 +2148,63 @@ bool ObjectNode::WriteTmpFile( Job * job, AString & tmpDirectory, AString & tmpF
     AStackString<> fileName( sourceFile->GetName().FindLast( NATIVE_SLASH ) + 1 );
     if ( GetFlag( FLAG_GCC ) )
     {
-        fileName += ".i";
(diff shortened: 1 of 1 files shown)


dummyunit:
Well... It looks like CI is already paying off :)  
I have applied the change from `g++` to `gcc` binary to `GCC_CI.bff` and quickly found out that this change can not be done right now because it breaks the build with the latest release version. Apparently I was still using v0.97 locally and didn't catch that.

So I have reverted the change to .bff files for now, it could be reapplied after the next release.

ffulin:
Thanks for the fix! This has been integrated and will be in the next release (v0.99)

PR #565, opened 2019-05-11
Update value of the "-x" option when compiling preprocessed source

dummyunit:
It is possible to explicitly specify the type of an input file to GCC and Clang via `-x` option.
Currently, when we are compiling a preprocessed source we don't touch `-x` options and this forces GCC and Clang to preprocess the source a second time because they are told so by the option.

To fix this problem we now change the value of the `-x` option to a corresponding "cpp-output" variant.

This has 2 benefits:
1. Compilation of the preprocessed source is slightly faster.
2. During distributed compilation GCC running on the remote machine doesn't try to inject contents of `/usr/include/stdc-predef.h` from that machine into preprocessed source.
This avoids potential problems that may arise when that file is different on the remote machine.

Quoting of some `-I` had to be changed in configuration files because currently code in `ObjectNode::BuildArgs` doesn't recognize options that are quoted entirely. This leads to these `-I` options not being removed when compiling preprocessed code in Clang (now Clang actually knows that the input is preprocessed) and Clang producing a warning which breaks the build:
```
clang-8: fatal error: argument unused during compilation: '-I ../External/LZ4/lz4-1.9.1/lib' [-Wunused-command-line-argument]
```

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1670,6 +1670,34 @@ bool ObjectNode::BuildArgs( const Job * job, Args & fullArgs, Pass pass, bool us
                 {
                     continue; // skip this token in both cases
                 }
+
+                // To avoid preprocesing code a second time we need to update
+                // arguments of -x option to use the "cpp-output" variant.
+                // We must do this inplace because the argument order matters in this case.
+                if ( token == "-x" && i + 1 < numTokens )
+                {
+                    // Save the "-x" token
+                    fullArgs += token;
+                    fullArgs.AddDelimiter();
+
+                    // Change the argument to its "cpp-output" variant.
+                    const AString & language = tokens[ ++i ];
+                    if ( language == "c" )
+                    {
+                        fullArgs += "cpp-output";
+                    }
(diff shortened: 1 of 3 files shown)


ffulin:
I finally managed to integrate this fix (https://github.com/fastbuild/fastbuild/commit/cc619704dd2e8ee8bbe52962193a0dbb0ae75abf).

I didn't directly intergrate your PR, because I didn't want to apply the include path changes. I instead have fixed the command line processing to correctly handle the quotes in that case. Apart from that, the core fix is just as you implemented it.

This change will be in the next release (v1.05)

Thanks!

ffulin:
I found that fixing up the language arg as this change does is incompatible with Clang prior to v10.

In Clang 9 or earlier, the -D options on the command line are ignored when compiling with "-x c++-cpp-output", which breaks compilation if "-frewrite-includes" is used (which FASTBuild uses by default)

To retain backwards compatibility, I have added a .ClangGCCUpdateXLanguageArg toggle to Compiler() to enabled this change, and I made it off by default (https://github.com/fastbuild/fastbuild/commit/33d894dfc5621758bfe9dcadfdceeb338f3ed2c3). This will avoid a breaking change for users that would force them to upgrade the version of Clang they were using, while still allowing this feature to ship in v1.05

PR #567, opened 2019-05-17
Added .natvis file for array and string auto-expansion

belkiss:


ffulin:
A visualizer for these types already exists (Code/Core/Core.natvis).  It's also already added to the project as .natvis files are included by default.

PR #569, opened 2019-05-17
Reworked -debug option so that it displays a message box

belkiss:
- -debug option in fbuild is now usable in release


ffulin:
Thanks for the enhancement. This has been integrated and will be in the next release (v1.00)

PR #570, opened 2019-05-17
initialize startup helper early

belkiss:
This is done to ensure that it doesn't get unitialized a little bit too soon. DNS Lookup threads sometimes try to access the network after network get uninitialized during shutdown and this avoids it.


ffulin:
The NameResolutionThreadFunc was changed some time ago to have it's own NetworkStartupHelper instance to fix the issue you describe. Is it possibly this is an old change you we don't need? Or is there still some edge case that isn't handled?

ffulin:
This change should not be required (as NetworkStartupHelper has its own helper scoped around lookup). We can re-open this issue if I missed something.

belkiss:
Alright, will revert it in our internal version as well and update this PR if the error comes back.
Thanks!

PR #571, opened 2019-05-17
Change Idle Threshold hardcoded define to a setting in the worker UI dropdown

belkiss:
This is what it looks like:

image


belkiss:
We have another commit that makes the idle threshold a command line parameter, that I'll send when that one gets it.

ffulin:
I've integrated this change and it will be in the next release (v1.00). Thanks!

NOTE: I made some changes:
 - fixed bug where "Idle Threshold" drop down was not disabled if starting outside of "When Idle" mode
 - incremented worker settings version to v4 (was already v3 due to other changes)
 - implement missing UI features on OSX

belkiss:
Excellent, thanks!

PR #572, opened 2019-05-17
Allow FastBuildWorker to clean the broker folder of stale files

belkiss:
Activate with -allowbrokerclean.


Code/Core/Time/Time.cpp
```diff
@@ -38,5 +38,13 @@ uint64_t Time::GetCurrentFileTime()
...
     #endif
 }
-
+//------------------------------------------------------------------------------
+uint64_t Time::FileTimeToSeconds(uint64_t _filetime)
+{
+    #if defined( __WINDOWS__ )
+        return _filetime / 10000000U;
+    #else
+        #error Unknown platform
```
ffulin: This will break compilation on OS X and Linux.

ffulin: Cache.cpp contains the necessary math to convert:
```C++
        #if defined( __WINDOWS__ )
            const uint64_t oneDay = ( 24 * 60 * 60 * (uint64_t)10000000 );
        #else
            const uint64_t oneDay = ( 24 * 60 * 60 * (uint64_t)1000000000 );
        #endif
```

ffulin:
Overall this change looks pretty good, but it won't compile for Linux or OS X. It does look like very little needs to be fixed to make it work though.

Code/Core/Time/Time.cpp
belkiss: Fixed, sorry about that!

ffulin:
There is another PR pending that will result brokerage files being periodically updated by all workers (https://github.com/fastbuild/fastbuild/pull/636). Once that's in place, I think this change will be superseded if we allow the worker to ignore old brokerage files. To me that seems like an improved solution as it removes the need for a special worker to be running in a special mode.

Thoughts?

jspelletier:
Regarding #636, if there is no cleanup of the files as they are ignored, how would someone glancing at the worker folder quickly know about the real number of available workers ?

How about stale workers such as cloud machine that got killed? Having a lot of stale workers is  bad as it slows down a lot the search of workers.

Before we had that code we often had to delete manually the content of the folder because we had a ton of killed cloud machines and their files were still there... We no longer have this problem. 

BTW we run all workers with the switch, so there is frequent cleanups done in the folder as each of the workers does a cleanup every 12h because we got hundreds of workers in our worker pools,

bburgin:
In pull request #636, there is no cleanup of the files. The purpose of that PR is to have each worker set the modification timestamp for its brokerage file, allowing you to write a separate script that periodically cleans up the brokerage files that are older than a specific time.
This solution also allows you to define the interval you want for the cleanup operation.
And it means only one machine does the cleanup operation.

I think pull request #572 here is not optimal, since with it all workers would try to enumerate and delete the files. So, there would be n times more network usage for it, where n is the number of live workers.
Also with #572, workers would be generating more throw away work, when they enumerate files but in the mean time some other worker deletes the files, before they do. Their delete operations would be noop wasted effort.

ffulin:
My original intent was to ignore old files and not worry about cleanup. The brokerage was never intended to be "human readable" and i imagined builds could just ignore old workers one the tokens were being periodically updated by valid workers (#636).

I hadn't really considered the case where the number of stale tokens could get out of control due to the use of short-lived VMs or use of "cloud" machines.

So with this new information in hand, my plan is to first:
 - integrate #636 to have tokens updated periodically. Whatever else we do this allows us to detect old/stale workers.

For the next step I can see two possible ways forward. Either:

a) integrate this PR and have workers delete tokens periodically. I think however I would make the feature "always on" and not require an argument. The scans doesn't need to be that frequent so the traffic will be relatively low even with many workers and this will avoid more configuration/args for users to manage (things will just work)

OR

b) Ignore this PR and change the fbuild.exe so that when it starts a build and obtains the list of workers, it deletes any old ones. The advantage of this is that it will add zero overhead when there are no stale workers (we get the list which we'd do anyway and then see there is nothing to clean)

I'll brew on this a little more before moving ahead, but I plan to sort this out very soon.

bburgin:
Some concerns:

1. Client machines running fbuild.exe may not have write or delete access to the brokerage path. (They may have only read access, if they are using a pool of worker machines that they are not a part of.) Because of this, I think any auto-deletion logic should be in the fbuildworker.exe rather than the fbuild.exe.

2. How often would workers enumerate the brokerage paths and delete old ones? Would this be configurable? Would there be a way to disable it?
I ask because auto-cleanup in the workers would result in an m x n scaling problem, where m is the number of live machines enumerating files and n is the number of files. Does it scale? What if a worker pool has 2000 live machines and 2100 files? It seems like this would slow down workers and reduce the number of compiles / tests they could run, on average. I would much rather disable auto-cleanup in workers and use a dedicated script that does its work without slowing down workers.

jspelletier:
One of the idea we had, is we didn't want to depend on external scripts to do the cleanup, thus we've put the cleanup in the worker as it needs write access as Brian said. And since this was a new feature we've added a switch to control which machines has this feature active but we've put the switch in the default switches for executing fbuildworker on my project. 

This has been running well for close to a year in a pool with at least 300 workers on my project(not sure of the peak numbers but as of this morning there is > 300 available workers). There are also other big pools elsewhere in Ubisoft using that feature.

We've got to remember that since the code deletes files older than 24h, this doesn't occurs that much often in practice. In general most workers won't delete anything when executing the cleanup and if there is some files to delete, it will generally be 1-2 files to delete... it is not that much.

Each machine does 2 cleanups per 24h period so for a 300 machines pools, this is only 600 folders enumerations. This is really nothing compared to the number of enumerations resulting of new builds. On my project we've got up to a thousand changelists to compile everyday by close to 100 build machines and many other compilation done by programmers executing compilation on their pc or preflight builds on build machines... This results in thousands of directory enumerations at the start of each fbuild.exe invocation so I think the cleanup enumerations are negligible.

Thanks!

ffulin:

This feature where workers update the timestamp of their brokerage file was also integrated prior to this (https://github.com/fastbuild/fastbuild/commit/daa51080a1861312eb14a918d3b44fb364f55280).

I removed the command line option so this feature is always on, because (as noted) enumerating of the brokerage is very infrequent and I think this simplicity of not having another thing to tweak is worth the minor network traffic cost.

jspelletier:
thanks. this is perfect for me.

PR #573, opened 2019-05-17
Added special case for out of RAM occuring while executing cl.exe

belkiss:


ffulin:
This case (C1060) is already handled by FASTBuild, but this implementation considering C1076 or C3859 is better. I've merged the two checks and added some additional comments (https://github.com/fastbuild/fastbuild/commit/7b8e75475c751c1b80eb403fdc9338a275b2515b)

This new behavior will be in the next release (v0.99). Thanks!

PR #574, opened 2019-05-17
Added support for free OS memory detection

belkiss:
If free memory is smaller than a threshold, disable the worker.

- By default the threshold is 1GB.
- Threshold is configurable by command line option(-minfreememory)


ffulin:
This change has been integrated and will be in the next release (v1.00) Thanks!

belkiss:
Thanks!

PR #575, opened 2019-05-17
Add optional command line for the worker: -writeextrainfo

belkiss:
This command line will add extra info in the file created in the brokerage:
- Version number
- Username (using the USERNAME or USER environment variable)
- Number of CPUs shared/total
- Sharing mode (dedicated/idle)

This is mostly for debug purposes, and is intended to be read by humans, not parsed

Ex:
```
v0.96_UBI_6
username
6/12 cpus
idle (40%)
```


ffulin:
I've actually been thinking about such a change for a while now. I may make this always the behavior (removing the option), since I think the additional expense is small and the extra info can be quite useful.

ffulin:
I've integrated this change. It'll be in the next release (v1.00). Thanks!

I made a few changes (https://github.com/fastbuild/fastbuild/commit/180a5f7bc34a0e918408fce93e99dbada9d490cd):
 - the command line option is removed and this feature is always on
 - the contents of the file changed slightly (formatting)

belkiss:
Thanks!

belkiss:
After a quick look at your version, seems you removed the idle threshold value from the file. Any reason?

ffulin:
> seems you removed the idle threshold value from the file. Any reason?

This was not intentional, but the result of merging changes (this one plus the new "proportional" mode) as well as some refactoring.

I've restored the idle threshold output: https://github.com/fastbuild/fastbuild/commit/58bb73dceeb962089133f0c099128bca32bc3166

(That'll be in v1.00)

PR #385, opened 2018-02-14
Fixes not finding libraries in LinkerOptions

fran6co:
Libraries defined as part of `LinkerOptions` are ignored, breaking dependencies. This fixes a couple of bugs:

- The full path check worked only on Windows
- If `libPaths` is empty and the libraries are defined in relative form

PR #576, opened 2019-05-17
Linker detection by checking the full linker EXE name, instead of partial match

belkiss:


ffulin:
Looks like a good tidy up. Apart from seemingly fixing the "mwldeppc." case (bad trailing .) does this impact other cases?  i.e. did you see some compilers not correctly detected before this change?

belkiss:
That change was done originally to be able to support `lld-link`, without having to rely on the order of the ifs.

fran6co:
I reimplemented this in my pull request #385, but this one looks a lot better. I would recommend merging this one first

ffulin:
Prior to this change, "lld-link.exe" would be detected as LINK_FLAG_MSVC (i.e. msvc compatible) but now it won't be detected at all.

I note that you mentioned this change was to support lld-link, but I can't see how it does that. Would you mind clarifying please?

belkiss:
This change has been split out from the commit we made internally, which is one big chunk that adds support for clang-cl and lld-link, which I have sent as a separate pull request: https://github.com/fastbuild/fastbuild/pull/591

ffulin:
I did look at that change, and it seems these changes are incompatible.

This PR explicitly stops "lld-link" being detected as an MSVC compatible linker, while #591 adds a comment to the very line this change removes, noting that it's needed to "recognize lld-link".

Maybe I'm missing something here?

PR #577, opened 2019-05-17
Improve error reporting when TCPDEBUG is active.

belkiss:


ffulin:
Thanks! This improvement will be in the next source release (v0.99)

belkiss:
Thanks!

Issue #411, opened 2018-04-13
Killing cl.exe yields unhandled error code 1


aganea:
We've discovered something interesting on our build system: occasionaly builds were failing, and we didn't understand why. It was always MSVC compilation (cl.exe) which was failing remotely, on a machine belonging to a real user (not a build machine, we have both mixed which participate to the build).

We ended investigating each case with the user who was logged on the filing PC. It turned out that FBWorker.exe was slowing down their computer and that they manually killed cl.exe.

Our build system indeed reports the agent's result, with an error code:

` Failed to build Object (error 0x1) 'C:\fastbuild-work\.fbuild.tmp\0x00000000\core_1005\myproject_unity15.obj'`

Of interest here is the error code (0x1). It turns out, from the stack overflow post below, that error code 1 means that the process was terminated by the Task Manager:

https://stackoverflow.com/questions/4344923/process-exit-code-when-process-is-killed-forcibly

I was thusly wondering if adding return code 0x1 to ObjectNode.cpp/HandleSystemFailures as a system error, would be a good idea in your opinion? (juste where it says "// If remote PC is shutdown by user, compiler can be terminated"). You already seem to do this for clang and gcc.

ffulin:
Thanks for reporting this. I was able to reproduce the problem.

I've made return code 0x1 now be considered a system failure (so the item will be retried). Some additional logic was needed, because Clang and GCC return 0x1 when terminating normally if there is a compile error (and we want to continue report that error rather than silently retrying).

I've made the necessary changes to fix this problem (https://github.com/fastbuild/fastbuild/commit/7f4189260f676787679f8f9fbb1258fb04b2e6de) and manually terminating tasks on the remote worker (both clang and cl) no longer causes the build to fail (and clang compile errors generated remotely are still correctly reported).

This fix will be in the next release (v0.96). I'll leave this issue open until that version is available.

aganea:
I did the following change, similar to yours. Please note the additional changes to Client.cpp to prevent killing the <LOCAL RACE>, if any. In that case the local compilation should finish its job.

```
From e7cce35655f8a9d3772b476a190fc88bcc85b91d Mon Sep 17 00:00:00 2001
From: Alexandre Ganea <alexandre.ganea@ubisoft.com>
Date: Fri, 13 Apr 2018 17:01:39 -0400
Subject: [PATCH] Retry fix for remote builds that fail when the remote user
 kills the compilation

[...]

PR #579, opened 2019-05-17
Fix remote build failures, when the user kills cl.exe

belkiss:
- Retry fix for remote builds that fail when the remote user kills the compilation

When a compilation process (cl.exe or orbis-clang++.exe) is spawned by the
FBuildWorker, sometimes that can make the machine unresponsible, and as a result
it can happen that the users are killing the compilation process manually.

When that happens the process terminates with code 1, as documented here:
https://stackoverflow.com/questions/4344923/process-exit-code-when-process-is-killed-forcibly
In that case the remote worker (FBuildWorker) will tag that job as a
"system failure", which will force the client to re-spawn the job on a different
worker. The client will also "blacklist" that worker as a result, which prevents
from spawning subsequent jobs on that client (as long as the client runs).

The change is a bit tricker when clang is used, because clang uses 1 as a exit
code when the compilation fails. We consequently check if the log contains
"error: ", and if not we consider the exit code as a call to TerminateProcess()
initiated by the remote user.

First issue is described here: https://github.com/fastbuild/fastbuild/issues/411

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -2389,6 +2389,15 @@ bool ObjectNode::CompileHelper::SpawnCompiler( Job * job,
         // MSVC errors
         if ( objectNode->IsMSVC() )
         {
+            // The compiler executable was manually killed by the user, through the TaskManager
+            // In that case the error code is set to 1 by TerminateProcess()
+            // See https://stackoverflow.com/questions/4344923/process-exit-code-when-process-is-killed-forcibly
+            if ( result == 0x01 )
+            {
+                job->OnSystemError();
+                return;
+            }
+
             // But we need to determine if it's actually an out of space
             // (rather than some compile error with missing file(s))
             // These error codes have been observed in the wild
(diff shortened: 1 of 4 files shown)


Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -2389,6 +2389,15 @@ bool ObjectNode::CompileHelper::SpawnCompiler( Job * job,
         // MSVC errors
         if ( objectNode->IsMSVC() )
         {
```
ffulin: The two added conditions seem to duplicate the code above:
```C++
        // If process is manually terminated by a user, consider that a system failure
        // Only do so if there is no error output, as some compilers (like Clang) also
        // use return code 1 for normal compilation failure
        if ( ( result == 0x1 ) && ( stdErr == nullptr ) )
        {
            job->OnSystemError(); // task will be retried on another worker
            return;
        }
}
```
albeit with slightly different conditions. A the time I added the above fix (added ~0.96) I was able to reproduce the behavior and confirm killing the process was handled gracefully. Is it possible we don't need this change, or do you know if the differences in implementation handle some additional case?

ffulin:
This change seems to contain two things:
a) detection of termination
b) changes to handling remote failures when racing

The PR description doesn't mention the second thing. Do you have information about the behavior that was seen/fixed?

ffulin:
FASTBuild was already modified to detect this termination case (thing a), and the other thing (thing b) seems unrelated to this change. I'm going to close this PR for now. We can re-open or create a new one if needed.

belkiss:
Alright will have a look again and update if needed. Thanks!

PR #580, opened 2019-05-17
Prettify the DirectoryListNodes display in the stats summary, to avoid the very long unreadable lines

belkiss:
- Add a GetPrettyName virtual to Node class, that can be overriden in derived classes if needed
- If not overriden, the pretty will be the name
- Override GetPretttyName in DirectoryListNode


ffulin:
Thanks for the improvement. This has been integrated and will be in the next release (v0.99)

belkiss:
Thanks!

PR #583, opened 2019-05-17
Add some logs about the cache

belkiss:
- Path to the cache plugin, if indicated
- Init failure of the plugin
- Cache mode (Read, Write, Info, Trim)
- Cache path


ffulin:
This seems to add several additional lines to the output when using the cache, some of which seem redundant (Cache::Init already emits a warning when the cache can't be mounted for example). For the cache plugin case, I would expect the plugin itself to emit an error message when it fails (failure to load the dll is already reported by the wrapper).

I can see value in emitting a single line describing cache settings (as we do for distributed compilation), but i'm worried this is a bit spammy as it currently stands.

jspelletier:
This only adds two lines per build if using dll and one if no dll is used. I think this is not so bad. These are only displayed once per build.

PR #584, opened 2019-05-17
"Database has moved" is not an error anymore - it will trigger a recompile

belkiss:
Fix unit-test about database move change, now initialize will succeed

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
@@ -1652,7 +1652,8 @@ bool NodeGraph::ReadHeaderAndUsedFiles( IOStream & nodeGraphStream, const char* 
     if ( PathUtils::ArePathsEqual( originalNodeGraphDBFile, nodeGraphDBFileClean ) == false )
     {
         FLOG_WARN( "Database has been moved (originally at '%s', now at '%s').", originalNodeGraphDBFile.Get(), nodeGraphDBFileClean.Get() );
-        return false;
+        compatibleDB = false;
+        return true;
     }
 
     uint32_t numFiles = 0;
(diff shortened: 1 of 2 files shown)


ffulin:
Some FASTBuild users map a common root (via subst) to have consistent caching between machines even if the code lives in different paths. For those users, loading a Visual Studio solution accidentally via the original path instead of the path is a deliberate failure to avoid costly unnecessary recompiles. This change would be harmful to those users.

If you have use case where you deliberately alternate paths, or other wise move the source code, you can excplicitly delete the .fdb to have it rebuild, so this change in behavior is not needed there.

Is there another scenario or problem I haven't considered, that this change in behavior is helps with?

belkiss:
We typically are in such a case, we (sometimes) use the subst, and having this error about the database moved required us to have some heavy machinery based on fdb versioning etc. to allow activating/deactivating it in some occasions.
For instance, if we want to activate fastlink (which doesn't work with the subst), or disable the cache and the subst for testing, so we found that it was easier if this error could be demoted as a warning.

If you prefer, we could add a cmd line argument for this specifically to not treat db move as an error?

ffulin:
I went ahead and integrated this change and put it behind a command line option (-continueafterdbmove in https://github.com/fastbuild/fastbuild/commit/c5b40fa6adea33d82f0f4e0a2646a7b50c282fb6).

This will be in the next release (v1.00)

Thanks!

belkiss:
Excellent, thanks!

belkiss:
In fact, this case is a little bit like the one described in https://github.com/fastbuild/fastbuild/pull/586, where we don't want to stop the user from compiling. Imho the command line option is not really necessary, but it's not a blocker.

ffulin:
Unlike the corruption case, which could happen through no fault of the user (due to a power outage or somesuch) the DB moving would (as far as cases I can think of) always be caused by a specific user action, and that action would always be something I think they'd want to know about. Having a silent rebuild in that case seems problematic. I personally wouldn't ever want this to just rebuild in a way that I might not notice, even in the case of a build machine (since that would indicate some misconfiguration I'd want to fix)

I added the command line option as a compromise/future proofing since it was relatively low maintenance overhead and allows users to opt-in to something even if I might not chose that myself.

PR #585, opened 2019-05-17
Added support for multiple broker folders separated my semi-colon.

belkiss:
Fastbuild worker only registers in the first specified folder but we are allowed to find workers in all specified folders in order of priority


jspelletier:
Note: 
We use this in the following case.
Suppose you have studio A and B, with both having a pretty fast connection between both studios and A having a lot more machines than B. 

Machines in studio A would have their broker setup like this:
FASTBUILD_BROKERAGE_PATH=\\server\fastbuild\studios\A
While machines in studio B:
FASTBUILD_BROKERAGE_PATH=\\server\fastbuild\studios\B;\\server\fastbuild\studios\A

In this case, machines in both studio get registered  in their respective studio folder, but machines in studio B could also build using machines from studio A.

Since studio A have a lot of machine there is no point in using machines from studio B. That will also decrease the bandwidth used between the two studios and improve a bit latency for users in studio A as they won't try to build in studio B. For studio B user, well, it is better to add a little bit of latency than trying to build with only a few workers.

This feature was the result of a quick development that is really useful for special cases such as the one stated above.

Currently, however, there is no priority system for assignment of workers from your studio as all workers are merged in one big list(only enumeration of worker files was modified)

ffulin:
Thanks for the PR. This has been integrated and will be in the next release (v1.00)

jspelletier:
thanks

PR #586, opened 2019-05-17
Create a new FDB when there was an issue parsing the old one

belkiss:


ffulin:
Are you able to provide some more context about this change? The current behavior is intentional and I think it's strange to silently cope with corrupt databases (something that should never occur and would hide clean builds).

jspelletier:
I think the idea is if the db is corrupted on your pc or some build machines what are you going to do anyways in general ? You probably are going to delete the file and this will force a rebuild.... The main interest of this MR is that by doing that automatically it avoids build failures on build machines. In general programmers are pissed when the build system is spitting avoidable errors with no relations to their code.

bburgin:
I like the idea of auto-remediation, but I am unsure what auto-remediation logic is appropriate to include in FASTBuild and what would be better done in the caller.
On one side, Microsoft Office tries to recover from corrupt document files on open and notifies the user.
On the other side, FASTBuild is a build automation tool and so remediation there would require demotion of issues to warnings, in order to maintain the unattended operation of the tool.
Warnings are often ignored, so remediation would often hide issues.

I guess FASTBuild could do remediation and also provide a promote warnings to errors option, like compilers have. Would users use such an option? And would there be some warnings that users don't want to treat as errors, once promoted?

ffulin:
This change has been integrated and will be in the next release (v1.00). Some follow up changes/improvements were made in https://github.com/fastbuild/fastbuild/commit/8412eeb91a976fc10f0a0202b5f281b31376362f (including fixing a regression in the "moved DB" case).

A case like this is an interesting one. On the one hand, something terrible has happened, and you probably want to know about that (in case you are getting clean builds every time due to hardware problems for example). On the other, as noted, if this happens during automated build scenario and is due to something like power outage, the problem can be incorrectly attributed to the "next" build, which is annoying.

In the end, I chose to:
 - make a corrupt DB be ignored as per this PR
 - modify the existing message, which is now effectively demoted to warning, to more explicitly mention a corrupt DB
 - the corrupt DB is backed up before being replaced, so if desired can be used for investigation

jspelletier:
tthis is perfect for me. Thanks a lot!

PR #588, opened 2019-05-17
Add option "-nolocalrace"

belkiss:
With this, a job is either done locally or remotely and FB waits for it to finish. There is no more kill involved.

This was done to solve the issue we experienced related to kill process tree.


ffulin:
This option seems like it would have been useful for debugging originally, but it seems like there's no good reason to keep it now, right?

jspelletier:
we often still use this option internally at Ubisoft for debugging some issues. This is really a debugging option and wouldn't mind that this is put in a special experimental section in -help(same thing applies to --debugpreprocessor)

ffulin:
This change has been integrated and will be in the next release (v1.00)

jspelletier:
thanks

PR #590, opened 2019-05-17
Remove /DLL from compiler options, as this will add the define LL...

belkiss:


ffulin:
Thanks! This change will be in the next source code release (v0.99)

belkiss:
Thx :)

PR #591, opened 2019-05-17
Add proper support for clang-cl and lld-link

belkiss:
- Considering lld-link as it was link.exe (LINK_FLAG_MSVC) and clang-cl as its own type of compiler
- Fixed /IMPLIB support for LLD-link.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.cpp
@@ -68,7 +68,7 @@ bool FunctionObjectList::CheckCompilerOptions( const BFFIterator & iter, const A
         }
         else
         {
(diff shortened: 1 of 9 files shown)


ffulin:
I've integrated this change. It will be in the next release (v1.03). Thanks!

I had some problems with conflicts, so wasn't able to merge this PR directly. Instead I re-applied that changes manually. The commit is here: https://github.com/fastbuild/fastbuild/commit/44643a6b57d9aa4a99a8d4d79929d044836d0063

belkiss:
Oh wow nice, thanks!

I actually rebased that recently on our internal fork, I'll have a look if we have additional things and let you know if we do :)

ffulin:
There was a small error in the merge, which is fixed here: https://github.com/fastbuild/fastbuild/commit/06cdc86615e6583e60bed70bd0909bbb2d20cf3a

PR #594, opened 2019-05-17
Set the vs version for the generated fastbuild sln to the matching one

belkiss:
That way, opening the sln with vs selector (default) will open it in the correct ide version


ffulin:
This has been intergrated and will be part of the next source release (v0.99)

belkiss:
Thanks!

PR #596, opened 2019-05-24
Fixed merge conflict issue in FBuildOptions.cpp

poppolopoppo:
Just a small fix for '-compdb' switch which isn't recognized on dev branch :)


ffulin:
Nice catch - thanks!

PR #598, opened 2019-05-25
Some improvements for Travis configuration

dummyunit:
1. Added option `-nostoponerror` to fbuild invocations to ensure that all tests are run even if one of them fails early.
2. Added workaround for TravisCI timeout based on lack of output from running commands.
    Function `travis_wait` (provided by TravisCI) is not used because its behavior is inferior: it stores all output of the wrapped command in a file and displays it after command finishes.
    Such approach has some problems:
    1. Redirection to the file removes color from the output.
    2. If the wrapped command hangs then its output will not be displayed at all.
3. Option `-summary` is deduplicated and moved into `FBUILD_CMD` variable.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -136,7 +136,7 @@ install:
   - chmod +x ./${FBUILD}
   # Set up some variables for convenience.
   - |
-  - FBUILD_CMD="../${FBUILD} -noprogress -summary"
+  - FBUILD_CMD="../${FBUILD} -noprogress -nostoponerror -summary"
   - GCC=${GCC:-gcc}
   - GXX=${GCC/gcc/g++}
   - CLANG=${CLANG:-clang}


.travis.yml
```diff
@@ -143,7 +154,7 @@ script:
   - |
     case "${TRAVIS_OS_NAME}:${CFG}" in
       linux:|osx:|windows:)
-        ${FBUILD_CMD} Tests
+        no_timeout ${FBUILD_CMD} Tests
```
ffulin: I don't know too much about Travis, but I saw it has a built-in "travis_wait" command: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

Can that be used instead?

.travis.yml
```diff
@@ -109,6 +109,17 @@ install:
       ${SUCCESS}
     ;;
     esac
+  # Helper function to work around Travis timeouts based on lack of output
```
ffulin: Maybe a comment (as per your commit description) explaining why travis_wait is not used. Quite unfortunate you have to do something like this.

dummyunit:
(Thread about `travis_wait` somehow vanished so replying here.)

We can use `travis_wait`.  
I just think that it provides bad user experience:
1. No color in the output because commands detect that output goes to a file.
2. All output is displayed at the end, so can't watch the build process in real time.
3. If something goes wrong (timeout fires anyway due to bugs, etc) we get no output.

.travis.yml
dummyunit: I have updated the comment to list reasons why this custom wrapper is used instead of `travis_wait`.

And by the way, if you prefer to use `travis_wait` instead that is fine. I don't have a strong preference either way, just implemented what seemed like a better solution to me.

ffulin:
I removed my comment about travis_wait after I properly read your explanation about not using it.

Related to this, I just submitted what I believe is a fix for the very slow build and test times on Windows: https://github.com/fastbuild/fastbuild/commit/a5f90a7911222069c29dd1bee1e4a9e056b502a8. It might be that with this fix, we don't need this timeout prevention.

dummyunit:
> It might be that with this fix, we don't need this timeout prevention.

That would be nice. In that case we should instead reduce `.TestTimeOut` for `CI_BUILD` to something lower than 600 to timeout "on our terms" in the future.

ffulin:
> In that case we should instead reduce .TestTimeOut for CI_BUILD

Agreed. Ideally we unify it with the default and remove the special case.

I see a few other opportunities to speed up the tests in all cases, so I'll spend a bit more time on that before cleaning up the CI config.

dummyunit:
I have removed the `no_timeout` function since the `.TestTimeout` is now set to a value less than 600 and that should prevent Travis CI from terminating the job.

ffulin:
Thanks!

PR #599, opened 2019-05-25
Ensure that name of the test is written to stdout before running the test

dummyunit:
When the test binary crashes not all output sent to stdout may be actually written to the console/pipe/file, including the name of the test that was last started.

This change will allow us to know exactly which test caused the crash when it happens.

### DIFF
--- a/Code/TestFramework/UnitTestManager.cpp
+++ b/Code/TestFramework/UnitTestManager.cpp
@@ -13,6 +13,7 @@
 #include "Core/Strings/AString.h"
 #include "Core/Tracing/Tracing.h"
 
+#include <stdio.h>
 #include <string.h>
 #if defined( __WINDOWS__ )
     #include "Core/Env/WindowsHeader.h"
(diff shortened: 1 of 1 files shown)


Code/TestFramework/UnitTestManager.cpp
```diff
@@ -196,6 +197,9 @@ void UnitTestManager::TestBegin( UnitTest * testGroup, const char * testName )
     ++s_NumTests;
 
     OUTPUT( " - Test '%s'\n", testName );
+    // Flush the output to ensure that name of the test will be logged in case the test will crash the whole binary.
+    fflush( stdout );
```
ffulin: Nice. This might help us understand what some of the weird sporadic failures are (like the occasional Windows stack overflow that I cannot reproduce locally even if I halve the size of all stacks)

ffulin:
This change has been integrated and will be part of the next source release (v0.99)

PR #601, opened 2019-05-26
Add missing locks around accesses to ClientState

dummyunit:
In two places `ClientState` members were modified without grabbing a lock on `ClientState::m_Mutex`. This lead to data race on `m_NumJobsAvailable` between those writes and the reads in `Server::FindNeedyClients`.

This was found by ThreadSanitizer.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Protocol/Server.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Protocol/Server.cpp
@@ -302,6 +302,7 @@ void Server::Process( const ConnectionInfo * connection, const Protocol::MsgConn
 
     // take note of initial status of client
     ClientState * cs = (ClientState *)connection->GetUserData();
+    MutexHolder mh( cs->m_Mutex );
     cs->m_NumJobsAvailable = msg->GetNumJobsAvailable();
     cs->m_HostName = msg->GetHostName();
 }
@@ -312,6 +313,7 @@ void Server::Process( const ConnectionInfo * connection, const Protocol::MsgStat
 {
     // take note of latest status of client
     ClientState * cs = (ClientState *)connection->GetUserData();
+    MutexHolder mh( cs->m_Mutex );
     cs->m_NumJobsAvailable = msg->GetNumJobsAvailable();
 
     // Wake main thread to request jobs


ffulin:
This appears to be a false positive.

m_NumJobsAvailable is only ever read from the server thread, and only ever written from the connection thread (one thread per connection), and the reads and writes are all atomic on all supported platforms. This member is completely stand alone with on other state inferred from it that would require a lock for synchronization.

When originally writing this code, I deliberately avoided acquiring the lock to reduce any additional contention this might cause between the server thread and the connection thread. I didn't witness any specific cases of contention alleviated by this, but rather I preemptively avoided taking the lock. I should have commented the omission of this lock.

Looking at the code now, all the above still seems to hold true.

I can see a few of options going forward:

**Add the lock**
Pro: Less subtlety in the code.
Pro: Avoids false positives (both human and automated)
Con:  _may_ increase contention unnecessarily

**Ignore the warning, adding a comment to the code**
Pro: No risk of behavior change
Con: Complicates use of the sanitizer for finding other legit bugs

**Mark up the code for ThreadSanitizer**, if this is possible
Pro: No risk of behavior change
Con: Might not be possible?

I'm open to other options as well.

dummyunit:
> This appears to be a false positive.

I wouldn't call this a false positive, just a report that **we** don't want to see.  
ThreadSanitizer doesn't care about rules specific to the architecture the program is running on, its job is to find code that violates rules of the standard. And this is great because it allows to catch and fix bugs that would manifest itself only on some rare architectures without even having access to them to run tests on. But as you noted this is currently of no use to FASTBuild.

> When originally writing this code, I deliberately avoided acquiring the lock to reduce any additional contention this might cause between the server thread and the connection thread.

This is totally reasonable, but it is not obvious from looking at the the code:
1. `m_NumJobsAvailable` is not `volatile`.
2. In `Process` for `MsgNoJobAvailable` it is modified via non-atomic increment, but under the mutex.
3. In `FindNeedyClients` it is read (together with `m_NumJobsRequested`) under the mutex.

So I was under the impression that all accesses to `ClientState` members should happen under the mutex.

> **Add the lock**

It would make handling of `m_NumJobsAvailable` consistent with how it is done in other places in `Server`. But on the other hand we can make it consistent by making accesses in those other places atomic instead :)

On the side of performance side taking a lock would definitely impact performance. But I think that impact would not be noticeable because each lock would be accompanied by a previous read from the network, which itself is not a very fast operation in comparison to mutex lock.

> **Ignore the warning**

That would be bad. Without proper instrumentation for reads/writes that are supposed to be atomic ThreadSanitizer produces dozens of warnings when running the tests. It would be very hard to sort through them each time to find something we are interested in.

> **Mark up the code for ThreadSanitizer**, if this is possible

This is definitely possible, either by doing actual atomic reads/writes using compiler intrinsics (`__atomic_load_n`,  `__atomic_store_n`, etc), or by using special functions provided by TSan runtime. First option is obviously better because it may actually fix some problems instead of providing false sense of security.

Soon I will make a PR that adds TSan build configuration and contains changes necessary to make tests run clean under TSan.

dummyunit:
Actually we might need the lock in `Process` for `MsgConnection` for a different reason.  
That function sets `m_HostName`, which is read from the main thread in FBuildWorker via this call chain: `Worker::Work` -> `Worker::UpdateUI` -> `JobQueueRemote::GetWorkerStatus` -> `WorkerThreadRemote::GetStatus` -> `Server::GetHostForJob`.

ffulin:
I think the flow of network events is such that the hostname is only used after a connection is validly established and the worker is performing work, so I believe this access is also safe. (I didn't do a specific check, so this is from memory and I might be wrong).

I think all things considered, the amount of time being spent just questioning if the code is safe or not indicates we're better of simplifying the problem by simply adding the locks as you have done in your PR. This will avoid potential future problems if access to these things changes in the future.

dummyunit:
> I think the flow of network events is such that the hostname is only used after a connection is validly established and the worker is performing work, so I believe this access is also safe. (I didn't do a specific check, so this is from memory and I might be wrong).

I looked at the code and this seems to be correct for correctly behaving clients: `Server::GetHostForJob` is only ever called on `WorkerThreadRemote::m_CurrentJob` and this is populated only after receiving MsgJob (which should happen after receiving MsgConnection I suppose).
On the other hand misbehaving (or malicious) client may send MsgConnection many times and at any time and the worker will write m_HostName on each received message. But I don't think we can ignore that, there are probably simpler ways to crash the worker.

ffulin:
I decided the best course of action is to integrate this change with the locks. The code is now:
 - obviously safe on basic inspection
 - consistent with other functions
 - more likely to remain safe should some access patterns change
- ThreadSanitizer friendly

Thanks again!

PR #603, opened 2019-05-28
Check SO_ERROR to determine if connect() was successful on Linux

dummyunit:
On Linux `select()` sets a flag in write fd_set both when `connect()` succeeds and when it fails. To determine the result we must get the value of `SO_ERROR` via `getsockopt()`.

Before this change connection attempts to workers that were not running were always treated as successful. For each such attempt a new thread was created and then immediately terminated after a `send()` failure on the non-connected socket.


dummyunit:
The new test failed on OSX on TravisCI, so I assume that OSX has the same behavior for `select()` as Linux. I enabled the new check for `__APPLE__` and now the test is passing on OSX.

ffulin:

PR #604, opened 2019-05-28
Remove possibility of crash in ShutdownMemoryLeak test

dummyunit:
`Helper::AbortBuild` waits for distributed jobs to appear and asserts that it happens in a reasonable amount of time. Unfortunately, because that function runs in a separate thread, if that assert fails then the whole process crashes via unhandled exception.

To fix that `AbortBuild` now doesn't perform assertion itself and just sets a flag which is later asserted in the main thread.

This also fixes a highly theoretical race found by ThreadSanitizer:  
Nothing prevents the main thread from sleeping for 5 seconds after it starts the `AbortBuild` thread and then execute `Tracing::AddCallbackOutput` (via `FLog::StartBuild`) while `AbortBuild` thread tries to write output via `TEST_ASSERT`, thus creating a race on `Tracing::s_CallbacksOutput`.

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
@@ -334,14 +334,18 @@ void TestDistributed::ShutdownMemoryLeak() const
     class Helper
     {
     public:
-        static uint32_t AbortBuild( void * )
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the fixes. I've integrated the changes and they'll be part of the next source release (v0.99)

PR #605, opened 2019-05-28
Add support for building with ThreadSanitizer and fix minor related issues

dummyunit:
ThreadSanitizer helped to found some number of issues in FASTBuild, so it be nice to add support for building with it into main repo and to TSan runs to TravisCI to promptly catch any new issues that might come up.

The main problem with TSan integration is different view on what is an atomic operation in TSan and in the FASTBuild code base. FASTBuild uses the fact that it supports only x86 and therefore uses plain reads/writes instead of atomic reads/writes when synchronization is not required. TSan on the other hand strictly follows the standard requiring atomic operations when a variable is accesses from multiple threads without synchronization.

So the main part of this PR is addition of new helper function to `Atomic.h` and replacement of plain reads/writes with relaxed atomic reads/writes for the variables for which TSan otherwise reports races.  
As always, more details in commit messages.

Also this PR includes changes from #604 because they are needed to fix ShutdownMemoryLeak test when it runs under TSan.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -46,9 +46,22 @@ matrix:
           packages:
             - clang-8
             - llvm-8 # for llvm-symbolizer
+    # Linux, ThreadSanitizer
+    - os: linux
+      env: CFG=tsan CLANG=clang-8
+      # Use latest clang when building with sanitizers.
+      addons:
+        apt:
+          sources:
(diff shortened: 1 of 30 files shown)


ffulin:
Great work as always. This will help robustify FASTBuild going forward!

PR #606, opened 2019-05-28
Fix crashes caused by usage of SmallBlockAllocator while it is in single-threaded mode

dummyunit:
Two tests in TestDistributed were calling `FBuild::Initialize` after they have created a `Server` instance to act as a FBuildWorker replacement.
This lead to a situation where single threaded mode of `SmallBlockAllocator` was enabled while there was more than one thread in the program. When any of this extra threads tried to allocated memory it would trigger assertion and crash the process via unhandled exception. For some reason this looked like a stack corruption error (0xc0000409) in TravisCI builds.

Also because this two tests did nothing useful on Linux/OSX they are moved under ifdef `__WINDOWS__`.

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
@@ -32,8 +32,10 @@ private:
     void TestLocalRace();
     void RemoteRaceWinRemote();
     void AnonymousNamespaces();
-    void ErrorsAreCorrectlyReported() const;
-    void WarningsAreCorrectlyReported() const;
+    void ErrorsAreCorrectlyReported_MSVC() const;
+    void ErrorsAreCorrectlyReported_Clang() const;
+    void WarningsAreCorrectlyReported_MSVC() const;
+    void WarningsAreCorrectlyReported_Clang() const;
     void ShutdownMemoryLeak() const;
     void TestForceInclude() const;
     void TestZiDebugFormat() const;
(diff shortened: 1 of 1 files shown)


ffulin:
Great work! 

Interestingly, there is a safety check in the SmallBlockAllocator, the intent of which is to catch this type of error. It seems that the conidtions of the heavily constrained VM are just right to cause a problem without this check being fired.

I was able to make the safety assert fire by adding a lot of unnecessary small allocations to the Server thread.

ffulin:
This change has been integrated and will be part of the next source release (v0.99). This should greatly improve the Travis CI stability. Thanks again!

dummyunit:
> Interestingly, there is a safety check in the SmallBlockAllocator, the intent of which is to catch this type of error.

From my understanding this assert is what actually caused the crash and not some stack corruption caused by non thread safe usage of `SmallBlockAllocator`.

I have some ideas how to make these asserts that throw exceptions under tests less painful to debug:
1. Don't buffer stdout/stderr output in tests, similarly to FBuild and FBuildWorker.
    This will guarantee that assert will print its message before crashing when that assert happens in non-main thread.
2. Flush output in `OnAssert`, `AssertFailure` and `AssertFailureM` if `Thread::IsMainThread` returns false.
    This would also guarantee that assert message is printed but without unnecessary flushes.

ffulin:
I like the idea of flushing the output in those functions - I think we could safely do that regardless of which thread they occurs on.

Perhaps the assert behavior in the callback(s) for other threads could be changed (to "int 3" instead of throw for example). This would sill need to be coupled with the above, but might result in a less confusing error (assuming the throw is responsible for the stack overflow)

PR #607, opened 2019-05-29
Implement KillProcessTree for Linux/OSX

### LINKED ISSUE #595: [OSX][Linux] Race between FBuild and the compiler when writing to the object file
I have managed to reproduce locally the TestLocalRace failure that was [happening on Travis](https://github.com/fastbuild/fastbuild/pull/450#issuecomment-491411913). From the `strace` log it seems that the following is happening:
1. `fbuild` starts a remote compilation for an ObjectNode and then a local compilation for the same node.
2. Remote compilation completes and `fbuild` receives the result from network.
3. `fbuild` cancels the local job, which results in SIGTERM sent to the compiler.
4. `fbuild` doesn't wait for the compiler process(es) to exit (judging by the comment in `Client::WriteFileToDisk`) and writes the result of the remote compilation to the destination object file.
5. Compiler (assembler `as` to be precise) receives SIGTERM and `unlink`s the destination object file as part of the [cleanup process](https://sourceware.org/git/gitweb.cgi?p=binutils-gdb.git;a=blob;f=gas/as.c;h=c7bdf0cd8896c721f22439c46f0e0adfd2b9c5ff;hb=HEAD#l1294).
6. `fbuild` executes the librarian and it fails to open the object file which doesn't exist anymore.

So it seems that `fbuild` should stop writing to object files while local compiler processes are running. Instead it should either wait for them to exit in the thread that handles remote compilation results or write remote results to the files after waiting for compiler processes to exit in the thread that handles local compilation.  
And I suspect that in both cases a blocking implementation of `Process::KillProcessTree` would be required, otherwise we will wait only for `gcc` to exit and not for `as` started by it.

Here is the `strace` log for anyone interested: https://gist.github.com/dummyunit/b03f0190554831228fceba99257e45ce

dummyunit:
To do the job we put each spawned process into a new process group. This allows us to send SIGKILL to all processes in that group without any kind of process tree traversal. This also works for processes that were reparented to PID 1 due to their parent dying.

This should fix #595.

### DIFF
--- a/Code/Core/Process/Process.cpp
+++ b/Code/Core/Process/Process.cpp
@@ -156,8 +156,8 @@ void Process::KillProcessTree()
             ::CloseHandle( hChildProc );
         }
     #elif defined( __LINUX__ ) || defined( __APPLE__ )
-        // TODO: Kill process tree if necessary?
-        kill( m_ChildPID, SIGTERM );
+        // Kill all processes in the process group of the child process.
+        kill( -m_ChildPID, SIGKILL );
     #else
         #error Unknown platform
     #endif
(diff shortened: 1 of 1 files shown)


ffulin:
Great work! This fix will be in the next release (v0.99). Thanks!

BTW - I was able to reliably reproduce the bug noted in #595 (that is seen frequently on Travis CI) by running inside a VM under heavy CPU load. Running the test would reliably fail before reaching 10 test iterations. With your fix, I was not able to reproduce this issue after several hundred iterations.

PR #612, opened 2019-06-17
Support for forcably isolating files in Unity nodes

wickerwaka:
A common case I am running into while trying to get unity compilation working is that a directory will mostly work fine compiled as a unity file expect for one or two problematic files. I want the management of those problematic files to remain part of the Unity node, rather than having to manage it in both the Unity node and ObjectList node(s). So:
```
Unity( 'Unity-foo' )
{
    .UnityInputPath = 'src\foo'
    .UnityInputExcludedFiles = { 'problematic_file.cpp' }
}

ObjectList()
{
    .CompilerInputUnity = 'Unity-foo'
    .CompilerInputFiles = { 'src\foo\problematic_file.cpp' }
}
```
Can become:
```
Unity( 'Unity-foo' )
{
    .UnityInputPath = 'src\foo'
    .UnityInputIsolatedFiles = { 'problematic_file.cpp' }
}

ObjectList()
{
    .CompilerInputUnity = 'Unity-foo'
}
```


ffulin:
Thanks for the improvement! I've integrated the change and it will be in the next release (v0.99).

I made a small follow up review to minimally update documentation and tests: https://github.com/fastbuild/fastbuild/commit/8c5d4571e9c6357954e157b12a04d033a2bad37b

PR #616, opened 2019-07-01
Require header for CopyDir and RemoveDir

dummyunit:
Header is the only source of the name for these nodes and thus it should be required. Documentation already says that it is required but the code doesn't enforce that.

Without the change this code is silently accepted in Release but triggers assert about non-empty name in `Function::Commit` in Debug:
```
    CopyDir
    {
        .SourcePaths = 'src'
        .Dest = 'dst'
    }
```


ffulin:
Thanks for the fix! This will be in the next release (v1.00)

PR #617, opened 2019-07-01
Fix array self-assignment when concatenating with an empty array of different type

dummyunit:
This code was triggering the assert in `Array::operator=` in Debug and producing wrong results (empty `.B`) in Release:
```
    .S=[]
    .A={.S}
    .A={}
    .B={'foo'}+.A
```


ffulin:
Thanks for the fix! This has been integrated and will be in the next release (v0.99)

PR #618, opened 2019-07-01
Support operators of the If function in Error_1034

dummyunit:
Previously code like this:
```
    .Bool = true
    If (.Bool >= .Bool) {}
```
triggered an assert in Error_1034 function and produced strange looking results in Release: "Operation not supported: 'Bool' ??? 'Bool'".


ffulin:
Thanks for the fix. This has been integrated and will be in the next release (v0.99)

PR #621, opened 2019-07-08
Improve lookup time for LightCache hits

ben-craig:
The old implementation used a hash to select a locked shard, then did a
linear search.  Astring::op== dominated those profiles.

We still use shard to reduce lock contention, but we now use an open
addressed, quadratically probed hash map instead of a linear search.

On a clean, but cached build, I was able to reduce my build time from
146 seconds to 51 seconds, with ~18 seconds devoted to linking.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
@@ -48,6 +48,114 @@ public:
     inline bool operator <  ( const IncludedFile & other ) const    { return ( m_FileName < other.m_FileName ); }
 };
 
+#define LIGHTCACHE_DEFAULT_BUCKET_SIZE 1024
+class IncludedFileHashSet
+{
+public:
+    IncludedFileHashSet( const IncludedFileHashSet & ) = delete;
+    IncludedFileHashSet &operator=( const IncludedFileHashSet & ) = delete;
+    IncludedFileHashSet()
+    {
+        m_Buckets.SetSize( LIGHTCACHE_DEFAULT_BUCKET_SIZE );
+        for( auto &elt : m_Buckets )
+        {
+            elt = nullptr;
+        }
+    }
+    IncludedFile **Find( const AString & fileName, uint64_t fileNameHash )
+    {
+        if( m_Buckets.IsEmpty() )
+           return nullptr;
+        size_t probe_count = 1;
+        size_t startIdx = fileNameHash & ( m_Buckets.GetSize() - 1 );
+        IncludedFile **bucket = &m_Buckets[ startIdx ];
+        while( *bucket != nullptr )
+        {
+            if( (*bucket)->m_FileNameHash == fileNameHash && **bucket == fileName )
+            {
+                return bucket;
+            }
+            bucket = Next( m_Buckets, startIdx, probe_count );
+        }
+        return bucket;
+    }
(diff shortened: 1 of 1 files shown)


ben-craig:
Oh boy.

There's thread safety issues here well beyond the time-of-check-time-of-use bug I fixed in commit 3, even before my change gets involved.  We lock the hash table long enough to protect the state of the hash table itself, but then we return a mutable reference to the content of the hash table (i.e. the IncludedFile).  Multiple threads could all be mutating m_ContentHash or m_Includes concurrently, with really nasty results.

ffulin:
I don't see any issues with thread safety for the previous implementation. Items are never added to the cache until they are complete, and once they are added, are only accessed via a const pointer. I'm not really sure where you are a "mutable reference" to the IncludedFile.

ffulin:
On the subject of your optimization, I'm curious how much of your speedup comes from searching by hash first versus the more complex storage mechanism (which introduces additional memory allocations and lock contention in the memory manager).

ben-craig:
After looking again, I'm not seeing any issues either.  The signature of the Hash function and the population of m_Includes got me worked up about nothing.

There was a failure on the Mac Travis build.  It was talking about a tcp connection failure, but it was in the general vicinity of the cache tests.  That had me a bit concerned that there was some assert or crash that took everything down.  I guess that was just a spurious CI failure?

I did a quick experiment.  I took the old version of the code, but instead of using Array::FindDeref to do the in-bucket search, I iterated over the array and called a new IncludedFile::Equals function that took both the file name and the file hash.  I got the following timings on a clean, but cached build of my code base:

No changes: 113.7 seconds
Hash equality only: 100.2 seconds
Hash table implementation: 49.3 seconds
All of those timings include a ~15 second link step at the tail end.

I think the memory layout of the new scheme and the old scheme is very similar, it's largely the indexing that is different.
* Both schemes start with a C array (old: 128 elements, new: 64 elements).
* The elements of the array went from a Mutex + an Array<IncludedFile *> to a Mutex + (Array<IncludedFile *> + size_t).
* The old Array was dense, and only contained live pointers to IncludedFiles.  The new Array is sparse and has lots of null pointers.
* The old Array was between ~67% and 100% full at any given time and grows 1.5x at a time.  The new Array is between 25% and 50% full at any given time and grows 2x at a time.

The number of allocations should be extremely similar between the two schemes, though the new scheme will definitely use more memory.  There may be a bit more memory manager contention when dealing with racy duplicates though.  Two threads can fail to find the same header, both can populate the IncludedFile, and then each can attempt to add the IncludedFile.  In the old scheme, you would end up with a duplicate entry, which is mostly harmless.  In the new scheme, I delete the duplicate, which does introduce more contention, but only in a fairly rare case.

ffulin:
Thanks for the extra info. I really appreciate the extra effort to obtain the break down of timings. Looks like the lookup is a good optimization, and the small complexity increase is well worth the gain over and above a simple string compare optimization.

> There was a failure on the Mac Travis build. I guess that was just a spurious CI failure?

There are a few occasional failures in the Travis CI which I'm working through. The really low spec and seemingly overloaded VMs have been good for uncovering some real bugs, but also suffer from some weird network problems like not being able to establish a connection to local host after 5 seconds :) The aim of course is to have zero false positive failures. It's likely some of the timeouts will just have to be much higher.

ffulin:
Thanks for the improvement. This has been integrated and will be in the next release (v0.99)

PR #624, opened 2019-07-19
Fix crash in light cache

bburgin:
Use index here, since m_IncludeDefines iterators may be invalidated when recursing

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
@@ -464,8 +464,11 @@ void LightCache::ProcessInclude( const AString & include, IncludeType type )
         {
 
             // Find macro - expand each possible value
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the fix. The array indexing is unsafe as you've identified, but making it traverse the entire range introduces a new problem (we should only be iterating the original range or macros and not those we might have added). That's easily fixed (by iterating by index instead), but looking into it further, I think the macros for includes are not being applied correctly at all anyway.

I'll need to make some more in-depth changes to fix all of these issues. I'll leave this PR open for now to track this issue.

ffulin:
Thanks again for this fix. 

I spent some time investigating this issue and some others with the LightCache, and ended up integrating a fix by Francisco Facioni (https://github.com/fastbuild/fastbuild/pull/716). That fix resolves this issue (iterating an array that can move/resize) and also another problem (incorrect dependency gathering with macro includes)

That change, and thus fixes for both issues, will be in the next release (v1.00)

PR #625, opened 2019-07-28
Strip -MF and -MD tokens when compiling preprocessed source

NEdwards-SumoDigital:
At this point any dependency info produced is bogus, and could stomp on top of valid dependency info produced by the preprocessing pass.

This PR fixes full rebuilds when compiling UE4, due to UE4 using the generated dependency data to figure out when files have changed etc


ffulin:
I have changed the way args are processed to allow most types of command line processing fixes to be done without a need to change the worker protocol or updating the worker.

For example, cases like this one can now be fixed entirely on the client side (the client strips these args before passing them to the worker).

I went ahead and re-applied your fix in that new system: https://github.com/fastbuild/fastbuild/commit/58b9bde1b07e233788b329f41f3e5fd7863b36fb

This fix will be in the next release (v1.05)

(NOTE: I made a slight change vs your fix, which seemed to erroneously strip an extra arg following the -MD. If you are running a local fork, you might want to fix that locally until the new version is available)

PR #626, opened 2019-07-28
Print can now search for what to print based on a pattern and options

mkujawa:
Feature:
  If given a // string instead of a '' or "" string, it will treat the string as a pattern and print the variables that match the wildcard pattern.  Two options are currently provided, to print what's declared in the current scope or to print everything in scope; and to expand arrays/structs or not.

Rationale:
  I was tasked with adding a feature to an existing (large) build system, and figuring out where I was supposed to get certain state from was nigh impossible without a feature like this.  I expect it will almost always be used with the 'P' option (iterate parent scopes), since you usually don't know where you're looking when you invoke this form of Print.

  I had added a setting to a bff, but didn't know what, if any, struct that setting had been stored in to.  By dumping everything in scope, I was able to find it quickly.

Notes:
  I added the pattern matching because it was easy to add, but I always used it with just /*/P+

  The // syntax is borrowed from a variety of sources, but most notably perl.  However, in those uses the pattern is always a regex.  I didn't see an existing regex in fbuild, and I didn't need regex powers, so it's not a regex.  But a future option could enable regex or something; the optons add a lot of power.  'i' would be an obvious option to add; there's even already a MatchI to call.

  I went with 'P' and '+' because they aren't used by the other existing // search precedents out there, and thus shouldn't confuse people.


Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionPrint.cpp
```diff
@@ -124,7 +171,7 @@ FunctionPrint::FunctionPrint()
 
 // PrintVarRecurse
 //------------------------------------------------------------------------------
-/*static*/ void FunctionPrint::PrintVarRecurse( const BFFVariable & var, uint32_t indent )
+/*static*/ void FunctionPrint::PrintVarRecurse( const BFFVariable & var, uint32_t indent, uint32_t maxIndent )
```
mkujawa: I went with `maxIndent` instead of a boolean because I thought it might be useful to specify a struct would only expand one level deep (if `maxIdent == indent+1`), or to control struct depth + stack depth to keep the output from getting too "wide".  I only ended up using it as a boolean, though, and didn't hook up any options that would control width.

Code/Tools/FBuild/Documentation/docs/functions/print.html
```diff
@@ -29,9 +29,26 @@ <h1>Print</h1>
...
+  Prints some text or variable state during BFF parsing.
   </p>
-  <div class='code'>Print( "String or $VarName$" )</div>
+  <div class='code'>Print( "String with optional $VarName$" ) // (1)
+Print( .VarName )                         // (2)
+Print( /String with optional $VarName$/opts )        // (3)</div>
+<ol>
+  <li>Prints a string, embedding any scalar variables included.  To print a $, escape it with a ^.  
+      Arrays and structs cannot be printed this way.</li>
+  <li>Prints a variable.  If the variable is an array or struct, it will recursively print the members.</li>
```
mkujawa: I didn't know about this variation for a long time; I may have only found out about it from looking at the code.

mkujawa:
Seems master was the wrong target branch, trying again

PR #627, opened 2019-07-28
Print can now search for what to print based on a pattern and options

mkujawa:
Feature:
  If given a // string instead of a '' or "" string, it will treat the string as a pattern and print the variables that match the wildcard pattern.  Two options are currently provided, to print what's declared in the current scope or to print everything in scope; and to expand arrays/structs or not.

Rationale:
  I was tasked with adding a feature to an existing (large) build system, and figuring out where I was supposed to get certain state from was nigh impossible without a feature like this.  I expect it will almost always be used with the 'P' option (iterate parent scopes), since you usually don't know where you're looking when you invoke this form of Print.

  I had added a setting to a bff, but didn't know what, if any, struct that setting had been stored in to.  By dumping everything in scope, I was able to find it quickly.

Notes:
  I added the pattern matching because it was easy to add, but I always used it with just /*/P+

  The // syntax is borrowed from a variety of sources, but most notably perl.  However, in those uses the pattern is always a regex.  I didn't see an existing regex in fbuild, and I didn't need regex powers, so it's not a regex.  But a future option could enable regex or something; the optons add a lot of power.  'i' would be an obvious option to add; there's even already a MatchI to call.

  I went with 'P' and '+' because they aren't used by the other existing // search precedents out there, and thus shouldn't confuse people.

branches print_pattern, mkujawa/print_pattern, origin/print_pattern

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/print.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/print.html
@@ -29,9 +29,26 @@
 <div id='print' class='newsitemheader'>Summary</div>
 <div class='newsitembody'>
   <p>
-  Prints some text during BFF parsing.
+  Prints some text or variable state during BFF parsing.
   </p>
-  <div class='code'>Print( "String or $VarName$" )</div>
+  <div class='code'>Print( "String with optional $VarName$" )     // (1)
+Print( .VarName )                             // (2)
+Print( /String with optional $VarName$/opts ) // (3)</div>
+<ol>
+  <li>Prints a string, embedding any scalar variables included.  To print a $, escape it with a ^.  
+      Arrays and structs cannot be printed this way.</li>
+  <li>Prints a variable.  If the variable is an array or struct, it will recursively print the members.</li>
+  <li>(0.99)Prints all variables in the current scope that match the given pattern.  The pattern string may include variables, 
+      as in variant 1. The pattern is a wildcard, not a regex.  Note that variables names start with a period.  
+	  After the closing slash, a set of characters may be included to enable certain options.  The following options are supported:
+      <table>
+        <tr><td valign=top><b>P</b></td><td>: Includes the entire chain of parent scopes.  Output starts with the global variables and then then 
+		                  first child scope in the chain is indented, and so on until the current scope is reached, which 
+						  will be the most indented.  Without P, only variables declared in the current scope will be printed,
+						  and there will be no extra indenting.</td></tr>
+        <tr><td valign=top><b>+</b></td><td>: Recurses into array and struct variables.  Without the +, just the variable name and type are printed.</td></tr>
+      </table></li>
+</ol>
 </div>
 	
 <div id='print' class='newsitemheader'>Details</div>
@@ -56,9 +73,17 @@
 .X64_Debug     = [ .Config = 'Debug'   .Platform = 'X64' ]
 .X64_Release   = [ .Config = 'Release' .Platform = 'X64' ]
 .Configs       = { .X86_Debug, .X86_Release, .X64_Debug, .X64_Release }
+{
+  .X64Configs = { ^X64_Debug, ^X64_Release }
+  .X86Configs = { .X86_Release }
   
+  Print(/.X64*/P)
+  Print(/.X86*/+)
+  Print('----------')
+}
 ForEach( .Config in .Configs )
 {
+  Print( .Config )
   Using( .Config )
   .TargetName = '$Platform$-$Config$'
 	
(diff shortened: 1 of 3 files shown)


Code/Tools/FBuild/Documentation/docs/functions/print.html
```diff
@@ -29,9 +29,26 @@ <h1>Print</h1>
...
+  Prints some text or variable state during BFF parsing.
   </p>
-  <div class='code'>Print( "String or $VarName$" )</div>
+  <div class='code'>Print( "String with optional $VarName$" )     // (1)
+Print( .VarName )                             // (2)
+Print( /String with optional $VarName$/opts ) // (3)</div>
+<ol>
+  <li>Prints a string, embedding any scalar variables included.  To print a $, escape it with a ^.  
+      Arrays and structs cannot be printed this way.</li>
+  <li>Prints a variable.  If the variable is an array or struct, it will recursively print the members.</li>
```
mkujawa: I didn't know about this variant for a long time; I was making for loops to dump out arrays and stuff...

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionPrint.cpp
```diff
@@ -79,6 +79,60 @@ FunctionPrint::FunctionPrint()
...
+                    default:
+                    {
+                        AStackString<32> err;
+                        err.Format("Print pattern option %c", *iterOpt );
+                        Error::Error_1030_UnknownDirective(iterOpt, err);
+                        return false;
+                    }
+                }
+                stop++;
+                stop.SkipWhiteSpaceAndComments();
```
mkujawa: I put this in to consume whitespace between the options and the close parens.  I'm ok with it allowing whitespace between the options, even though that's not "standard"

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionPrint.cpp
```diff
@@ -124,7 +178,7 @@ FunctionPrint::FunctionPrint()
 
 // PrintVarRecurse
 //------------------------------------------------------------------------------
-/*static*/ void FunctionPrint::PrintVarRecurse( const BFFVariable & var, uint32_t indent )
+/*static*/ void FunctionPrint::PrintVarRecurse( const BFFVariable & var, uint32_t indent, uint32_t maxIndent )
```
mkujawa: I went with `maxIndent` instead of a boolean because I thought it would be useful to mimic the behavior of `dt -r2` in windbg, where it will stop expanding members after a certain depth (`maxIndent = indent+2`, for example) or to just constrain how far to the right stuff goes for readability (`maxIndent = 5` or something.)  I didn't end up making those options, but the power is there for the future

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionPrint.cpp
```diff
@@ -169,27 +227,79 @@ FunctionPrint::FunctionPrint()
...
+        --iFrame;
+        const BFFStackFrame* const pFrame = aFrames[ iFrame ];
+        const uint32_t iDepth = iNumFrames - iFrame - 1;
+        const Array< const BFFVariable* >& rVariablesInFrame = pFrame->GetLocalVariables();
+        for ( size_t iVar = 0; iVar < rVariablesInFrame.GetSize(); ++iVar )
+        {
+            const BFFVariable* const pVar = rVariablesInFrame[ iVar ];
+            if ( bPatternMatchesAll || pVar->GetName().Matches( pattern.Get() ) )
+            {
+                PrintVarRecurse(*pVar, iDepth, iMaxDepth);
```
mkujawa: The existence of this helper function is one of the reasons I opted to make this a power of Print instead of adding a new function

ffulin:
Thanks for the PR. This feature adds a brand new syntax to the parsing language, specific to only one Function (Print), and as such there is quite a high bar to reach for it to be accepted. I'd like to understand your problem space a little better, to ensure that:
 - that this problem isn't better solved in another way
 - that this feature doesn't encourage bad patterns in the bff

*Use-case*

>  figuring out where I was supposed to get certain state from was nigh impossible without a feature like this. 

Maybe I've misunderstood, but this comment implies that the state you end up with inside any leaf function in your bff is being assembled by state that is scattered around at various parent scopes or the global namespace. The use of variable scope and shadowing can be good for some cases, but in general I would advise instead to built state through the use of Structures instead.

Then you can end up with bff such as:
```C++
ForEach( .Config .Configs )
{
    Using( .Config )
```
where each .Config encapsulates all the state. You can then Print( .Config ) inside the loop.

Even in the shadow/override case, I don't quite understand why you'd want to print variables matching a pattern. Perhaps you can provides some more information about this problem?

*Tests*

Thankyou for updating the documenation along with this change. In addition, a new feature such as this would require tests to be written before it could be integrated. Having tests as part of the PR makes integration significantly fasted.

Having said this, I would hold off spending any time on writing tests until we have ensured the feature is needed/desirable.

mkujawa:
> Maybe I've misunderstood, but this comment implies that the state you end up with inside any leaf function in your bff is being assembled by state that is scattered around at various parent scopes or the global namespace. The use of variable scope and shadowing can be good for some cases, but in general I would advise instead to built state through the use of Structures instead.
> 
I'm not going to re-write our build system, even if a cleaner way exists.  I coded this as a way to help understand an existing build system.  

I understand if this isn't the approach you want to take to solve this problem; the PR exists in large part to get you thinking about how someone can understand what's going on in an large, existing bff set.

> Then you can end up with bff such as:
> 
> ```c++
> ForEach( .Config .Configs )
> {
>     Using( .Config )
> ```
> 
We do have some things like that.  But the next line will be `#include ConfigCommon.bff` or somesuch.  It can be difficult to know, from inside of some bff, what all is in scope.

Our bffs could be simplified a lot, I suspect, now that `If()` is a thing.  We don't even use variable indirection anywhere. But without spending a couple weeks rewriting all our build rules, what I needed was something to understand how to modify a part of it.

PR #628, opened 2019-07-28
Copy on a symlink didn't fix the target path

mkujawa:
Previously, instead of the default behavior of following a link and copying the contents to a new file, CopyFile attempted to make a new link pointing at the same place.

But it did this by copying the link contents without modification, and a link's contents can be a relative path.  So if /a/b/foo was a symlink that contained simply "bar", then copying it to /a/d would result in a broken link (there is no /a/d/bar).

There may be situations where that is the desired behavior, but that really needs not to be the default.  Even with this change, there may be times that following the link is preferable to creating a new link (copying to removable media or a network share, for some examples.)  But I think the "normal" desire is that you can copy a path and the result never a broken link.

This change makes the link target always be an absolute path, starting from /, so that it will work regardless of where the link is copied to.  So a/d/foo would be a link containing "/a/b/bar"

### DIFF
--- a/Code/Core/CoreTest/Tests/TestFileIO.cpp
+++ b/Code/Core/CoreTest/Tests/TestFileIO.cpp
@@ -150,7 +150,7 @@ void TestFileIO::FileCopySymlink() const
         // by the file copy API.  Also on Windows, it would make unit
         // tests require administrator privileges.
     #elif defined ( __LINUX__ )
-        AStackString<> symlinkTarget( "symlink" );
+        AStackString<> symlinkTarget( "/symlink" );
 
         // generate a process unique file path
         AStackString<> path;
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -24,6 +24,7 @@
     #include <libgen.h>
     #include <limits.h>
     #include <stdio.h>
+    #include <string.h>
     #include <sys/stat.h>
     #include <unistd.h>
 #endif
@@ -222,6 +223,20 @@
             return false;
         }
         linkPath.SetLength( length );
+
+        if (linkPath.Get()[0] != '/' )
+        {
+            // Convert linkPath to a full path in case the relative path
+            // is wrong from dstFileName
+            if ( srcFileName[0] != '/' )
+            {
+                return false;
+            }
+            const char *pLastSlash = strrchr(srcFileName, '/');
+            const size_t lenSrcPath = (pLastSlash - srcFileName) + 1;
+            linkPath.Format("%.*s%s", (int)lenSrcPath, srcFileName, linkPath.Get());
+        }
+
         return symlink( linkPath.Get(), dstFileName ) == 0;
     }
 


Code/Core/FileIO/FileIO.cpp
```diff
@@ -222,6 +223,20 @@
             return false;
         }
         linkPath.SetLength( length );
+
+        if (linkPath.Get()[0] != '/' )
+        {
+            // Convert linkPath to a full path in case the relative path
+            // is wrong from dstFileName
+            if ( srcFileName[0] != '/' )
```
mkujawa: This was always true in my tests; I believe it's guaranteed that fbuild has expanded `srcFileName` and `dstFileName` to absolute paths by this point?  Need advice on what, if any, check is appropriate here and what to do if the check fails.

Code/Core/FileIO/FileIO.cpp
```diff
@@ -24,6 +24,7 @@
     #include <libgen.h>
     #include <limits.h>
     #include <stdio.h>
+    #include <string.h>
```
mkujawa: For `strrchr`

Code/Core/FileIO/FileIO.cpp
```diff
@@ -222,6 +223,20 @@
...
+        {
+            // Convert linkPath to a full path in case the relative path
+            // is wrong from dstFileName
+            if ( srcFileName[0] != '/' )
+            {
+                return false;
+            }
+            const char *pLastSlash = strrchr(srcFileName, '/');
+            const size_t lenSrcPath = (pLastSlash - srcFileName) + 1;
+            linkPath.Format("%.*s%s", (int)lenSrcPath, srcFileName, linkPath.Get());
```
mkujawa: `%.*s` is a bit obscure, is it ok to use?

Code/Core/CoreTest/Tests/TestFileIO.cpp
```diff
@@ -150,7 +150,7 @@ void TestFileIO::FileCopySymlink() const
         // by the file copy API.  Also on Windows, it would make unit
         // tests require administrator privileges.
     #elif defined ( __LINUX__ )
-        AStackString<> symlinkTarget( "symlink" );
+        AStackString<> symlinkTarget( "/symlink" );
```
mkujawa: My changes don't alter a link that is already an absolute path, so making the link target an absolute path so the unit test will pass.  If the target is just "symlink", then Copy() will fix it to an absolute path, and it will no longer be identical.  

Because the source and destination are in the same directory, the old behavior was also correct, and Copy() could be made smart enough to detect this (At which point I'd use PathUtils to make a relative path), but I KISS for now.

Code/Core/FileIO/FileIO.cpp
ffulin: The higher level code during the build deals with full paths in most cases, but this lower level code shouldn't assume the path passed in is a full path.

Code/Core/FileIO/FileIO.cpp
ffulin: This is probably fine, but it might be better to use Append() to construct the string. It'll certainly be more performant if nothing else (not that this is of particular concern here)

ffulin:
I think whether you want relative or absolute resolution of the links when copying comes down to the use-case. I think we need to retain the existing behavior for backwards compatibility, but we could expose an option on the Copy() to activate this newbehavior.

mkujawa:
> I think whether you want relative or absolute resolution of the links when copying comes down to the use-case. I think we need to retain the existing behavior for backwards compatibility, but we could expose an option on the Copy() to activate this new behavior.

Backwards compatibility with what? The whole reason I've made a PR is that this behavior changed on me.  :)

I agree that an option is the best course of action, but that still leaves the question of what the behavior should be if no option is specified.  I think having `CopyFile` work like `cp` is the default of "least surprise", even though it's not the current behavior.  At least with an option, there is an immediate remediation for anyone relying on the current behavior.

mkujawa:
This PR has some major flaws:
1) I didn't change the OSX behavior to match linux.  (Windows doesn't have relative links, so it didn't need updating.)
2) I didn't handle symlink needing to overwrite an existing link (I think this is a bug even without my changes)
3) As stated above, this isn't going to be the desired behavior for everyone (like whatever inspired the change in CopyFile behavior), so some configuration is needed

Replacing with https://github.com/fastbuild/fastbuild/pull/667

PR #629, opened 2019-07-29
New function, TextFile, for creating text files

mkujawa:
The envisioned use of this function is the creation of response files with compiler parameters and text files with build system constants in it.  I'm sure there will be other uses once it's "in the wild."

* Command-line length is still an issue on windows, and the "solution" is to put the command-line arguments into a text file (for the tools that support it, such as cl.)  Until fastbuild is generating these "response files" automatically, the TextFile function will allow people to do that generation themselves.

* Some languages do not have a facility for passing define macros to the compiler, and the best equivalent is to generate a tiny source file with those constants in it.  (This is the use case I authored this for.)

* TextFile could be used to "touch" a file to trigger some other activity (such as an auto deploy.)

I didn't provide any facility to include existing data at this time because that's so much more complicated and it's not important to my use case.


Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionTextFile.h
```diff
@@ -0,0 +1,22 @@
+// FunctionTextFile
```
mkujawa: This was based on FunctionExec

Code/Tools/FBuild/FBuildCore/Graph/TextFileNode.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "Core/FileIO/FileIO.h"
+#include "Core/FileIO/FileStream.h"
+#include "Core/Math/Conversions.h"
+#include "Core/Strings/AStackString.h"
+#include "Core/Process/Process.h"
+
+// Reflection
+//------------------------------------------------------------------------------
+REFLECT_NODE_BEGIN( TextFileNode, Node, MetaName( "TextFileOutput" ) + MetaFile() )
+    REFLECT_ARRAY( m_TextFileInputStrings, "TextFileInputStrings", MetaNone() )
```
mkujawa: I considered making TextFileInputStrings optional: it would just be a file with no contents.  However, I figured there was basically no situation where a 3-byte file wasn't just as good as an empty file, and it might be super frustrating if someone had set `TextFileInputstrings` or something which was just silently ignored and their file was empty.

mkujawa: What does make me sad is that empty strings aren't valid in an array of strings, but that's not something I can control from here

Code/Tools/FBuild/FBuildCore/Helpers/Report.cpp
```diff
@@ -45,6 +45,7 @@ uint32_t g_ReportNodeColors[] = {
                                   0x77DDAA, // SLN_NODE
                                   0x77DDAA, // XCODEPROJECT_NODE
                                   0x000000, // SETTINGS_NODE (never seen)
+                                  0xFFFFFF, // TEXT_FILE_NODE
```
mkujawa: No idea how the colors are chosen, so I copied COPY_FILE

ffulin:
Adding new node types to FASTBuild is something that must be done with very careful consideration, as once the functionality is released, it becomes close to impossible to remove should a better pattern emerge and carries a significant backward compatibility overhead.

To be sure I understand the use cases behind this PR, could you please provide some more information?

> Until fastbuild is generating these "response files" automatically

FASTBuild already supports response files and generates them for tools that support it, on platforms where it's necessary. This is mostly derived by FASTBuild knowing if a specific tool supports this mechanism, and this detection can be extended or manually controlled for any places it can't be auto-derived. This seems like a better option than a generic TextFile command, as it doesn't bypass the normal options used for command lines which various things depend on (dependency extraction, Intellisense, distributed compilation andcaching to name just a few).

> Some languages do not have a facility for passing define macros 

Do you have en example (a specific language and use-case) that you could provide some more details about?

> TextFile could be used to "touch" a file 

Do you have a specific use-case for this in mind you could provide some more details about?

mkujawa:
> FASTBuild already supports response files and generates them for tools that support it, on platforms where it's necessary. This is mostly derived by FASTBuild knowing if a specific tool supports this mechanism, and this detection can be extended or manually controlled for any places it can't be auto-derived. This seems like a better option than a generic TextFile command, as it doesn't bypass the normal options used for command lines which various things depend on (dependency extraction, Intellisense, distributed compilation andcaching to name just a few).

Good to know! And good point about where fastbuild parses the options.   Where in the fbuild code does that happen?  Can it be enabled for `Exec` nodes? If Exec nodes can be made command-line length aware, I can always make a TextFile.exe (and a linux and osx version...) and then `Exec` it. 

> > Some languages do not have a facility for passing define macros
> 
> Do you have en example (a specific language and use-case) that you could provide some more details about?
> 
  The feature I wrote this for generates a .cs file.  csc supports `-define`, which can be used to set a symbol for `#if`.  But C# can't set `BUILD_TYPE=Debug` on the command-line like C++ can.  So I'm generating a C# class with `public static readonly` variables that are based on build variables.

  Another time I had a similar need was generating gradle files for android APKs.  I did that by using `Exec` to run a perl script and passing it what I wanted it to do on the command-line.  I didn't do that for the cs file because
1) The build machines don't have perl, and we don't want to install it there
2) I didn't want to run into command-line length limits on windows
So I'd rather replace the perl script with TextFile if I can.

  Another place I've used a TextFile-like feature in the past was generating .rc files for the version resource in a windows exe.

  This feature is in msbuild: https://docs.microsoft.com/en-us/visualstudio/msbuild/writelinestofile-task

> > TextFile could be used to "touch" a file
> 
> Do you have a specific use-case for this in mind you could provide some more details about?

This was pretty speculative; pure brainstorm.  But the thinking was that the last build step, after everything else has succeeded, could touch a file as a way of signalling another process that the build succeeded.  I was drawing inspiration from the way some organizations use git commit hooks.  At a previous job, they did have a file system change monitor running that would know to run certain tools when certain files changed.  

Another thing I've seen in my past was a system that looked at the build log to determine if the official nightly had finished, and was a success.  It would occasionally get that wrong, and then proceed to use a failed build as if it were a good build.  Having a "ThisIsAGoodBuild.txt" would have simplified something like that.  Something like this could be done with the current functions (by building a dummy exe last), it just wouldn't be a text file and couldn't have any build system state dumped into it (like head SHA/changelist or whatever.)

ffulin:
I've integrated your changes. They will be in the next release (v1.00. Thanks :)

I added some basic tests and extended the documentation to warn about unsafe use of TextFile (it can be very dangerous if used to generate response files).

PR #633, opened 2019-08-29
Add .Environment field for TestNode

bburgin:
This change allows users to pass env variables to tests.
An example use case is sharding googletest exes using the following env variables:
GTEST_SHARD_INDEX=0
GTEST_TOTAL_SHARDS=16


ffulin:
Exposing the Environment on the Test is a good feature, but I think this needs to be done in a way that is consistent with the other functions that expose it, which is that it overrides rather than extends the base environment.

bburgin:
> 
> 
> Exposing the Environment on the Test is a good feature, but I think this needs to be done in a way that is consistent with the other functions that expose it, which is that it overrides rather than extends the base environment.

Good point. I updated the code to overwrite the env, instead of merging it.

ffulin:
Thanks for the improvement! This has been integrated and will be in the next version (v1.00)

PR #636, opened 2019-09-10
Periodically update modified time of brokerage files for workers that are still active

bburgin:
This change allows an external script to easily delete older brokerage files (clean up orphaned ones from crashed workers). No downtime is needed for active workers, since the script can be made to only delete older brokerage files, based on the modification time of the files.


Code/Tools/FBuild/FBuildCore/WorkerPool/WorkerBrokerage.cpp
```diff
@@ -146,9 +146,26 @@ void WorkerBrokerage::SetAvailability(bool available)
...
                 FileIO::EnsurePathExists( m_BrokerageRoot );
 
-                // create file to signify availability
-                FileStream fs;
-                fs.Open( m_BrokerageFilePath.Get(), FileStream::WRITE_ONLY );
+                // Periodically update the modified time of brokerage file for workers that are still active,
+                // so an external script can easily delete older brokerage files (clean up orphaned ones
+                // from crashed workers).
+                const bool timeSetOK = FileIO::SetFileLastWriteTimeToNow(
+                    m_BrokerageFilePath );
```
ffulin: The behaviour change seems reasonable, but I think there are 2 small problems with this PR:
a) we'll only get here if the brokerage file doesn't exist, so once it's created, we will no longer update the timestamp.
b) if we fail to set the time due to failed network connectivity, we don't reset the timer which means the next update will immediately try again. This could cause the worker to hit multiple network timeouts in a row.

I think we could solve these problems and simplify this entire block of logic a bit by doing something like the following:
`if ( elapsedTime >= 10000.0f )
{
    // Try to update time (common case is that file already exists)
    const bool timeSetOK = FileIO::SetFileLastWriteTimeNow( m_BrokerageFilePath );
    if ( timeSetOK == false )
    {
        // Create token for the first time
        FileIO::EnsurePathExists( m_BrokerageRoot );
        FileStream fs;
        fs.Open( m_BrokerageFilePath.Get(), FileStream::WRITE_ONLY );
    }

    // Restart the timer
    m_TimerLatUpdate.Start();
}`

What do you think?

bburgin: Thanks for the suggestions. I updated the diff. Thoughts?

ffulin:

I made some follow up changes (https://travis-ci.com/fastbuild/fastbuild/builds/148050292) to fix some issues that have come about from the interaction of various changes in this code:
 - avoid updating the time if we're going to re-create the file anyway
 - avoid ensuring the directory exists twice if the brokerage file doesn't exist
 - fix missing timer reset when network can't be accessed

PR #639, opened 2019-10-01
Fix GetRelativePath case sensitivity

ohoughton:
GetRelativePath is currently case sensitive on all systems.  As per the other code in PathUtils, change GetRelativePath to assume case insensitivity on windows and macosx.  This can solve issues when using. say, file paths which are using different cases for the drive letter on windows.
Due to the requirements of some of the current tests, preserves case in the output relative file path.

### DIFF
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -220,7 +220,7 @@
 
     #if defined( __WINDOWS__ ) || defined( __OSX__ )
         // Windows & OSX: Case insensitive
-        if ( ( compA >= 'A' ) && ( compB <= 'Z' ) )
+        if ( ( compA >= 'A' ) && ( compA <= 'Z' ) )
         {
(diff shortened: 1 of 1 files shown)


Code/Core/FileIO/PathUtils.cpp
```diff
@@ -215,17 +215,47 @@
...
+        compB = *itB;
         if ( dirToken )
         {
             pathA = itA;
             pathB = itB;
         }
+        
+        #if defined( __WINDOWS__ ) || defined( __OSX__ )
+            // Windows & OSX: Case insensitive
+            if ( ( compA >= 'A' ) && ( compB <= 'Z' ) )
```
ffulin: This should be compA as well

ffulin:
Thanks for the fix. This has been integrated and will be in the next release (v0.99). I've also added some tests to cover the problems you found (https://github.com/fastbuild/fastbuild/commit/36964e7e0cf293fba5e6aa75b8302d131def4f66)

PR #641, opened 2019-10-08
Add UseRelativePaths Object/ObjectList option

ohoughton:
Things to think about -- look for other potential places using full paths that this feature might wish to adjust (and whether it should adjust them), such as the compiler spawn exe.

Would add an option to the ObjectList or Object nodes to force the usage of relative paths during compilation, as well as during the LightCache caching process.  When set to false (the default), the compilation process should execute as it did prior to this feature existing.  When enabled, adjusts includes and other file paths to be relative if they aren't already.

This feature should help in the look forward towards reproducible builds, as well as help make caching across machines easier to accomplish.


ohoughton:
I note that much of the functionality provided by this pull requested is duplicated in recent commit ca96606.  @ffulin should I continue to maintain this?  The most major differences currently are the additions to LightCache and the include path replacement, from what I can gather through code inspection.

I am currently making use of this pull request's feature, so additionally I'd like to maintain as much functionality as possible while still being prepared for your future changes on this.

ffulin:
There is definitely some overlap. I'd like to keep this issue open to continue to think about some of the things this addresses that the PR is submitted does not (like the LightCache changes for example).

One thing I'm curious about is the part of this PR that makes include paths relative. It's usually the case that include paths are already relative, so I'm not entirely sure why this is needed. Was this just for completeness? Or do you have a situation where you include paths are full paths for some reason?

ohoughton:
Right now, our build system autogenerates a number of full paths as includes due to wildly varying build paths.  I can see the argument for that being specific to our system and not necessarily generally applicable; however, if someone is enabling this feature, I find it difficult imagine they wouldn't want to force the include paths to be relative.  They won't get the relative caching benefits without it.  Following a somewhat "better safe than sorry" mentality.

I haven't directly tested the efficiency impact of adjusting the include paths -- it should be minimal, but I could see anything notable being a good reason to not adjust them.

ohoughton:
Abandoning -- outdated.

PR #644, opened 2019-10-15
Fix the logic for retries after remote system errors.

samhocevar:
This tries to address a problem described in #437.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Protocol/Client.cpp
@@ -485,7 +485,7 @@ void Client::Process( const ConnectionInfo * connection, const Protocol::MsgJobR
 
     // Has the job been cancelled in the interim?
     // (Due to a Race by the main thread for example)
-    Job * job = JobQueue::Get().OnReturnRemoteJob( jobId );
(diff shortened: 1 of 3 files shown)


ffulin:
I finally found some time to look into this issue. Thanks for all the info you provided - it was very helpful.

The core of the issue seems to be that when a remote failure that is a "system failure" occurs and that task is being built locally in a "race", the local task is cancelled when it shouldn't be.

This PR avoids that case, but introduces a new problem, which is that it completely disables the blacklisting mechanism for remote workers. Addiitionally, I fear that it will result in memory leaks.

Armed with an understanding of what's going on, I spent some time on an alternative bug fix, but it quickly spiralled into something quite complex. I don't yet have a good way to address this issue. The code involved is complex and nuanced so without a clean fix, I'm worried about introducing other regressions.

We certainly want to fix this, but it'll require some more thought. Even though I don't have a solution, I wanted to comment so you know this issue is not being ignored.

vinesworth:
BTW, Optionally disabling blacklisting might be a useful issue on its own. I'm currently facing a situation where the remote worker's compilers occasionally crash with OOMs, and there's not much I can do about it (no free memory slots in the machine). So I'd rather want fbuild to keep feeding the worker with jobs, since most of the succeed anyway.

ffulin:
This issues this PR was intended to cover have been resolved as of v1.06 but I missed closing this when that version was released.

PR #647, opened 2019-10-22
VS2019 build fixes for build tools v. 14.23.28105

lucianm:
Hello, I was only able to successfully build FASTBuild under VS2019 Enterprise with the latest build tools 14.23.28105 after this explicit cast and disabling a warning.


ffulin:
Thanks for letting me know about this problem. I could not integrate your pull request as it break OSX and Linux. I've fixed these issues here: https://github.com/fastbuild/fastbuild/commit/fb6eaff74b5f7bb6b69d3cd2ccd1f06e3c19b21b (notably, that change improves the TEST_ASSERT macro to avoid future false positives, while still working on all platforms)

PR #648, opened 2019-10-22
VS2017/2019 and Win10 SDK detection, paths and versions override possibility

lucianm:
These modifications to the Windows VS2017 / VS2019, as well as the Windows10 SDK BFF files allow overriding installation paths and versions via environment variables.

For convenience, 2 scripts are included, which based on the VS installer tool _**vswhere.exe**_ detect VS2017 / VS2019 respectively, for installation paths, Windows 10 SDK version and set appropriate environment variables which are then imported in the FASTBuild BFF build system under Windows.

Only a valid PATH set to a working FBuild.exe is then needed. Sample workflow:

`set "PATH=__LOCATION_OF_EXISTING_FBUILD_EXECUTABLE__;%PATH%"`

Change to the "Code" directory and execute the following (output is included here):
`cd D:\Src\Git\_OSS\_mine\fastbuild\Code

D:\Src\Git\_OSS\_mine\fastbuild\Code>..\External\SDK\VisualStudio\_pathVS2019.cmd
Executing "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\Tools\\vsdevcmd\core\winsdk.bat" ...
################################################################################
"my_VS2019_BasePath=C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise"
"my_DevEnv_2019=C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise\Common7\IDE\devenv.exe"
"my_Windows10_SDKBasePath=C:\Program Files (x86)\Windows Kits\10"
"my_Windows10_SDKVersion=10.0.18362.0"
################################################################################`

Now before actually starting this instance of VS2019, the desired build tools version can be specified:
`set "my_VS2019_Version=14.23.28105"

D:\Src\Git\_OSS\_mine\fastbuild\Code>%my_DevEnv_2019%
`
Now load the FASTBuild solution you should have generated previously in the stated VS instance, it should be buildable out of the box with the help of the old existing executable you placed in the PATH environment, Enjoy ;-)


belkiss:
I like this! It would be great to have this ability indeed to specify the path to the various toolchains and SDKs, and it could also be used to replace occurences of
```
#if CI_BUILD
    .VS2017_BasePath     = 'C:\Program Files (x86)\Microsoft Visual Studio\2017\BuildTools'
#endif
```

Only issue I currently have with this PR is that the env var doesn't propagate to the tests, hence they fail because in my case VS2017_BasePath still has the default value.

fsfod:
Could the VS version environment variable be populated from VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt if unset that instead of having to manually find the version.

lucianm:
> 
> 
> Could the VS version environment variable be populated from VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt if unset that instead of having to manually find the version.

Hi, that's a good idea, but however, the possibility to override that is also important, I'll consider this. Sometimes it's important to be able to stick with a certain toolset version (for example when using a certain CUDA version).

lucianm:
@belkiss:
I've reworked a bit my PR and force-pushed it, could you please check if you're still experiencing this:
> Only issue I currently have with this PR is that the env var doesn't propagate to the tests, hence they fail because in my case VS2017_BasePath still has the default value.

lucianm:
@fsfod:

I've analized this:
> Could the VS version environment variable be populated from VC\Auxiliary\Build\Microsoft.VCToolsVersion.default.txt if unset that instead of having to manually find the version.

and found out that the version written to that file does not correspond to the current _**toolset**_ version from my current up-to-date VS2019 installation: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.25.28610`, but the **_redistributable_** version available here: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Redist\MSVC\14.25.28508`. So it seems to me that parsing that file is not feasible.

PR #657, opened 2019-11-05
Add more special characters to XCodeProjectGenerator

jblazquez:
It looks like Xcode also treats the `+`  sign as a special character that requires quoting.


ffulin:
Thanks for the fix. This change has been integrated and will be in the next release (v1.00)

PR #667, opened 2019-12-03
[wip] Attempting to give more control over the behavior of symlinks when co…

mkujawa:
This PR supercedes https://github.com/fastbuild/fastbuild/pull/628

I have an sdk with libfoo.so, and libfoo.so.10, with libfoo.so as a symlink to libfoo.so.10.  My build rules from 0.94 had a CopyFile from libfoo.so to a distribution folder which housed all the files needed to run the build result.  This now just creates a broken link in that distribution folder instead of a usable .so file.  From the point of view of my build, the current behavior is buggy.

My previous attempt to fix this had numerous flaws, not the least of which is that the build can't run twice because it can't overwrite the link (I think this is a problem with the current implementation as well!)  So, another reason why the current behavior is buggy?

I've attempted to make good on our discussion in the previous PR, and have added some options to CopyFile.  There's still some work to do to finish this up (documentation, more testing), but I wanted to get your thoughts on it before working on it more.

In this PR, I've set the default behavior back to the 0.94 behavior, which is the same as the behavior of `copy` or `cp`, as I think those define what will be "least surprise".

### DIFF
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -15,6 +15,7 @@
 
 // system
 #if defined( __WINDOWS__ )
+    #include <winioctl.h>
     #include "Core/Env/WindowsHeader.h"
     #include "Core/Time/Time.h"
 #endif
@@ -70,6 +71,23 @@
         FuncPtr         m_FuncPtr       = nullptr;
     } gOSXHelper_utimensat;
 #endif
+
+/*static*/ FileIO::SymlinkBehavior FileIO::ParseSymlinkBehavior( const AString& value )
+{
+	if ( value.IsEmpty() || value.CompareI( "FollowSrcAndDest" ) )
+	{
+		return FOLLOW_SRC_AND_DST;
+	}
+	if ( value.CompareI( "FollowSrc" ) )
+	{
+		return FOLLOW_SRC;
+	}
+	if ( value.CompareI( "NoFollow" ) )
+	{
+		return FOLLOW_NEITHER;
+	}
+	return INVALID_SYM_LINK_BEHAVIOR;
+}
 
 // Exists
 //------------------------------------------------------------------------------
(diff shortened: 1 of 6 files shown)


Code/Core/FileIO/FileIO.cpp
```diff
@@ -71,6 +72,23 @@
     } gOSXHelper_utimensat;
 #endif
 
+/*static*/ FileIO::SymlinkBehavior FileIO::ParseSymlinkBehavior( const AString& value )
+{
+	if ( value.IsEmpty() || value.CompareI( "FollowSrcAndDest" ) )
```
mkujawa: This is where to change the default behavior of CopyFile in bff code.  If you want it to match the new way and not the normal way, you'd have IsEmpty return `FOLLOW_NEITHER`

Code/Core/FileIO/FileIO.cpp
```diff
@@ -147,12 +165,41 @@
...
+			if ( DeviceIoControl( hFile, FSCTL_GET_REPARSE_POINT, NULL, 0, buf, sizeof( buf ), &bufSize, NULL ) )
+			{
+				ULONG* pTag = reinterpret_cast<ULONG*>( buf );
+				isSymLink = ( *pTag == IO_REPARSE_TAG_SYMLINK );
+			}
+			CloseHandle( hFile );
+
+			// CopyFileEx only supports FOLLOW_SRC_AND_DST and FOLLOW_NEITHER
+			// based on the COPY_FILE_COPY_SYMLINK flag.  
+			// (and COPY_FILE_COPY_SYMLINK isn't really FOLLOW_NEITHER)
```
mkujawa: The documentation is that the new link will point to the same file as the old link, which leads me to believe that even if you make a relative symlink, COPY_LINK isn't going to produce a link relative to the new path.  But I haven't actually tested that assumption

Code/Core/FileIO/FileIO.cpp
```diff
@@ -198,7 +245,15 @@
...
-    bool result = ( copyfile( srcFileName, dstFileName, nullptr, COPYFILE_DATA | COPYFILE_XATTR | COPYFILE_NOFOLLOW ) == 0 );
+    copyfile_flags_t flags = COPYFILE_DATA | COPYFILE_XATTR;
+    switch ( linkBehavior )
+    {
+    case FOLLOW_NEITHER:
+        flags |= COPYFILE_NOFOLLOW_SRC;
+    case FOLLOW_SRC:
+        flags |= COPYFILE_NOFOLLOW_DST;
+    }
+    bool result = ( copyfile( srcFileName, dstFileName, nullptr, flags ) == 0 );
```
mkujawa: I'm not able to test OSX

Code/Core/FileIO/FileIO.cpp
```diff
@@ -212,17 +267,39 @@
...
+            {
+                return false;
+            }
+            linkPath.SetLength( length );
+            if ( symlink( linkPath.Get(), dstFileName ) == 0 )
+            {
+                return true;
+            }
+            // symlink will not clobber.
+            if ( errno == EEXIST )
```
mkujawa: I think this EEXIST test is probably a needed fix even without the rest of this PR

Code/Core/FileIO/FileIO.cpp
```diff
@@ -234,11 +311,26 @@
...
+    if ( noDstFollow )
+    {
+        dstFlags |= O_NOFOLLOW;
+    }
+    int dest = open( dstFileName, dstFlags, 0644 );
     if ( dest < 0 )
     {
-        close( source );
-        return false;
+        if ( ( errno == ELOOP && noDstFollow ) &&
```
mkujawa: This probably only comes up when changing link behavior in a CopyFile all, like will happen for me when I go back to the old behavior in my rules

Code/Tools/FBuild/FBuildCore/Graph/CopyFileNode.cpp
```diff
@@ -21,6 +21,7 @@ REFLECT_NODE_BEGIN( CopyFileNode, Node, MetaNone() )
     REFLECT(        m_Source,                   "Source",                   MetaFile() )
     REFLECT(        m_Dest,                     "Dest",                     MetaPath() )
     REFLECT_ARRAY(  m_PreBuildDependencyNames,  "PreBuildDependencies",     MetaOptional() + MetaFile() + MetaAllowNonFile() )
+    REFLECT( m_SymlinkBehavior, "SymlinkBehavior", MetaOptional() )
```
mkujawa: Naming is hard

Code/Core/FileIO/FileIO.cpp
mkujawa: Also welcome any feedback on these names.  Are there other examples of reflected "enums"?

mkujawa:
https://travis-ci.com/fastbuild/fastbuild/jobs/262587667 is failing because of a bug in the 17134 Windows 10 headers that CI uses.  Non-CI does not have this issue as 17763 has fixed this (changed `_Out_writes_bytes_` to `_Inout_updates_bytes_` for `Input` of `DeviceDsmAddDataSetRange`)

PR #671, opened 2019-12-07
add support for PCH node reuse

fran6co:


ffulin:
Thanks for the PR. This outcome of this is definitely desirable, but I think we want a slightly different implementation.

As the PR stands, it doesn't test all the necessary properties between the two PCH definitions for equality, allowing for potential inconsistencies that would be hard to diagnose. For example, it doesn't test PreBuildDependencies, AllowCaching and a few others.

Rather than allowing redundant definitions of the PCH, it would be preferable to instead allow one ObjectList to reference a previously-defined PCH instead of having to define it again. This would avoid the existing bug in this PR as well as ensure changes to ObjectNode don't introduce such a bug in the future.

(I appreciate that the TODO was a little misleading, as it pretty much says to implement it as you have done. Sorry about that :( )

It would also make it much easier to get this change integrated, if (in order of importance):
 - the tests were extended for this case (important)
 - the documentation (for errors) was updated (moderately important, and possibly not needed if we make the change I mention above)
 - the code adhered to the coding standard (minor)

fran6co:
@ffulin Is this closer to what you had in mind?

- it's enough to have the same PCHOutputFile to reuse a PCH
- if PCHOptions is set, it must match the one in the node. Otherwise it can be ignored
- Added a simple test that reuses PCH
- Not sure what I'm missing code standard wise, I tried to follow the surrounding code style

fran6co:
Hi @ffulin I made a CMake generator for fastbuild but this issue is blocking it (https://gitlab.kitware.com/cmake/cmake/merge_requests/4200). Is there anything else I can do to unblock it?

ffulin:
I started to integrate this change, but realized it needed some more involved changes to handle all the cases I think it should. Rather than more back and forth which might hold up getting this resolved, I went ahead and made those changes in https://github.com/fastbuild/fastbuild/commit/4cd6eaf9061a98b676a0357ccd19cd62d8d101a8, instead of integrating this PR.

The end result should be the same as your PR, in terms of allowing re-use of the same PCH via .PCHOutputFile, but also:
 - added detection for some error-cases/misuse
 - allows one ObjectList to only define a PCH, and another one (or more) to use it (as a convenience/more flexible way of setting things up)

Please let me know if for some reason I missed something about your use-case.

These changes will be in the next release (v1.00)

fran6co:
Looks good, thanks!

PR #680, opened 2020-01-04
Use different token type for opening and closing brackets

dummyunit:
All places in the code that check for a particular bracket type also need to distinguish between opening and closing brackets.

In some places only the bracket type was checked and it resulted in code like this to be parsed without errors:
```
    .Array = } 42 }
    .Struct = ]]
```

To simplify checks and to eliminate possibility of forgetting to check for opening vs closing bracket, opening an closing bracket now use different `BFFTokenType` values.


ffulin:
I've made some changes to the parsing implementation which address the issues with open vs close braces: https://github.com/fastbuild/fastbuild/commit/c7952fcba4c1ba55fec6dcde4eb8e8ccbef792a3

PR #682, opened 2020-01-08
Fixes MSVC compilation with 14.24.28314

fran6co:
Getting errors when compiling:
- `warning C4626: 'BFFToken': assignment operator was implicitly defined as deleted`
- `warning C5027: 'BFFToken': move assignment operator was implicitly defined as deleted`
- `warning C4800: Implicit conversion from 'const char *' to bool. Possible information loss`


ffulin:
Thanks :)

Issue #228, opened 2017-03-29
FASTBuild will always rebuild file with #line directives


Leandros:
If a file contains `#line` directives, FASTBuild will always rebuild the file. 
The problem is likely related to clang, since it appears to only happen with clang on macOS/Linux. Windows + MSVC doesn't have the issue.

From the `-verbose` logs, it seems like FASTBuild considers the files mentioned in `#line 1 "foo.c"` directives to be dependencies.
You can reproduce the problem by using this file inside a FASTBuild project:

```
#line 1 "missing.c"

int foo = 5;
```

FASTBuild output with verbose will reveal it:

```
-Build: 0 ms    /Users/arvidgerstmann/workspace/private/cc/src/foo.c
-Build: 0 ms    /Users/arvidgerstmann/workspace/private/cc/missing.c
Need to build '/Users/arvidgerstmann/workspace/private/cc/_build/src/foo.o' (dep missing: '/Users/arvidgerstmann/workspace/private/cc/missing.c')
```

`missing.c` doesn't exist, though. It's only mentioned in `#line` (and, in fact, fastbuild will not try to compile it as well).

ffulin:
FASTBuild uses the #line directives seen in the pre-processed output to discover dependency information. If you inject your own #line directives, it will see those as dependencies.

Can you modify whatever process you have that creates these #line directives to have properly formed paths (i.e. that match the behaviour of the compiler)?

Leandros:
The `#line` directives are generated internally by FASTBuild if I use it with `clang`. I can't change that.

ffulin:
FASTBuilds parsing of #line directives either expects paths relative to the working directory, or full paths (the two types I've seen from the supported compilers). I think it would be tricky to differentiate file-relative vs working-dir relative to support the example you've provided without a significant performance hit (we'd have to check files on disk for existence).

Is it possible to alter the behaviour of re2c so it provides full paths or working-dir relative paths?

Leandros:
> Is it possible to alter the behaviour of re2c so it provides full paths or working-dir relative paths?

Yes, I've went with the simple solution of simply disabling the #line directives, which solved it.

ffulin:
Glad to hear you found a solution!

ffulin:
Far too late to address your original issue, but https://github.com/fastbuild/fastbuild/pull/685 should allow you to turn directives back on if you desire. It has been integrated and will be in the next release (v1.00)

PR #685, opened 2020-01-20
Only add include if preprocessor included it

crondog:
When using tools like bison to generate a cpp file, it will leave `#line`
directives. This can cause fbuild to incorrectly add the file as a
dependency. Instead of always adding the included file, only add it if
the preprocessor included it. This can be detected by flags at the end of the
preprocessor line. Eg

```
cat main.cpp
 #include "found.h"
 #line 1 "notreal.h

touch found.h

g++ -E -c main.cpp
 # 1 "main.cpp"
 # 1 "<built-in>"
 # 1 "<command-line>"
 # 1 "/usr/include/stdc-predef.h" 1 3 4
 # 1 "<command-line>" 2
 # 1 "main.cpp"
 # 1 "found.h" 1
 # 2 "main.cpp" 2
 # 1 "notreal.h"
```

Lines which contain at least a ` 1` at the end should be the only files
tracked by fbuild since that is what the preprocessor includes.

https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html

The reason why `GetSize()` is reduced by 1 is because the actual file
being compiled is in the preprocessed output and it now skipped (because
it does not have a ` 1` and fbuild tracks the file anyway)

Lines starting with `#line` for the clang fms-extension are still always
added

Fixes: https://github.com/fastbuild/fastbuild/issues/228

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/CIncludeParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/CIncludeParser.cpp
@@ -258,6 +258,7 @@ bool CIncludeParser::ParseGCC_Preprocessed( const char * compilerOutput,
     (void)compilerOutputSize;
 
     const char * pos = compilerOutput;
+    bool hasFlags = true;
 
     // special case for include on first line
     // (out of loop to keep loop logic simple)
@@ -283,6 +284,7 @@ bool CIncludeParser::ParseGCC_Preprocessed( const char * compilerOutput,
         }
         if ( strncmp( pos, "line ", 5 ) == 0 )
         {
+            hasFlags = false;
             pos += 5;
             goto foundInclude;
         }
@@ -339,8 +341,16 @@ bool CIncludeParser::ParseGCC_Preprocessed( const char * compilerOutput,
         {
             continue;
         }
+        pos++;
 
-        AddInclude( lineStart, lineEnd );
+        // only add an include if the preprocessor included it (indicated by the `1` flag
+        // https://gcc.gnu.org/onlinedocs/cpp/Preprocessor-Output.html
+        // or if it is coming from -fms-extention which doesn't have flags
+        if ( strncmp( pos, " 1", 2 ) == 0 || !hasFlags )
+        {
+            AddInclude( lineStart, lineEnd );
+        }
+
     }
 
     return true;
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestIncludeParser.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestIncludeParser.cpp
@@ -277,9 +277,9 @@ void TestIncludeParser::TestGCCPreprocessedOutput() const
 
             // check number of includes found to prevent future regressions
             const Array< AString > & includes = parser.GetIncludes();
-            TEST_ASSERT( includes.GetSize() == 222 );
+            TEST_ASSERT( includes.GetSize() == 221 );
             #ifdef DEBUG
-                TEST_ASSERT( parser.GetNonUniqueCount() == 1029 );
+                TEST_ASSERT( parser.GetNonUniqueCount() == 308 );
             #endif
         }
     }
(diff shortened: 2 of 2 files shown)


ffulin:
Thanks for the enhancement! This has been integrated and will be in the next release (v1.00).

I am curious - did you ever test the behavior with the MSVC compiler - I'd assume it has a similar problem, though I don't think we can differentiate the includes in that case.

crondog:
With the msvc compiler it does not have the flags at the end of the line.

May I ask what the reason for using the preprocessor is instead of /showincludes or -MD which doesn't have this problem?

missmah:
Lines starting with `#line` for the `clang` `-f-ms-extension` are still always
added

@crondog, can you explain this?  What's the limitation in this case? Does this mean `-f-ms-extension` builds will still rebuild those files needlessly? Thanks!

crondog:
@missmah That seems to be the case for the example test file. If you version of clang produces lines without the flags at the end, then the cpp file will always be built since fbuild thinks its dependencies are missing.

However I can't seem to reproduce the output for `fbuildcore.clang.ms-extensions.ii`

All my invocations of clang produce output which contains the flags at the end. This is from clang 9 on linux and windows

```
cat main.cpp
#line 1 "missing.h"

clang++ -E -fms-extensions main.cpp
# 1 "main.cpp"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 389 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "main.cpp" 2
# 1 "missing.h"

clang-cl -E -fms-extensions main.cpp
# 1 "main.cpp"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 365 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "main.cpp" 2
# 1 "missing.h"
```

With clangc2 from microsoft I get this:

```
cat main.cpp
#line 1 "missing.h"

"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\ClangC2\14.10.25903\bin\HostX64\clang.exe" -E main.cpp
# 1 "main.cpp"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 337 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "main.cpp" 2
# 0 "missing.h"

"C:\Program Files (x86)\Microsoft Visual Studio\2017\Community\VC\Tools\ClangC2\14.10.25903\bin\HostX64\clang.exe" -E -fms-extensions main.cpp
# 1 "main.cpp"
# 1 "<built-in>" 1
# 1 "<built-in>" 3
# 337 "<built-in>" 3
# 1 "<command line>" 1
# 1 "<built-in>" 2
# 1 "main.cpp" 2
# 0 "missing.h"
```

@ffulin Do you remember which version of clang you used to generate `fbuildcore.clang.ms-extensions.ii`

ffulin:
@crondog 

> With the msvc compiler it does not have the flags at the end of the line.

Of course, this is what I mean. When using the msvc compiler with /E, I'm wondering if the #line directives emitted by the preprocessor will be the same as the ones from the external source. i.e. does the same bug exist that this PR fixes for Clang/GCC exist there, but in a way we can't really fix because the directives are indistinguishable?

> May I ask what the reason for using the preprocessor is instead of /showincludes or -MD which doesn't have this problem?

FASTBuild uses the preprocessed source for the cache and for distributed compilation. Since we have to use those to power those features, we take advantage of the preprocessed output containing the dependency information. 

It neither of those is active, we also gather the dependency information by parsing the preprocessed output. If neither option is active, we use /showIncludes for MSVC as it is a bit cheaper.

crondog:
@ffulin Currently the bug exists for msvc

```
cat main.cpp
#line 1 "missing.h"

cl /E -c main.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27032.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

main.cpp
#line 1 "main.cpp"
#line 1 "missing.h"
```

An interesting thing to note however is that we might be able to detect it. For example

```
cat main.cpp
#include "found.h"
#line 1 "missing.h"

cl /E -c main.cpp
Microsoft (R) C/C++ Optimizing Compiler Version 19.16.27032.1 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.

main.cpp
#line 1 "main.cpp"
#line 1 "c:\\code\\foo\\found.h"
#line 2 "main.cpp"
#line 1 "missing.h"
```

It appears that headers which cl finds get translated to absolute paths. The code could be updated to only keep track of lines with a full path to the header

Issue #642, opened 2019-10-13
Paths to shared folder


Jex-Object101:
Hello. Some of my libraries located in shared folder on remote pc and i'm not able to relocate they. When I pass paths to this libs to link.exe (as additional library directories), fastbuild cutting '\\' in front for path. I'm also try to escape ('^\\^\\'), but it doesnt work. Is there any workaround?
Visual Studio 2017
FASTBuild - v0.98

ffulin:
I've confirmed that path canonicalization (the NodeGraph::CleanPath function) doesn't respect the leading slashes of a UNC path, which causes it to be treated as a relative path.

A work-around in the short-term is to map a drive letter to the UNC path and reference the files via the mapping. Full paths starting with drive letters appear to be respected.

ohoughton:
I'm going to add on to this -- fastbuild doesn't interpret a decent amount of windows paths correctly, including some common and some quite uncommon.  Notably, any path starting with a single leading slash, when cleaned to a full path, will be set as relative to the current directory rather than the root of the current drive.  The NodeGraph::CleanPath function also removes any leading slashes, which prevents shared folders (as in the original poster's issue).

For reference, here are some basic windows path normalization rules: https://blogs.msdn.microsoft.com/jeremykuhne/2016/04/21/path-normalization/

For the most part, these aren't the worst to implement.  Additionally, certain paths (i.e. ones starting with \\?\) should not be cleaned at all.  However, certain specifications, such as the drive relative paths, would be quite difficult to implement if we want to actually create a full path.  I'm currently looking into setting up CleanPaths to work with the simpler paths, but I'm not sure how we might want to handle the specified drive relative paths.

PR #688, opened 2020-01-22
Add improvements to windows path handling

ohoughton:
Should solve #642 and improve some other windows path edge cases.  Might not fix **all** windows path edge cases, but the major ones should be here, namely:

- \\\\?\\ style (paths that are completely unaltered)
- Shared folder and device paths (\\\\.\\ and \\\\server\\shared\\)
- Can use \\\\ or \/\/ for shared or device paths
- Drive relative paths (\\path)


Code/Tools/FBuild/FBuildCore/Graph/NodeGraph.cpp
```diff
@@ -1466,7 +1466,6 @@ bool NodeGraph::CheckDependencies( Node * nodeToBuild, const Dependencies & depe
 
     // sanity checks
     ASSERT( cleanPath.Find( OTHER_SLASH ) == nullptr ); // bad slashes removed
-    ASSERT( cleanPath.Find( NATIVE_DOUBLE_SLASH ) == nullptr ); // redundant slashes removed
```
ohoughton: For clarification: now that I am trying to support UNC paths, double slashes can (and will) appear in correctly cleaned paths.

PR #689, opened 2020-01-23
Add ability to manually specify workers via an environment variable.

eclazi:
E.g FASTBUILD_WORKERS=10.0.0.1;10.0.0.2;10.0.0.3

Useful for managing groups of workers where worker discovery cannot be used (e.g DNS lookup not working).

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/FBuild.cpp
+++ b/Code/Tools/FBuild/FBuildCore/FBuild.cpp
@@ -373,9 +373,21 @@ bool FBuild::Build( Node * nodeToBuild )
         Array< AString > workers;
         if ( settings->GetWorkerList().IsEmpty() )
         {
-            // check for workers through brokerage
-            // TODO:C This could be moved out of the main code path
-            m_WorkerBrokerage.FindWorkers( workers );
+            // Check for workers for the FASTBUILD_WORKERS environment variable
+            // which is a list of worker addresses separated by a semi-colon.
+            
+            AString workersEnv;
+            if ( Env::GetEnvVariable("FASTBUILD_WORKERS", workersEnv) )
+            {
+                workersEnv.Tokenize(workers, ';');
+            }
+            
+            if ( workers.IsEmpty() )
+            {
+                // check for workers through brokerage
+                // TODO:C This could be moved out of the main code path
+                m_WorkerBrokerage.FindWorkers(workers);
+            }
         }
         else
         {


bburgin:
I think you could do this today without your changes, by importing your env variable in your bff file and then assigning the list to the Settings object, which allows a fixed list of worker machines. Have you tried this? Perhaps I am missing something in your use case that is not working using the Settings object?

eclazi:
No I haven't tried this. I think it will work for my case as well. Thanks

eclazi:
As far as I can tell there is no way in the bff syntax to split a string separated by a specified character.

bburgin:
> 
> 
> As far as I can tell there is no way in the bff syntax to split a string separated by a specified character.

Seems like the .bff format needs some kind of toArray function. You could file an issue suggesting one.

For now, it's not the best, but I think you could define numbered env variables that you conditionally import and then append to the Settings Workers field:

#if exists(FASTBUILD_WORKER1)
    #import FASTBUILD_WORKER1
    #if exists(FASTBUILD_WORKER2)
        #import FASTBUILD_WORKER2
        #if exists(FASTBUILD_WORKER3)
            #import FASTBUILD_WORKER3
        #endif
    #endif
#endif

Settings
{
#if exists(FASTBUILD_WORKER1)
    .Workers = .FASTBUILD_WORKER1
    #if exists(FASTBUILD_WORKER2)
                   + .FASTBUILD_WORKER2
        #if exists(FASTBUILD_WORKER3)
                   + .FASTBUILD_WORKER3
        #endif
    #endif
#endif
}

eclazi:
@ffulin is this something you would consider?

eclazi:
Manage to to do this by including a another .bff file which path is defined by an environment variable.

ffulin:
It would be pretty reasonable to expose a new environment variables to explicitly list workers, semi-colon delimited, which could then be imported. This would be consistent with https://github.com/fastbuild/fastbuild/pull/585 which extends the FASTBUILD_BROKERAGE_PATH environment variable to support a semi-colon delimited list of brokerage paths.

eclazi:
@ffulin sounds good. I will leave this open for now.

ffulin:
Thanks for enhancement. This change has been integrated and will be in the next release (v1.03)

I made some small followup changes (https://github.com/fastbuild/fastbuild/commit/113b1f93dbbb191d3842ce8f2791d7bdc111a011) to tidy up the change and add documentation.

PR #693, opened 2020-02-07
Only add include if preprocessor included it MSVC

crondog:
This is a followup to https://github.com/fastbuild/fastbuild/pull/685
for MSVC.

The MSVC preprocessor resolves `#include`s to an absolute path while
leaving the `#line`s as they are.

Unfortunately it cannot detect the following scenario:

```
 #include <string>
 #line 1 "C:\\missing.h"
int main()
{
  return 0;
}
```

I don't think it will matter too much as tools like bison seem to
generate `#line`s with relative paths.


crondog:
@ffulin I think this might have slipped under the radar. I forgot the checklist and the original commit didn't work correctly.

PR #697, opened 2020-02-19
VSProjectExternal DLL loading thread safeness

lucianm:
VspteModuleWrapper: make use of lazy Meyer's singleton thread safeness of the constructor for loading and initializing the DLL only once.


ffulin:
Thanks 👍

PR #699, opened 2020-02-21
VSSolution folders may also contain so-called "solution items"

lucianm:
which do not belong to a project, this may come very handy in organizing some more general files to be editable within  Visual Studio even if they do not belong to a specific project of the solution, like in the attached screenshot taken from the solution generated in the test of this PR:

2020-02-21 13_38_08-soution_with_items - Microsoft Visual Studio

It contains projects organized into solution folders, also one which is not included in any folder. Unlike projects, solution items always belong into some folder, even creating or adding existing ones from Visual Studio's GUI will create a default "Solution items" folder.
Projects and solution items within solution folders are now optional, this allows having solution folders with either of them, or both, or even empty, thought this would mimic better what is allowed in Visual Studio solutions.

It would be great if this could also make it into the upcoming v1.0 ;-)


ffulin:
This has been integrated and will be in the next release (v1.00). Thanks :)

I made one small change to make the solution folder comparison case-insensitive (https://github.com/fastbuild/fastbuild/commit/146a6e74588f69e46f8ef596093bc27570168a10)

PR #700, opened 2020-02-21
Update bash completion

dummyunit:
# Description:
* Add 7 new options.
* Combine several `case ${prev}` blocks.
* Consistently use `${}` for variable expansion.

# Checklist:
The PR:
- [x] **Is self-contained**
- [x] **Compiles on all platforms** - Not applicable because PR doesn't touch C++ code.
- [ ] **Has accompanying tests** - No tests for bash completion currently, do we want them?
- [x] **Passes tests**
- [ ] **Follows the code style** - Is there is a coding style for shell scripts?


ffulin:
Thanks for the update! This will be in the next release (v1.00)

**> No tests for bash completion currently, do we want them?**

I think this is something that's fine to have no tests.

**> Is there is a coding style for shell scripts?**

Nope. I think we're good on this front.

PR #701, opened 2020-02-23
Improve include path extraction in ProjectGeneratorBase

dummyunit:
# Description:
* Add `/imsvc` option supported by clang-cl.
* Add `-idirafter` option supported by GCC and Clang.
* Extract include paths according to the priority of options that introduce them. Eg. path added via `-I` will be searched before the one that is added via `-isystem` even if it appears later on the command line.
* Fix extraction of the option argument from `-isystem-afterDIR` which didn't work correctly due to a missing `break`. Previously `-afterDIR` was extracted in that case as the include path.
* Add test for the order of include paths in generated project files.
* Add all recognized options for specifying include paths to the documentation.

# Checklist:

The PR:
- [x] **Is self-contained** - Technically there are 2 bug fixes (`-isystem-after` and proper path order) and 1 feature (new recognized options), but they are all related to the same piece of functionality so I am going to tick the box.
- [x] **Passes tests**


ffulin:
This change has been integrated and will be in the next release (v1.03). Thanks!

PR #702, opened 2020-02-24
Improve extraction of libraries from command line for GCC

dummyunit:
# Description:
* Support `-l:filename` syntax which allows to link a specific file.
* Support `-Bstatic` and `-Bdynamic` options which control whether `-lfoo` checks for existence of a corresponding dynamic library file (`libfoo.so`) or not.
* Change interpretation of `-l` options to match behavior of GNU ld - `-Bdynamic` is the default for systems that support dynamic linking, so for `-l` options we check for dynamic libraries first.

# Checklist:

The PR:
- [x] **Is self-contained** - Technically there are 2 features but they are related to the same piece of functionality so I am going to tick the box.
- [x] **Passes tests**
  This is a hard one. I am going to say that this PR keeps Windows, OSX and Linux at the _lack of parity_.
  On one hand it completely ignores MacOS specifics: dynamic libraries are hardcoded to have `.so` extension and I don't know if Apple's linker actually supports `-l:file`, `-Bdynamic` or `-Bstatic`.
  On the other hand library dependency extraction wasn't at parity prior to this change - Windows (MVSC) had a much better implementation. And this PR at least improves implementation for one platform that was behind (if we assume that GCC == Linux).
- [ ] **Includes documentation** - Extraction of library dependencies isn't covered in the documentation. This PR doesn't change that.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp
@@ -989,10 +989,13 @@ void LinkerNode::GetImportLibName( const AString & args, AString & importLibName
     args.Tokenize( tokens );
 
     bool ignoreAllDefaultLibs = false;
+    bool isBstatic = false; // true while -Bstatic option is active
     Array< AString > defaultLibsToIgnore( 8, true );
     Array< AString > defaultLibs( 16, true );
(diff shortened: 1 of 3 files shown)


ffulin:
Thanks for the improvements and fixes! This has been integrated and will be in the next release (v1.03)

PR #703, opened 2020-02-24
Don't define nullptr

dummyunit:
# Description:
All supported compilers support `nullptr` natively.
Avoiding custom definition helps to find some unintended type conversions.

# Checklist:

The PR:
- [x] **Is created against the Dev branch**
- [x] **Is self-contained**
- [x] **Compiles on all platforms**
- [ ] **Has accompanying tests** - This if a refactoring/cleanup, so no new tests.
- [x] **Passes tests**
- [x] **Keeps Windows, OSX and Linux at parity**
- [x] **Follows the code style**
- [ ] **Includes documentation** - This if a refactoring/cleanup, so no documentation required.

PR #704, opened 2020-02-24
Use Array::EmplaceBack instead of Append where possible

dummyunit:
# Description:

This commit replaces usage of `Append` with `EmplaceBack` is 3 cases:
1. When `Append` is called on a temporary - patterns like `Append(T(...))` are replaced with `EmplaceBack(...)`.
   This saves one move construction for the stored type.
2. When a local variable is created and immediately passed to `Append`.
   This saves one copy construction for the stored type.
3. When the result of `AString::GetEmpty()` is used to put an empty string into `Array`. Same result can be archived by using `EmplaceBack()` which calls the default constructor of `AString`.
    
Unit tests for the Array itself are left intact to test `Append` and `EmplaceBack` separately.

This PR depends on #703 because GCC complains in several places about invalid conversion from `int` to pointer if the current definition of `nullptr` is used.

# Checklist:

The PR:
- [x] **Is self-contained** - ~~Required changes to `nullptr` are extracted into separate PR: #703, when it will be merged this PR will become self-contained.~~ Done
- [ ] **Has accompanying tests** - This if a refactoring, so no new tests.
- [x] **Passes tests**
- [ ] **Includes documentation** - This if a refactoring, so no documentation required.

### DIFF
--- a/Code/Core/CoreTest/Tests/TestHash.cpp
+++ b/Code/Core/CoreTest/Tests/TestHash.cpp
@@ -140,11 +140,11 @@ void TestHash::CompareHashTimes_Small() const
 {
     // some different strings to hash
     Array< AString > strings( 32, true );
-    strings.Append( AString( " " ) );
-    strings.Append( AString( "shOrt" ) );
-    strings.Append( AString( "MediumstringMediumstring123456789" ) );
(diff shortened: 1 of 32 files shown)


ffulin:
Thanks for the cleanup :)

PR #706, opened 2020-02-27
Adds support for manually setting the librarian type just like the linker type

fran6co:
# Description:

Adds a way to change the librarian type just like the linker type. This is useful when the librarian executable has a different name than the one fastbuild expects and/or has a launcher (CMake for example adds a launcher to this).

This is needed to land https://gitlab.kitware.com/cmake/cmake/merge_requests/4200

# Checklist:

Please consider this checklist to assist with integration of your changes:

The PR:
- [x] **Is created against the Dev branch** - PRs are only accepted against the development branch.
- [x] **Is self-contained** - A well-targetted, minimal change that doesn't mix multiple pieces of functionality. Unnecessary whitespace or formatting changes are avoided.
- [x] **Compiles on all platforms** - The PR compiles without error on Windows, OSX and Linux.
- [x] **Has accompanying tests** -  Tests for new functionality or regression tests for bug fixes are provided.
- [x] **Passes tests** - Existing and new tests pass on all platforms.
- [x] **Keeps Windows, OSX and Linux at parity** - The change makes sense for all platforms if not inherently platform-specific.
- [x] **Follows the code style** - The existing coding style and conventions are adhered to.
- [x] **Includes documentation** - Documentation is updated or added if appropriate.


bburgin:
Why not just set the .Librarian field?
It is the full path to the librarian executable.
Example:
.Librarian = '$MSVCPath$/win64/x64/bin/amd64_x86/lib.exe'

fran6co:
> Why not just set the .Librarian field?
> It is the full path to the librarian executable.
> Example:
> .Librarian = '$MSVCPath$/win64/x64/bin/amd64_x86/lib.exe'

@bburgin I am setting the .Librarian field but it's `.Librarian = 'C:\Program Files\CMake\bin\cmake.exe'` and the actual lib.exe is in the arguments. CMake adds some environmental variables to ensure that the command will run correctly

ffulin:
This change seems pretty reasonable in general, but I'd like to understand the CMake interaction a little better.

Is this to allow link.exe to be used as the librarian with /lib? Or is it for some other use-case?

I noticed there was some talk of "response files". Please note that you'll need to make sure CMake is not using response files and passing those to FASTBuild. You'll need to pass the full args to FASTBuild in the generated bff and let FASTBuild manage creation of response files. If you don't dependency information will be missing.

ffulin:
I'm also confused as to why you would want to invoke cmake.exe from FASTBuild.

fran6co:
@ffulin there are several things going on:

- Invoking link.exe with /lib doesn't work when using response files if the /lib flag ends up in the response file, so I had to make CMake use lib.exe instead. I tried to fix this in Fastbuild but I couldn't come up with a clean solution. I am passing the whole arguments to Fastbuild in the CMake generator
- The other problem is that CMake sometimes adds a launcher to commands to set some environmental variables. For example a call to link.exe would become `cmake.exe -E vs_link_exe -- link.exe`.  This is handled correctly in the LinkerNode but not in the LibraryNode. So I need to tell Fastbuild that its actually link.exe even when It's setting the exe to cmake.exe

fran6co:
Just to clarify I fixed the link.exe /lib with response files issue by side stepping it and making CMake use lib.exe instead

ffulin:
This PR can be integrated, but it currently:
 - does not adhere to the coding style (various whitespace and brace inconsistencies)
 - does not have any tests (although it modifies existing functionality in a fairly simple way, and that functionality did not have existing tests, we need to watch out for regressions)

I can fix both of those things during integration, but if you can address them (either or both) first, that will accelerate things.

fran6co:
I'll fix those, is there any clang format file or some style documentation? I find it really hard to maintain the style by hand

ffulin:
Thanks :)

I evaluated .clang-format some time ago and found it lacking the necessary features. I am taking a look at it again (literally right now :) ) and there have been some improvements since I last tried. I'm not certain it'll be possible to use yet though.

ffulin:
Unfortunately, it looks like clang-format is still too limited (can't differentiate certain cases) and too heavy handed (all or nothing) to be used. If works great 95% of the time, but in 5% of cases reformats the code in ways that significantly decrease legibility.

fran6co:
That's too bad, having clang-format would be nice. Just added the tests and the format fix (hope I got all of them)

ffulin:
Thanks for adding the tests. (Extra thanks for the additional test for LinkerType and the bug fix for the GreenHills linker detection!).

This has been integrated and will be in the next release (v1.00).

PR #716, opened 2020-03-11
Fixes lightcache randomly generating different cache keys

fran6co:
# Description:

Includes https://github.com/fastbuild/fastbuild/pull/624

Randomly lightcache would generate different cache keys for the same obj file. This is caused by the caching of the files in `g_AllIncludedFiles`, multi threading and not saving the defines in the file object itself. When a file was loaded from `g_AllIncludedFiles` it would not include the defines and depending what thread was running the file would be missing defines

# Checklist:

Please consider this checklist to assist with integration of your changes:

The PR:
- [x] **Is created against the Dev branch** - PRs are only accepted against the development branch.
- [x] **Is self-contained** - A well-targetted, minimal change that doesn't mix multiple pieces of functionality. Unnecessary whitespace or formatting changes are avoided.
- [x] **Compiles on all platforms** - The PR compiles without error on Windows, OSX and Linux.
- [ ] **Has accompanying tests** -  Tests for new functionality or regression tests for bug fixes are provided.
- [x] **Passes tests** - Existing and new tests pass on all platforms.
- [x] **Keeps Windows, OSX and Linux at parity** - The change makes sense for all platforms if not inherently platform-specific.
- [x] **Follows the code style** - The existing coding style and conventions are adhered to.
- [ ] **Includes documentation** - Documentation is updated or added if appropriate.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/LightCache.h
+++ b/Code/Tools/FBuild/FBuildCore/Cache/LightCache.h
@@ -24,7 +24,6 @@ class LightCache
 {
 public:
     LightCache();
-    ~LightCache();
 
     bool Hash( ObjectNode * node,                // Object to be compiled
                const AString & compilerArgs,     // Args to extract include paths from
--- a/Code/Tools/FBuild/FBuildCore/Cache/LightCache.h
+++ b/Code/Tools/FBuild/FBuildCore/Cache/LightCache.h
@@ -24,6 +24,7 @@ class LightCache
 {
 public:
     LightCache();
+    ~LightCache() = default; // Destruction heavy-lifting is currently done by IncludedFile::~IncludedFile()
 
     bool Hash( ObjectNode * node,                // Object to be compiled
                const AString & compilerArgs,     // Args to extract include paths from
--- a/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
@@ -46,11 +46,14 @@ public:
         IncludeType                 m_Type;
     };
 
+    ~IncludedFile();
+
     uint64_t                        m_FileNameHash;
     AString                         m_FileName;
     bool                            m_Exists;
     uint64_t                        m_ContentHash;
     Array< Include >                m_Includes;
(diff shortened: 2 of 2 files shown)


Code/Tools/FBuild/FBuildCore/Cache/LightCache.h
```diff
@@ -24,7 +24,6 @@ class LightCache
 {
 public:
     LightCache();
-    ~LightCache();
```
missmah, line 27: `~LightCache() = default; // Destruction heavy-lifting is currently done by IncludedFile::~IncludedFile()` ?

Not sure how strongly I feel, since source control should have the history of where this went. And it's not my call anyway -- but I do tend to prefer explicit "there's not supposed to be a destructor right now, because X" kind of message to the implicit which could always be misinterpreted by the reader.

Again, the implicit is that hey, none of these pointers are currently owned by the `LightCache`, so it doesn't need a destructor. But C++ is complex enough that someone could still read the code and be left unsure.

Anyway, this is an op-ed comment that you can safely ignore, since I'm not an owner of this project =)

fran6co: Probably this needs a bit of refactoring to do it properly, the ownership is not very clear right now. Maybe using smart pointers would make it a bit more clear but this is a no stl kind of project.

Having said that, I could add the explicit destructor and the comment. This code is hard to follow as is

ffulin:
Thanks for the fixes. I've integrated these changes and added tests (https://github.com/fastbuild/fastbuild/commit/da08cd2e0f4d6a08668347724cb1e635815bbc84). These fixed will be in the next release (v1.00)

To summarize, the two issues this PR resolves are:

a) If an IncludeFile from the in-memory cache was re-used, it would be lacking macro defines used in includes. This resulted in missing dependency information (which could cause rebuilds to not trigger when needed). The missing dependency information also resulted in different cache keys as you noted. This issue was not related to multi-threading. It could be reproduced in single threaded scenarios.

b) If macro defines were encountered which iterating a macro-based include, the list of includes could be resized and/or move. This could result in crashes due to invalid memory accesses.

Additionally, as part of my followup commit, I fixed an issue where failure to correctly process a macro-based include was silently ignored.

PR #730, opened 2020-04-13
Fix -cacheinfo and -cachetrim on case-sensitive filesystems

dummyunit:
# Description:
`ICache::GetCacheId` generates names for cache file using uppercase hex letters A-F. This also affects names of sub-directories in which these files are stored.
`Cache::GetCacheFiles` on the other hand generated names for these sub-directories using lowercase hex letters. As the result that function looked for cache files in different directories.

This worked fine on Windows (and on OSX?) due name matching being case-insensitive. But caused a severe undercounting of the cache directory size (at least by a factor of 6.5) on Linux, unless case-insensitive filesystem was being used for cache (eg. mounted VirtualBox shared folder).

# Checklist:

The pull request:
- [x] **Is self-contained**
- [ ] **Has accompanying tests** - There are no tests for `-cachetrim` currently. Should I add some? Or should I add tests for `Cache::GetCacheFiles`? (this would be easier)

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/Cache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/Cache.cpp
@@ -326,7 +326,7 @@ void Cache::GetCacheFiles( bool showProgress,
         {
             // Get Files
             AStackString<> path;
-            path.Format( "%s%02x%c%02x%c", m_CachePath.Get(),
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the fix. This has been integrated and will be in the next release (v1.00)

I think it might not worth worrying about a test for the moment, but I'd definitely be open to one if you think it can be done in a clean way.

Issue #732, opened 2020-04-16
A case for extending #if


ggnkua:
Hello,

I'm in the process of porting a rather large project over to fastbuild, and like the subject says, I've been taking hits when it comes to using #if with something more complex than one variable. However, my cases are not super crazy, mostly along the lines of:

`#if aaa & bbb`
`   ...`
`#endif`

or

`#if aaa | bbb`
`    ...`
`#endif`

The first one I can work around with a nested #if, no biggie. But the second really hurts as I have to copy the insides of the #if twice and have two separate #if one below the other. This gets messy rather quick.

Now, I took the liberty and writing some code that seems to work fine with the above cases, or any arbitrary combination of logical AND, OR, NOT. I can make an effort to produce a pull request for you but I'd like to know your thoughts on this first in order to not waste much time on this. By sheer luck I came across [a previous post of you discussing the matter ](https://github.com/fastbuild/fastbuild/issues/715#issuecomment-602077915), and I agree that going overboard with extending #if will not yield anything practical, but is for more academic use.

Just for reference, here is the test .bff that seems to work ok with my code:

```
#define aaa
#define bbb
#define ccc
#if aaa & bbb
    Print('AND test #1 ok')
#endif
#if aaa & bbb & ccc
    Print('AND test #2 ok')
#endif

#if ccc | ddd
    Print('OR test #1 ok')
#endif
#if ddd | ccc
    Print('OR test #2 ok')
#endif
#if ccc | ddd | aaa | bbb
    Print('OR test #3 ok')
#endif

#if aaa & !ddd
    Print('AND ! test #1 ok')
#endif
#if !ddd & !aaa
    Print('AND ! test #2 ok')
#endif
    
#if !eee | !ddd
    Print('OR ! test #1 ok')
#endif
#if aaa | !ddd
    Print('OR ! test #2 ok')
#endif
#if !ddd | aaa
    Print('OR ! test #3 ok')
#endif

;Should produce syntax error
#if | ddd
    Print('| at the start - should not reach this!')
#endif
#if & ddd
    Print('& at the start - should not reach this!')
#endif
```

Looking forward to your reply!

ffulin:
I think it'd be nice to extend #if to support OR and AND and I'd be open to a PR to that end. 

A couple of thoughts that spring to mind about this:
 - I think we'd want to use && and || as the operators like C/C++ (as opposed to & and |)
[...]

PR #733, opened 2020-04-17
Extending #if parsing with boolean && and ||

ggnkua:
# Description:

As proposed in issue #732 here are the relevant changes to enable || and ## in #if parsing

# Checklist:

The pull request:
- [x] **Is self-contained**
(only tested on Windows)
(hopefully!)


Code/Tools/FBuild/FBuildCore/BFF/Tokenizer/BFFTokenizer.cpp
```diff
@@ -38,7 +38,7 @@ namespace
     bool IsStringStart( const char c )      { return ( ( c == '\'' ) || ( c == '"') ); }
     bool IsVariableStart( const char c )    { return ( ( c == '.' ) || ( c == '^' ) ); }
     bool IsVariable( const char c )         { return IsUppercaseLetter( c ) || IsLowercaseLetter( c ) || IsNumber( c ) || IsUnderscore( c );  }
-    bool IsOperator( const char c )         { return ( ( c == '+' ) || ( c == '-' ) || ( c == '=' ) || ( c == '!' ) || ( c == '<' ) || ( c == '>' ) ); }
+    bool IsOperator( const char c )         { return ( ( c == '+' ) || ( c == '-' ) || ( c == '=' ) || ( c == '!' ) || ( c == '<' ) || ( c == '>' ) || (c == '&') || (c == '|')); }
```
ffulin, line 41: This seems to extend and use a single character operator instead of a two-character operator. This doesn't seem to match the documentation or the test.

Two-character tokens are tokenized in just below the comment:
```C++
// Two character operator?
```

Code/Tools/FBuild/FBuildTest/Data/TestBFFParsing/if_directive.bff
```diff
@@ -50,6 +50,53 @@ Print( "Platform = $Platform$" )
     #endif
 #endif
 
+// Boolean algebra
```
ffulin, line 53: Rather than extending this existing test file, it'd be preferable to add a new individual test cases to TestBFFParsing.cpp. 
You can use the helper macros to check that individual snippets parse correctly, or that they fail with an expected error.

For example (these won't be exactly right but are just to give the general idea):

```C++
void TestBFFParsing::IfBooleanOperators() const
{
    // Failure cases
    TEST_PARSE_FAIL( "#if ||",          "<expected error>" );
    TEST_PARSE_FAIL( "#if &&",          "<expected error>" );
    TEST_PARSE_FAIL( "#if X && || Y\n"
                     "#endif",          "<expected error>" );
    TEST_PARSE_FAIL( "#if X && Y Z\n",
                     "#endif",          "<expected error>" );

    // OR
    TEST_PARSE_OK( "#define A\n"
                   "#if A || B\n"
                   "\n" // Should be here
                   "#else\n"
                   "    #error Should not be here\n"
                   "#endif" );
    TEST_PARSE_OK( "#if A || B\n"
                   "    #error\n" // Should not be here
                   "#endif" );
   <other cases, negation etc>

    // AND
    <etc>

    // Check precedence
    TEST_PARSE_OK( "#define A\n"
                   "#define B\n"
                   "#if A || B && C\n"
                   "    // Should be here\n"
                   "#else\n"
                   "    #error Should not be here\n"
                   "#endif" );
```

Code/Tools/FBuild/FBuildCore/BFF/Tokenizer/BFFTokenizer.cpp
```diff
@@ -572,52 +572,108 @@ bool BFFTokenizer::HandleDirective_If( const BFFFile & file, const char * & pos,
...
     argsIter++;
 
-    // Check for negation operator
-    bool negate = false;
-    if ( argsIter->IsOperator( "!" ) )
-    {
-        negate = true;
-        argsIter++; // consume negation operator
-    }
+    enum { IF_NONE = 0, IF_AND = 1, IF_OR = 2, IF_NEGATE = 4 };
```
ffulin, line 575: You might like to consider using an enum class here:
```C++
enum OperatorType : class uin32_t { IF_NONE .... }
```
This prevents accidental misuse.

Code/Tools/FBuild/FBuildCore/BFF/Tokenizer/BFFTokenizer.cpp
ffulin: This seems to work because you actually inject two tokens and consume both, but that's not right and will likely cause problems.  There should be a single, two-character operator in the stream.

ggnkua: I see a check for 2 character operators at line [237](https://github.com/fastbuild/fastbuild/blob/092d66f371fb0818b895130578850c55a570249d/Code/Tools/FBuild/FBuildCore/BFF/Tokenizer/BFFTokenizer.cpp#L237) on this file. Is that the correct place to check for 2 character operators?

ffulin: That's right.

ffulin:
I was making some unrelated parser improvements, and as part of that I've extended the OK/FAIL macros I mentioned in a way that might be useful for tests for this feature. Specifically, the OK macro now takes an optional parameter to ensure something specific exists in the output. (https://travis-ci.com/github/fastbuild/fastbuild/builds/160919483)

That means you can now do this:
```C++
    TEST_PARSE_OK( "#define A\n"
                   "#define B\n"
                   "#if A && B\n"
                   "    Print( 'Success' )\n"
                   "#endif\n",                                "Success" );
```

Code/Tools/FBuild/FBuildCore/BFF/Tokenizer/BFFTokenizer.cpp
ggnkua: Hi,

I tried using enum class as you suggested but I stumbled in a few problems. Most noticeable is that I now cannot use the enum members in boolean arithmetic. For example, I do something like

```
ifOperator |= IF_NEGATE;
```

Adding the enum class as you suggest I either have to declare `ifOperator` as an `OperatorType` type or I have to use casts everywhere: `(uint32_t)OperatorType::IF_NEGATE`. And even that doesn't solve the boolean problem stated above. Reading some questions online I saw that you can overcome this by writing specific functions that handle this.

My knowledge of C++ is sketchy at best, so I might be missing something here. But this seems unnecessarily overkill and makes the code more unreadable than it is originally.

Again, any help is appreciated because I'm stuck with this for now and will revert back to normal enums for the reworking of the code (operator precedence).

ffulin: I think if you're using it as a bit field, it's reasonable to go back to what you had. Apologies for the confusion.

ffulin:
I've integrated this change and it will be available in the next release (v1.02). Thanks for your contribution!

I have made a followup commit (https://github.com/fastbuild/fastbuild/commit/9295d002efa426372cfe06c3d49733cde33c4ecd) which:
 - simplifies some logic
 - extends the tests
 - updates documentation to note operator precedence
 - adds documentation for new eeror
 - tidies up the code style

ggnkua:
Excellent, glad I could help!

PR #760, opened 2020-06-26
Target dependencies export with a new ListDependencies() function 

poppolopoppo:
# Description:

Hello,

This pull request is an initial proposition for adding a new function `DependencyList`, which can export all the dependencies discovered for a target.

The purpose is to provide interop with other build systems, UnrealBuildTool in our case (UBT).
Unreal made recent changes to list dependencies for each translation unit, to handle internally cache expiration.
UBT is running additional work to retrieve those informations, and use a wrapper process to parse the output of `/ShowIncludes` in the case of `cl.exe`. The later is particularly problematic for compilation distribution, and FASTBuild is already handling all of this hard work internally :/

Thanks to this new function we can pipe all those dependencies directly from FASTBuild, while keeping UBT innards mainly untouched (we simply have to bypass internal depencency list generation).

The `DependencyList` function has few parameters:
````
DependencyList('DepList_Action_0')
{
    .Source = 'Action_0'                        ; Source target (filename or node name)
    .SourcePattern = { '*.h', '*.c', '*.cpp' }  ; Pattern (or array of patterns) to filter exported dependencies
    .Dest = 'DepList_Action_0.txt'              ; Destination filename where the list of files will be exported
    .PreBuildDependencies = {}                  ; Like many other nodes, but Source should be enough already
}
````

This function will only list nodes returning true when calling `node->IsAFile()` (and we assume that their name is a filename).
Output format is straight-forward, as the function will print an absolute path for each dependency, one file per line:
````
C:\Source\Public\String.h
C:\Source\Private\String.cpp
C:\Source\Private\Main.cpp
[...]
````

Implementation in the associated `DepencyListNode` is also simple, since it simply needs to recursively visit `Source` node dependencies to get all file nodes. The source node is added as a static dependency to ensure correct order of execution (we want the dynamic dependencies to be filled).

Could you please give some feedback on this new feature ?
Thanks!

_Getting this integrated inside the main-line would allow to release our integration for UE4.25 without having to support a FASTBuild fork ;)_

# Checklist:

The pull request:
- [X] **Is self-contained**

### CODE REVIEW

File: Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp [hunk 1]
```diff
@@ -0,0 +1,243 @@
...
+                    dependencyList.Append( &node->GetName() );
+                }
+            }
+        }
+
+        // Sort collected filenames
+
+        dependencyList.Sort([]( const AString * a, const AString * b ) {
+            return ( a->Compare( *b ) < 0 );
+        });
```
ffulin, line 130: As a rule, FASTBuild code prefers "normal" functions over lambdas. This is for both style and practical reasons (as of 2019 the MSVC compiler has several off-by-one line number debug info bugs related to lambdas (as of VS2019 v14.25.28610))

Array implements SortDeref which avoid the need for this lambda, so that's easily resolved here.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+        FileStream dmp;
+        if ( dmp.Open( filename, FileStream::WRITE_ONLY ) == false )
+        {
+            return false;
+        }
+        else
+        {
+            dmp.WriteBuffer( buffer.Get(), buffer.GetLength() * sizeof(*buffer.Get()) );
+            dmp.Close();
```
ffulin, line 152: We should check if the write succeeded here by comparing the return result to the size.
Though ok, you don't need to multiply the GetLength() by the size of a char.
The close can also be omitted (FileStream is self-closing). You only need to call Close() if you explicitly control the point at which the file is closed.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+        m_StaticDependencies.Append(Dependency( srcNode ));
+    }
+
+    return true;
+}
+
+// DoDynamicDependencies
+//------------------------------------------------------------------------------
+/*virtual*/ bool DependencyListNode::DoDynamicDependencies( NodeGraph & nodeGraph, bool forceClean )
+{
```
ffulin, line 202: This function doesn't appear to be needed.

When something is a static dependency, there is no need to add it as a dynamic dependency as well.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        return false; // InitializePreBuildDependencies will have emitted an error
+    }
+
+    // Get node for Source of dependency list
+    Node* srcNode = nodeGraph.FindNode( m_Source );
+    if ( srcNode != nullptr )
+    {
+        m_StaticDependencies.Append(Dependency( srcNode ));
+    }
```
ffulin, line 194: Failure to retrieve the node should be an error - since it is a misconfiguration.

This is the preferred mechanism for retrieving a dependency, which handles error output for you:
```C++
    Dependencies compilerExeFile( 1, false );
    if ( !Function::GetFileNode( nodeGraph, iter, function, m_Source, ".Source", m_StaticDependencies ) )
    {
        return false; // GetFileNode will have emitted an error
    }
```

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include "Core/FileIO/FileIO.h"
+#include "Core/FileIO/FileStream.h"
+#include "Core/FileIO/PathUtils.h"
+#include "Core/Strings/AStackString.h"
+
+// REFLECTION
+//------------------------------------------------------------------------------
+REFLECT_NODE_BEGIN( DependencyListNode, Node, MetaNone() )
+    REFLECT(        m_Source,                   "Source",                   MetaFile() )
+    REFLECT(        m_Dest,                     "Dest",                     MetaPath() )
```
ffulin, line 25: Seems like the metadata on Dest is incorrect, since this outputs a file.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include "Core/Containers/Array.h"
+#include "Core/Env/ErrorFormat.h"
+#include "Core/FileIO/FileIO.h"
+#include "Core/FileIO/FileStream.h"
+#include "Core/FileIO/PathUtils.h"
+#include "Core/Strings/AStackString.h"
+
+// REFLECTION
+//------------------------------------------------------------------------------
+REFLECT_NODE_BEGIN( DependencyListNode, Node, MetaNone() )
```
ffulin, line 23: The nodes should have MetaName applied:
```C++
MetaName( "Dest" )
```
See TextFileNode for a real use-case.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+    REFLECT_ARRAY(  m_Patterns,                 "Patterns",                 MetaOptional() )
+    REFLECT_ARRAY(  m_PreBuildDependencyNames,  "PreBuildDependencies",     MetaOptional() + MetaFile() + MetaAllowNonFile() )
+REFLECT_END( DependencyListNode )
+
+// Anonymous DependencyListBuilder struct
+//------------------------------------------------------------------------------
+struct DependencyListBuilder {
+    enum {
+        iBuckets = 256,
+        mBuckets = (iBuckets - 1)
```
ffulin, line 35: This enum is rather confusing, particularly with one of the enumerations using member variable naming style.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+            return true;
+        }
+        else
+        {
+            return false;
+        }
+    }
+
+    void Collect( const Node * node )
+    {
```
ffulin, line 69: This function is recursive, but that doesn't seem correct.

If you compile a cpp file, and that depends on a header, that header may be generated. It's pretty common for code-gen to output many headers, so while the upstream dependency of the header might change, the header itself might not. By making the dependencies recursive, you will cause unnecessary rebuilds.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+        if ( FindOrAddNode( node ) )
+        {
+            if ( node->GetType() == Node::ALIAS_NODE )
```
ffulin, line 72: Aliases are flattened at build time when using the correct MetaData markup on the node, so this check shouldn't be needed.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+DependencyListNode::~DependencyListNode() = default;
+
+// DoBuild
+//------------------------------------------------------------------------------
+/*virtual*/ Node::BuildResult DependencyListNode::DoBuild( Job * /*job*/ )
+{
+    FLOG_VERBOSE( "DependencyList: '%s' -> '%s'", m_Source.Get(), GetName().Get() );
+
+    // create the dependency list file
+    if ( DependencyListBuilder::MakeListTxt( this, m_Name.Get(), m_Patterns ) == false )
```
ffulin, line 232: Since you only have one source, and with the right metadata no alias traversal is needed, I don't think you need this helper class at all. You should be able to query the node for it's dependencies directly (they should already be unique)

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,243 @@
...
+    {
+        FLOG_ERROR( "DependencyList failed. Error: %s Target: '%s'", LAST_ERROR_STR, GetName().Get() );
+        return NODE_RESULT_FAILED; // create file failed
+    }
+
+    m_Stamp = FileIO::GetFileLastWriteTime( m_Name );
```
ffulin, line 238: you should use RecordStampFromBuiltFile here as it has some work arounds file timetime precision problems on some platforms/filesystems

Code/Tools/FBuild/FBuildCore/BFF/Functions/FunctionDependencyList.cpp
```diff
@@ -0,0 +1,104 @@
...
+//------------------------------------------------------------------------------
+/*virtual*/ bool FunctionDependencyList::AcceptsHeader() const
+{
+    return true;
+}
+
+// Commit
+//------------------------------------------------------------------------------
+/*virtual*/ bool FunctionDependencyList::Commit( NodeGraph & nodeGraph, const BFFToken * funcStartIter ) const
+{
```
ffulin, line 35: You don't need any of this logic.  With the correct MetaData on the Node, it's all taken care of.
You only need this:
{code}
// CreateNode
/*virtual*/ Node * FunctionDependencyList::CreateNode() const
{
    return FNEW( DependencyListNode );
}
{code}

See FunctionTextFile/TextFileNode for an example:

Code/Tools/FBuild/FBuildCore/Graph/Node.h
```diff
@@ -79,6 +79,7 @@ class Node : public Struct
         SETTINGS_NODE       = 20,
         VSPROJEXTERNAL_NODE = 21,
         TEXT_FILE_NODE      = 22,
+        DEPENDENCY_LIST_NODE= 23,
```
ffulin, line 82: Adding a new node type requires bumping the NodeGraph version in NodeGraph.h:
```C++
    enum : uint8_t { NODE_GRAPH_CURRENT_VERSION = 151 };
```

ffulin:
I think conceptually this feature seems pretty reasonable. I like the idea of making it easier to support Unreal.

I think the implementation has some problems, which I've called out in the comments. Thankfully, I think all of them are resolved by making the implementation simpler. Let me know if anything I wrote isn't clear.

Finally, I'm not 100% sure about the name "DependencyList". The concern is that it could be confused with a way to declare a set of dependencies for something. I think we can proceed without resolving this though - I'll give it some thought.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.h
```diff
@@ -0,0 +1,38 @@
...
+//------------------------------------------------------------------------------
+class DependencyListNode : public FileNode
+{
+    REFLECT_NODE_DECLARE( DependencyListNode )
+public:
+    explicit DependencyListNode();
+    virtual bool Initialize( NodeGraph & nodeGraph, const BFFToken * funcStartIter, const Function * function ) override;
+    virtual ~DependencyListNode() override;
+
+    static inline Node::Type GetTypeS() { return Node::COPY_FILE_NODE; }
```
ffulin, line 24: This is returning the incorrect node type.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,138 @@
...
+/*virtual*/ Node::BuildResult DependencyListNode::DoBuild( Job * /*job*/ )
+{
+    FLOG_VERBOSE( "DEPLIST: '%s' -> '%s'", m_Source.Get(), GetName().Get() );
+
+    // Collect all file dependencies
+    Array< const AString * > dependencyList;
+
+    dependencyList.SetCapacity(
+        GetStaticDependencies().GetSize() +
+        GetDynamicDependencies().GetSize() );
```
ffulin, line 93: This node doesn't have any DynamicDependencies (everything is static) so you don't need to try and add stuff from that array.

You should be able to use m_StaticDependencies directly and loop through and write it without copying. You can do the pattern filtering directly in the loop.

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,138 @@
...
+
+    AddDependenciesToList( &dependencyList , m_Patterns , GetStaticDependencies() );
+    AddDependenciesToList( &dependencyList , m_Patterns , GetDynamicDependencies() );
```
ffulin, line 96: This node doesn't have any dynamic dependencies, so you don't need the second call using GetDynamicDependencies()

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,138 @@
...
+REFLECT_NODE_BEGIN( DependencyListNode, Node, MetaName("Dest") + MetaFile() )
+    REFLECT(        m_Source,                   "Source",                   MetaFile() )
+    REFLECT(        m_Dest,                     "Dest",                     MetaFile() )
+    REFLECT_ARRAY(  m_Patterns,                 "Patterns",                 MetaOptional() )
+    REFLECT_ARRAY(  m_PreBuildDependencyNames,  "PreBuildDependencies",     MetaOptional() + MetaFile() + MetaAllowNonFile() )
+REFLECT_END( DependencyListNode )
+
+// AddDependenciesToList()
+//------------------------------------------------------------------------------
+static void AddDependenciesToList( Array< const AString * > * list , const Array< AString > & patterns , const Dependencies & deps )
```
ffulin, line 31: This function doesn't seem right.

It's called twice, once with m_StaticDependencies and once with m_DynamicDependencies.

As noted lower down m_DynamicDependencies is always empty, so this call is not needed.

For the first call, with m_StaticDependencies, this will iterate the loop once and compare that name of the one thing you depend on against the patterns. Instead, we should get the dependencies of that object and iterate those.

(In the previous iteration, you were recursively iterating everything, which we don't want, but you do need to go down one level)

Code/Tools/FBuild/FBuildCore/Graph/DependencyListNode.cpp
```diff
@@ -0,0 +1,138 @@
...
+#include "Core/Env/ErrorFormat.h"
+#include "Core/FileIO/FileIO.h"
+#include "Core/FileIO/FileStream.h"
+#include "Core/FileIO/PathUtils.h"
+#include "Core/Strings/AStackString.h"
+
+// REFLECTION
+//------------------------------------------------------------------------------
+REFLECT_NODE_BEGIN( DependencyListNode, Node, MetaName("Dest") + MetaFile() )
+    REFLECT(        m_Source,                   "Source",                   MetaFile() )
```
ffulin, line 23: This will allow "Source" to point to a file (like a library) directly or via an alias, but won't allow you to point to an ObjectList or certain other types of nodes.

I think you might want MetaAllowNonFile here.

ffulin:
The second revision looks pretty close. I added a few new comments for some things that don't look quite right.

While doing this, I thought about the name some more. I was thinking ListDependencies() might be a good choice for the function (accompanied by ListDependenciesNode). What do you think?

I think you're pretty close to having it 100%. If you are able, please add some tests and documentation for this functionality as well. This will make it much easier to integrate.

poppolopoppo:
I updated the PR, now only direct dependencies will be listed.
I also added functional tests and documentation for this feature.
I renamed the function and node `ListDependencies` instead of `DependencyList`,
Tell me if you see any other issues ;)

ffulin:
The tests seem to be failing on Windows:
```
-------- TEST ASSERTION FAILED --------

C:\Users\travis\build\fastbuild\fastbuild\Code\Tools\FBuild\FBuildTest\Tests\TestListDependencies.cpp(139): Assert: files[0].EndsWithI( "HeaderA.h" )

-----^^^ TEST ASSERTION FAILED ^^^-----
```

poppolopoppo:
Yes I saw that, didn't get the time yet to see what went wrong though.

poppolopoppo:
I failed to reproduce Travis's issue on several machines, still added more explicit failure messages hoping for more details about what went wrong.

poppolopoppo:
The issue was due to a case sensitive mismatch,
ListDependencies() results are now sorted case insensitively on Windows (still case sensitive on other systems).
The tests are passing again :)

ffulin:
This new feature has been integrated and will be in the next release (v1.03).

I made some changes in (https://github.com/fastbuild/fastbuild/commit/ce5a26498da5ae5d89d3b7affcf9e52a3c179857). Most of these are just minor style fixes, but two things are behavior changes:

**Output when building**

A message will be emitted by default when building, rather than as part of -verbose. The message will be suppressed if -quiet is used. This makes this consistent with other nodes.

If this needs refinement (if it's too verbose for example), we can perhaps expose some options on the node to alter the behavior as needed. To begin with, I thought it best to first make it consistent. Sorry I didn't notice this during initial reviews - I though it best to get this integrated rather than wait for some additional back and forth (esp. since this will be easily altered after if needed)

**Output error codes on failure**

If the node fails to open or write the output file, the errors now contains error code information.

poppolopoppo:
Thank you!

Issue #764, opened 2020-07-02
Support for `/external:I` experimental option from MSVC


poppolopoppo:
Hello!

We would like to use the experimental feature `/experimental:external` from [Visual Studio 2017 version 15.6 Preview 1](https://devblogs.microsoft.com/cppblog/broken-warnings-theory/).
This option enables `/external:I`, an alternative to `/I` to specify an include directory, which is useful for handling compilation warnings in external code when used together with `/external:W[n]`.

From what I know FASTBuild is already aware of `/I` for `cl.exe`, which allows it to detect, transform or strip those arguments according to build phase; and stripping the include paths seems to be a required mechanism to enable distributed compilation of preprocessed sources for this compiler. 

But apparently this mechanism isn't available for the option `/external:I`, as FASTBuild doesn't know of this feature.
When I ran a BFF config using this flag I got the following issues for distributed jobs:
````shell
c1xx: warning C5044: An argument to command-line option /external:I points to a path 'c:\program files (x86)\windows kits\netfxsdk\4.6.2\include\um' that does not exist
c1xx: warning C5044: An argument to command-line option /external:I points to a path 'c:\program files (x86)\windows kits\10\include\10.0.18362.0\ucrt' that does not exist
c1xx: warning C5044: An argument to command-line option /external:I points to a path 'c:\program files (x86)\windows kits\10\include\10.0.18362.0\shared' that does not exist
c1xx: warning C5044: An argument to command-line option /external:I points to a path 'c:\program files (x86)\windows kits\10\include\10.0.18362.0\um' that does not exist
c1xx: warning C5044: An argument to command-line option /external:I points to a path 'c:\program files (x86)\windows kits\10\include\10.0.18362.0\winrt' that does not exist
````
I can confirm that disabling the distribution fixed those.

I think similar issues might happen with `/imsvc`/`-imsvc` from `clang-cl.exe` [LLVM driver](https://llvm.org/builds/), which we also use by forcing the compiler family to MSVC (which is sad since explicit `clang-cl` support could be a [win](https://fastbuild.org/docs/features/caching.html), but this is another topic ;) ).

Do you think FASTBuild should support those new flags ?

Thanks!

ffulin:
FASTBuild should support these flags. Updating the main compiler invocation to remove the new option when needed should be trivial.

A little more work will be required to correctly process these paths in the LightCache, but that should still be quite manageable.

ffulin:
[...]

PR #765, opened 2020-07-04
'/imsvc' and '/external:I' flags support for MSVC

poppolopoppo:
# Description:

This PR tries to address the issues described in #764, which yield to distribution not working when those flags are used.
The patch is small, but it seems to be all that was needed ?

# Checklist:

The pull request:
- [X] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1748,10 +1748,13 @@ bool ObjectNode::BuildArgs( const Job * job, Args & fullArgs, Pass pass, bool us
                 // we expand relative includes as they would be on the host (so the remote machine's
                 // working dir is not used, which might be longer, causing this overflow an internal
                 // limit of cl.exe)
-                if ( ( job->IsLocal() == false ) && IsStartOfCompilerArg_MSVC( token, "I" ) )
+                if ( ( job->IsLocal() == false ) && (
+                        IsStartOfCompilerArg_MSVC( token, "I" ) ||
+                        IsStartOfCompilerArg_MSVC( token, "imsvc" ) ||
+                        IsStartOfCompilerArg_MSVC( token, "external:I" ) ) )
                 {
                     // Get include path part
-                    const char * start = token.Get() + 2; // Skip /I or -I
+                    const char * start = token.Get() + 2; // Skip /I or -I or /imsvc or -imsvc or /external:I or -external:I
                     const char * end = token.GetEnd();
 
                     // strip quotes if present
@@ -1771,6 +1774,8 @@ bool ObjectNode::BuildArgs( const Job * job, Args & fullArgs, Pass pass, bool us
                     {
                         // Remove relative include
                         StripTokenWithArg_MSVC( "I", token, i );
+                        StripTokenWithArg_MSVC( "imsvc", token, i );
+                        StripTokenWithArg_MSVC( "external:I", token, i );
 
                         // Add full path include
                         fullArgs.Append( token.Get(), (size_t)( start - token.Get() ) );


Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -1748,10 +1748,13 @@ bool ObjectNode::BuildArgs( const Job * job, Args & fullArgs, Pass pass, bool us
...
                 // limit of cl.exe)
-                if ( ( job->IsLocal() == false ) && IsStartOfCompilerArg_MSVC( token, "I" ) )
+                if ( ( job->IsLocal() == false ) && (
+                        IsStartOfCompilerArg_MSVC( token, "I" ) ||
+                        IsStartOfCompilerArg_MSVC( token, "imsvc" ) ||
+                        IsStartOfCompilerArg_MSVC( token, "external:I" ) ) )
                 {
                     // Get include path part
-                    const char * start = token.Get() + 2; // Skip /I or -I
+                    const char * start = token.Get() + 2; // Skip /I or -I or /imsvc or -imsvc or /external:I or -external:I
```
ffulin, line 1757: This still seems to add +2 to the token.Get(), which doesn't seem correct for "imsvc" and "external:I"

poppolopoppo: Yes you are absolutely right,
I'll update this patch soon after additional testing.

ffulin:
It seems /external:I can't currently be supported with distributed compilation. See: #764 for more details.

ffulin:
For reference, I changed this code to fix a couple of issues. I have this shelved should it become possible to use (if the MSVC issues are resolved):
```C++
                if ( job->IsLocal() == false )
                {
                    // Check for various styles of include token
                    static const char * const includeTokens[] =
                    {
                        "I",
                        "imsvc",
                        "external:I",
                    };
                    const char * start = nullptr;
                    for ( const char* includeToken : includeTokens )
                    {
                        if ( IsStartOfCompilerArg_MSVC( token, includeToken ) )
                        {
                            start = token.Get() + ( AString::StrLen( includeToken ) + 1 ); // +1 for - or /
                            break;
                        }
                    }
                    if ( start )
                    {
                        // Get include path part, removing quotes if present
                        if ( *start == '"' )
                        {
                            ++start;
                        }
                        const char * end = token.GetEnd();
                        if ( end[ -1 ] == '"' )
                        {
                            --end;
                        }
                        AStackString<> includePath( start, end );
                        const bool isFullPath = PathUtils::IsFullPath( includePath );

                        // Replace relative paths and leave full paths alone
                        if ( isFullPath == false )
                        {
                            // Remove relative include
                            for ( const char* includeToken : includeTokens )
                            {
                                if ( StripTokenWithArg_MSVC( includeToken, token, i ) )
                                {
                                    break;
                                }
                            }

                            // Add full path include
                            fullArgs.Append( token.Get(), (size_t)( start - token.Get() ) );
                            fullArgs += job->GetRemoteSourceRoot();
                            fullArgs += '\\';
                            fullArgs += includePath;
                            fullArgs.Append( end, (size_t)( token.GetEnd() - end ) );
                            fullArgs.AddDelimiter();

                            continue; // Include path has been replaced
                        }
                    }
                }
```

ffulin:
The issue i originally raised in the Developer Community was incorrectly closed as a duplicate, but I re-reported the issue, split in two:
* https://developercommunity2.visualstudio.com/t/external:I-requires-directory-to-exist/1340525
* https://developercommunity2.visualstudio.com/t/experimental:I-incorrectly-ignores-dire/1306689

It seems those issues have now been closed as 'Fixed In: Visual Studio 2019 version 16.10'. Pending verification of those fixes, we can hopefully resolve this soon.

ffulin:
As with the related issue (https://github.com/fastbuild/fastbuild/issues/764) I think this PR slipped through the cracks. FASTBuild has been working with '/imsvc' and '/external:I' for some time now (since around v1.08 or so), assuming a recent enough version of the MSVC compiler to avoid the issues it had originally.

I'll close this as there are no know outstanding issues. We can of course raise new issues for any problems that are found.

PR #767, opened 2020-07-13
Add support for IPv6 connections

Rinn:
# Description
This change updates the networking code to accept IPv6 connections by changing the listen server to create an IPv6 socket that also accepts IPv4 connections. 

I ensured the remote connection worked correctly on Windows with hostnames being resolved to both IPv4 & IPv6. I wasn't able to test with multiple machines on Linux or MacOS however all code compiled locally and the connection unit tests I added passed and the localhost connection was successful.

Any feedback is appreciated. I'll also say I'm not an expert in networking code so this was written with the help of a lot of documentation.

## Please provide details for the change or fix
Support for ipv6 is desired as servers in our data center are not allocated ipv4 ip addresses by default and it's preferred we only use ipv6 networking

### Core/Process/Semaphore.cpp\h
- Update Semaphore::Wait to return false if time out occurs for use in Network.cpp
### Core/Network/Network.cpp\h
- Add GetAddressFromString helper functions to convert string to binary ip address using inet_pton
- Add GetAddressAsString helper functions to convert binary ip address to string using inet_ntop
- Add NameResolution helper functions to resolve hostnames using getaddrinfo
- Update NameResolutionThreadFunc to check for completion with semaphores, to return ip addresses via shared memory instead of thread exit code, and to be responsible for freeing allocated memory
- Attempt to resolve a hostname to ipv6 but only if the ipv4 lookup fails
### Core/Network/TCPConnectionPool.cpp\h
- Remove GetAddressAsString which was moved to Network.cpp\h
- Update CreateSocket to add af parameter so an ipv6 socket can be created
- Add version of Connect & CreateConnectionThread functions to use ipv6
- Move socket wait for connection logic to helper function so it can be shared
- Listen server now creates an ipv6 socket connection which also accepts ipv4 connections by disabling IPV6_V6ONLY
- Replace ConnectionInfo::GetRemoteAddress with GetRemoteAddressString, as the ip address stored was only used for printing the readable string
- Fix analyze errors when compiling with TCPCONNECTION_DEBUG defined (TCPDEBUG %i -> %u)
### Core/CoreTest/TestNetwork.cpp
- Add unit tests for Network changes
### Core/CoreTest/TestTCPConnectionPool.cpp
- Refactor test functions to take in ip address so they can be shared
- Add test functions to connect with ipv6 instead of ipv4
### Core/CoreTest/TestMain.cpp
- Register TestNetwork tests
### Tools/FBuild/FBuildCore/Protocol/Server.cpp
- Update TCPConnectionPool::GetAddressAsString reference
### Tools/FBuild/FBuildCore/Protocol/Protocol.h
- Bump PROTOCOL_VERSION to 22

# Checklist

## The pull request
- [x] **Is self-contained**


Rinn:
It would appear I'm getting different test results on TravisCI than I did locally and there's an error, I'll investigate.

Rinn:
Unfortunately for me it looks like ipv6 is unavailable with Travis CI on Linux amd64 https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

Rinn:
I'm still looking into the CI results, a run from my fork [passed](https://travis-ci.com/github/Rinn/fastbuild/builds/175478798) but it's possible I could have caused an intermittent error.

Rinn:
Okay I've update this to incorporate the changes from #770 (I needed to replace the colons in ipv6 addresses with a different character as they were not valid for a windows filename, I chose a semicolon as that is not a valid character for a hostname). Something to note, the PROTOCOL_VERSION bump could be removed, the downside would be that any newer worker with an ipv6 address would be unconnectable if the client wasn't updated, however any ipv4 connections should continue to work as before.

PR #769, opened 2020-07-16
Add .SourceMapping_Experimental to Compiler node

jblazquez:
# Description:

This change adds a new `.SourceMapping_Experimental` property to `Compiler` nodes that allows you to remap source file paths as they appear in the debugging information to make them appear to be stored under a different path. This can be useful for reproducible builds (https://reproducible-builds.org/docs/build-path/) or for improving the debugging experience under `gdb` and `lldb` which - unlike the Visual Studio debugger - don't have an **automatic** way to remap source paths when debugging a binary built on another machine (you can set the mappings during debugging, but this allows you to bake that into the binary so you don't have to).

This new option amounts to passing `-fdebug-prefix-map=$_WORKING_DIR_$=$SourceMapping_Experimental$` to the compiler. The key reason for FASTBuild to do this itself via this new property instead of just passing that option yourself via `.CompilerOptions` is to ensure that caching still works across machines that use different source roots.

This only works on GCC and Clang. MSVC doesn't have a source remapping option (but Visual Studio has better support for automatic remapping, so it's much less of an issue).

# Checklist:

The pull request:
- [x] **Is self-contained**
- [x] **Keeps Windows, OSX and Linux at parity**: MSVC doesn't have source mapping capabilities.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/compiler.html
@@ -50,8 +50,9 @@ distribution, caching and more.
   .AllowResponseFile            // (optional) Allow response files to be used if not auto-detected (default: false)
   
   // Temporary Options
-  .UseLightCache_Experimental   // (optional) Enabled experimental "light" caching mode (default: false)
-  .UseRelativePaths_Experimental// (optional) Enabled experimental relative path use (default: false)
+  .UseLightCache_Experimental   // (optional) Enable experimental "light" caching mode (default: false)
+  .UseRelativePaths_Experimental// (optional) Enable experimental relative path use (default: false)
+  .SourceMapping_Experimental   // (optional) Use Clang's -fdebug-source-map option to remap source files
   .ClangFixupUnity_Disable      // (optional) Disable preprocessor fixup for Unity files (default: false)
 }
 </div>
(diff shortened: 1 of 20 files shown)


jblazquez:
I believe remapping doesn't currently work for distributed builds. Should we disable distributed compilation somehow if `.SourceMapping_Experimental` is set?

Code/Tools/FBuild/Documentation/docs/functions/compiler.html
```diff
@@ -50,8 +50,9 @@ <h1>Compiler</h1>
   .AllowResponseFile            // (optional) Allow response files to be used if not auto-detected (default: false)
   
   // Temporary Options
-  .UseLightCache_Experimental   // (optional) Enabled experimental "light" caching mode (default: false)
-  .UseRelativePaths_Experimental// (optional) Enabled experimental relative path use (default: false)
+  .UseLightCache_Experimental   // (optional) Enable experimental "light" caching mode (default: false)
```
jblazquez, line 53: Fixed a small typo: Enabled -> Enable

Code/Tools/FBuild/Documentation/docs/functions/compiler.html
```diff
@@ -50,8 +50,9 @@ <h1>Compiler</h1>
...
+  .UseRelativePaths_Experimental// (optional) Enable experimental relative path use (default: false)
+  .SourceMapping_Experimental   // (optional) Use Clang's -fdebug-source-map option to remap source files
```
jblazquez, line 55: This should say `-fdebug-prefix-map` instead. I'll update on a new commit.

ffulin:
Thanks for the PR. This has been integrated and will be in the next release (v1.02).

I made a follow up change (https://github.com/fastbuild/fastbuild/commit/c63edd46b2b3779206aad87048fa28935001443a) which:

 * Enables the test on Windows
The test was disabled with a comment referring to MSVC, but the test uses Clang on Windows and seems to work just fine.

* Disables distributed compilation when SourceMapping_Experimental is enabled
As the mapping is not available remotely and the code path is disabled, for SourceMapping_Experimental to be respected requires distributed compilation be disabled. I added this to the documentation. I think we can remove this limitation in the future, but we'll need to find a way to transmit the source mapping to the remote host.  This will probably require some changes to the network protocol.

Issue #160, opened 2016-10-28
gethostname doesn't contains the domain name


ktamail666:
Our servers are in different domain. In this case the ip that the worker got was wrong because the WorkerBrokerage don't saved the domain.
Similar at nslookup can return two different ip depend on I used the netbios name or FQDN.
I hope it is understandable.
This little change solve the problem on Windows in Network.cpp:
`TCHAR local[256];`
`DWORD hstSize = sizeof(local);`
`GetComputerNameEx(ComputerNameDnsDomain, local, &hstSize);`
`if ( ::gethostname( buffer, 64 ) == 0 )`
  `{`
       `hostName = buffer;`
     `hostName += ".";`
    `hostName += local;`
 `}`

ffulin:
This seems reasonable enough and presumably we'd need something similar for Linux and OSX.

ffulin:
In the next release (v1.03), the brokerage token name is the IP address instead of the host name. The IP address is used to initiate the connection instead of the hostname, which improves connectivity in various scenarios. That should resolve this issue. I will leave this issue open until that version is released.

ffulin:
v1.03 has now been released and contains the changes described above.

Issue #643, opened 2019-10-15
Mac connect failed.


James-Z:
Hi,
Thanks for your great job, it really helps us a lot on Windows.
But Mac still not working now, when I use '-distverbose' always told me "Connect _another mac_ (Failed)"
Could I get more information from here?
Thanks a lot.

ffulin:
In order to be able to help you, i'll need a little more information:
a) What version of FASTBuild are you running?
b) How did you build the FBuildWorker for OSX?
c) How have you configured the workers (brokerage path or explicit .Workers list in the Settings)?
d) Are you able to ping one mac from the other?

James-Z:
Thanks for the relay.
a) version is 0.98
b) I just build the FASTBUILD source code for the test.
c) BROKERAGE is already setting, remote Mac is set like
"export FASTBUILD_BROKERAGE_PATH=/Volumes/Mac4" **in** "/Users/_username_/.bash_profile file"
d) yes ping is working

and I printf the IP address, and got  "0", maybe `uint32_t hostIP = Network::GetHostIPFromName( host, timeout );` is not work correctly.
Code in`TCPConnectionPool.cpp`:
```c
// Connect
//------------------------------------------------------------------------------
const ConnectionInfo * TCPConnectionPool::Connect( const AString & host, uint16_t port, uint32_t timeout, void * userData )
{
    ASSERT( !host.IsEmpty() );

    // get IP
    uint32_t hostIP = Network::GetHostIPFromName( host, timeout );
    printf("\niiiiiiiiipppppppp name  is : %s\n", host.Get());    // name
    printf("\niiiiiiiiipppppppp address  is : %d\n", hostIP);      // address
    if( hostIP == 0 )
    {
        TCPDEBUG( "Failed to get address for '%s'\n" , host.Get() );
        return nullptr;
    }
    return Connect( hostIP, port, timeout, userData );
}
```

James-Z:
More info.
FBuildWorker is running without any arguments.

_By the way, how could I control the "FBuildWorker" set "Disable" or "Working for others always" on Mac?_

James-Z:
Finally, I found the key to this problem.
Add `192.168.xxx.xxx username.local` into /etc/hosts, FBuild gonna work well.
It looks like `Network::GetHostIPFromName( host, timeout );` can't get remote Mac IP from "_username.local_" correctly.

ffulin:
Thanks for the extra info.

[...]

Issue #271, opened 2017-08-16
Cannot connect to a worker running in virtual machine


Domain:
I am not sure this is an issue or not. I setup a win7 vm, and running FBuildWorker in it. The worker can be found in FASTBUILD_BROKERAGE_PATH, and the firewall is disabled. I can telnet to the port 31264. Everying seems fine. But there is no connections while compiling. Other workers running in real machine work fine.
Host: CentOS 6.9
VM: KVM
Guest: WIndows 7
FBuild: 0.93 x64

ffulin:
I've hosted the worker inside a VM before without issue (in VMWare). If you can telnet in, I can't imagine why it wouldn't work.

Here are some things to try/check:
 - make sure the worker thinks the machine is idle (or is set to "Always" mode) as shown in the UI
 - make sure the worker is writing a token into the brokerage folder
 - try hard coding the IP of the VM in the .Workers section of the Settings() in the bff to force it to use only that machine
 - try the host name instead of the IP to see if that's a factor

Domain:
Thanks! It works well after hard coding the IP! Looks like it is a hostname resolving problem.

ffulin:
Do you think there is something in the way the hostname is being resolved that needs improvement in FASTBuild?

Domain:
I think there should be a switch to use IP instead of hostname, the worker may be not in the same LAN.

ffulin:
I wonder if the worker should always use the IP address. That would save the local host a hostname lookup as well.  The downside of only using the hostname (either always or with an option), is that right now we use the hostname on the local side to display during the build, which helps finding problem machines easier.

It would be nice to find a solution that doesn't require configuration by the user.

Perhaps we could make the marker files contain both the IP and the hostname, either:
 - In the filename: ip.ip.ip.ip.hostname (would need to resolve when two IPs claim to be the same hostname - perhaps by date of the stamp?)
 - Make marker files ip.ip.ip.ip and contain the hostname (would make finding the list of workers slower as we'd need to open each file) This would allow for future builder information to be obtained by the local machine, like HW specs for example. This could be interesting for prioritization of which workers to use)

ffulin:
BTW - If the hostnames of the workers in your pool are all fixed, you can set an explicit list in the Settings() of the bff via the .Workers property. This can be a list of IPs. (I don't think that negates the usefulness of the ip I mention above, but could be useful in some scenarios).

fire:
[...]

PR #770, opened 2020-07-17
Update brokerage file to use ip address instead of host name

Rinn:
# Description:
Updates the brokerage availability file to use IP address instead of host name, to address situations where machines are on different domains, or if a host name is unresolvable but the ip address can be directly connected to.

- Use ip address instead of host name for brokerage filename, if an ip address is not available use the host name instead as a fallback
- Added rerequesting host name and domain name, and resolving the local host name to ip address in WorkerBrokerage on a 5 minute timer. If the ip address changes the previous brokerage file will be deleted and a new one with an updated filename will be written
- Write host name, ip address, and memory threshold to brokerage file
- Write domain name and FQDN to brokerage file if the domain name is not empty

This change should resolve issues https://github.com/fastbuild/fastbuild/issues/160, https://github.com/fastbuild/fastbuild/issues/271, & probably helps https://github.com/fastbuild/fastbuild/issues/643

I'll need to update PR https://github.com/fastbuild/fastbuild/pull/767 if this is accepted first.

# Checklist:

The pull request:
- [x] **Is self-contained**


fsfod:
I ran in to an issue where a worker wasn't connected to because the resolved ip of the broker hostname was stale but the ip in the broker name was accessible . I don't think  TCPConnectionPool::Connect loop should early exit if  a connect fails and there are still more ips from the broker name to try.

I changed the loop to this and the worker was connected to successfully. Idk if the loop should maybe some have accumulated timeout if its setup like this. 
```c
	const ConnectionInfo* conn = NULL;
	for ( const AString& value : hosts )
	{
		hostIP = Network::GetHostIPFromName( value, timeout );
		if ( hostIP == 0 )
		{
			TCPDEBUG("Failed to get address for '%s'\n", host.Get());
			continue;
		}

		conn = Connect( hostIP, port, timeout, userData );

		if ( conn )
		{
			return conn;
		}
	}

	return conn;
```

Rinn:
Yeah I was considering that. It would need to keep track of which IPs were attempted so it doesn't possibly waste time if the dns resolution returns the same ip address that was provided from the broker file.

Rinn:
Another CI build that [passed on my fork](https://travis-ci.com/github/Rinn/fastbuild/builds/176365238), might be a flaky test.

ffulin:
Thanks for the PR. I was finally able to take some time to look it over.

I like the idea of writing the brokerage token in a way that improves connectivity, but the current change is implemented in a way that makes it a breaking change (which you obviously noted as you updated the protocol version) While some users automate the update of their workers (submitting them to source control for example) others have a more manual process, so breaking changes to the protocol can cause non-trivial inconvenience for some users and has been noted by some as a reason for them updating less frequently.

Here are a few things we can talk about to figure out how to move forward:

**Could we just write the qualified hostname OR just the IP?**

If we wrote just the qualified hostname, we could maintain backward compatibility. Similarly, just writing the IP would also be backwards compatible (as well as avoiding a hostname lookup on the client for every worker it connects to). I can see how having both might improve connectivity in some scenarios, but I think either of them alone likely solves for common connectivity problems. Since you implemented this to fix a real-world scenario, I'm curious if you have any more info on the behavior you saw that you could share?

If necessary, we could allow a new version to write two versions of the token to provide a smooth transition, though of course this would add some complexity.

**Network::GetHostName taking a + delimiter list is a bit odd**

It seems strange that a low level function (in the FASTBuild-agnostic Core libraries) is aware of the custom way the higher level FASTBuild code happens to store multiple entries in a string. I think the higher level code should do the tokenization if we retain this logic.

**The FBuildWorker UI doesn't display the correct hostname**

The FBuildWorker UI shows the local hostname. I believe a hostname change requires a reboot (at least on Windows), but assuming the domain can change does not, the UI can now be out of date. I think this can be resolved by having the WorkerWindow ask the brokerage for the hostname instead of retrieving itself.

Rinn:
**Could we just write the qualified hostname OR just the IP?**
Yeah I'm fine with this, for my use case it would need to be the IP instead of hostname because I'm trying to work around corporate DNS issues when on VPN that has issues with resolving user pc hostnames to IP addresses. My original thoughts was to put as much into the filename as I could as a fallback, the FQDN was something I believed would help the three issues I mentioned, but IP alone will probably resolve any connection issue related to not being able to dns resolve.

Another reason I left the host name in there was because the output uses that string to display which worker is being used for a task and I figured hostname would be good to keep as it's more descriptive than IP address and it continued to prevent the need to open the file to get additional information. If that's a non-issue that I'm fine with only ip address.

**Network::GetHostName taking a + delimiter list is a bit odd**
Yeah fair point, I can move the logic, but if we switch the filename to ip only instead of + delimited this will be a non-issue.

**The FBuildWorker UI doesn't display the correct hostname**

Rinn:
Okay I fixed the FBuildWorker UI. Let me know what you think about the filename. If changing it to IP address is what you want I can make that change (falling back to hostname if the ip address isn't available) and I'll remove the + delimited code.

ffulin:
I think using the IP as the filename is the best way to go for now (retains compatibility, fixes your problem and saves some DNS lookups).

Since having the hostname can be useful for human inspection of the brokerage, one thing we could do is write that entry inside the brokerage file. That means in the worst case, someone can do something like a "Find in Files" to find the hostname of a misbehaving IP. Since normal builds won't need to open the file, I don't see a downside to having this additional info there.

How does that sound?

Rinn:
Sure sounds good to me, I'll make that change.

Rinn:
Alright, this should be done. To summarize what is now in this PR:

- Use ip address instead of host name for brokerage filename, if an ip address is not available use the host name instead as a fallback
- Write host name, ip address, and memory threshold to brokerage file
- Write domain name & FQDN to brokerage file if the domain name is not empty
- Host name, domain name, and ip address is reacquired on a timer in WorkerBrokerage to detect any changes and recreate the brokerage file if necessary.

ffulin:
Thanks for the updated PR. This change has been integrated and will be in the next release (v1.03)

PR #774, opened 2020-08-11
Add Apple Silicon support

jblazquez:
# Description:

During WWDC Apple announced the upcoming Apple Silicon Macs. Those Macs will include an ARM CPU instead of an Intel CPU. 

While those ARM CPUs are also capable of running most unmodified `x86_64` programs using the Rosetta 2 dynamic translation tech, that translation will most likely come at a cost, and it would be better to run native `arm64` programs whenever possible.

This change provides support for building FASTBuild for Apple Silicon by adding a new set of `ARMOSX` build configs, parallel to the existing `x64OSX` ones. The resulting executable can then run on the Apple Developer Transition Kit without dynamic translation.

The change also adds two new predefined defines that allow you to test what the native host architecture is from a .bff file: `__X64__` and `__ARM64__`. These are similar to the existing `__WINDOWS__`/`__OSX__`/`__LINUX__` defines, and enable you to write .bff files that can select native toolchains and executables when the host is an Apple Silicon Mac. For example, if you had a native `arm64` version of Python, doing something like this would allow you to run the Python interpreter without dynamic translation:

```
#if __X64__
.PythonExe = 'External/python/x64/bin/python3'
#endif
#if __ARM64__
.PythonExe = 'External/python/arm64/bin/python3'
#endif

Exec('Run-Python')
{
    .ExecExecutable = .PythonExe
    // ...
}
```

# Checklist:

The pull request:
- [X] **Is self-contained**
- [N/A] **Has accompanying tests**
- [N/A] **Keeps Windows, OSX and Linux at parity** (FASTBuild doesn't currently have ARM64 build configs for Windows and Linux, but they could be added)


Code/Core/CoreTest/Tests/TestTimer.cpp
```diff
@@ -32,7 +32,7 @@ void TestTimer::Validate() const
     Timer t;
     t.Start();
     const int64_t before = t.GetNow();
-    Thread::Sleep( 10 ); // sleep for 100ms
+    Thread::Sleep( 100 ); // sleep for 100ms
```
jblazquez, line 35: On Apple Silicon, a 10ms sleep was not enough and would trigger the `TEST_ASSERT` on line 44. Maybe the timer resolution on the DTK machines is larger than that.

Code/Core/Time/Timer.cpp
```diff
@@ -59,7 +64,12 @@ int64_t Timer::GetNow()
         VERIFY( QueryPerformanceCounter( &now ) );
         return now.QuadPart;
     #elif defined( __APPLE__ )
-        return (int64_t)mach_absolute_time();
+        #if defined( __ARM64__ )
+            // mach_absolute_time seems to return the wrong time on Apple Silicon
```
jblazquez, line 68: From some limited testing, it looked like `mach_absolute_time` was not properly returning the elapsed time, and would cause certain tests to hang. The documentation for `mach_absolute_time` says to prefer the `clock_gettime_nsec_np` function, so that's what I'm using here for the Apple Silicon builds, which fixes the issue and makes all tests pass.

I'm not using `clock_gettime_nsec_np` on Intel Macs because that function was introduced on macOS 10.12, and I assume FASTBuild would not like to bump its macOS minspec that high. Checking only for `__ARM64__` is fine because only macOS 11.0+ can run on ARM CPUs, so `clock_gettime_nsec_np` is guaranteed to be available there.

Code/Tools/FBuild/FBuildTest/Data/testcommon.bff
```diff
@@ -45,6 +45,18 @@
                             + .CommonCompilerOptions
                             + ' -std=c++11'
     ]
+
+    #if CLANG_SUPPORTS_ARMOSX
```
jblazquez, line 49: See `Clang12.bff`

External/SDK/Clang/OSX/Clang12.bff
```diff
@@ -0,0 +1,117 @@
+// Apple Clang 12.0.0
```
jblazquez, line 1: This is the version of Clang from Xcode 12.0, which is currently in Beta.

jblazquez:
Here's `x86_64` FASTBuild running on an Apple DTK under Rosetta 2 translation, building `arm64` FASTBuild:

```
$ ./fbuild All-ARMOSX-Release
...
3> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Release/Core/CoreTest/CoreTest
1> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Release/Tools/FBuild/FBuildTest/FBuildTest
3> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Release/Tools/FBuild/FBuildApp/FBuild
4> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Release/Tools/FBuild/FBuildWorker/FBuildWorker
FBuild: OK: All-ARMOSX-Release
Time: 12.166s
```

Here's proof that the resulting executable is a native `arm64` executable:

```
$ file /private/tmp/fastbuild/tmp/ARMOSX-Release/Tools/FBuild/FBuildApp/FBuild
/private/tmp/fastbuild/tmp/ARMOSX-Release/Tools/FBuild/FBuildApp/FBuild: Mach-O 64-bit executable arm64
```

Here's `arm64` FASTBuild running natively on an Apple DTK, building and running native `arm64` tests, which all pass:

```
$ /private/tmp/fastbuild/tmp/ARMOSX-Release/Tools/FBuild/FBuildApp/FBuild Tests
...
8> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Release/Core/CoreTest/CoreTest
5> Running Test: /private/tmp/fastbuild/tmp/ARMOSX-Release/Core/CoreTest/TestOutput.txt
3> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Debug/Tools/FBuild/FBuildTest/FBuildTest
6> Exe: /private/tmp/fastbuild/tmp/ARMOSX-Debug/Core/CoreTest/CoreTest
2> Running Test: /private/tmp/fastbuild/tmp/ARMOSX-Debug/Core/CoreTest/TestOutput.txt
4> Running Test: /private/tmp/fastbuild/tmp/ARMOSX-Debug/Tools/FBuild/FBuildTest/TestOutput.txt
FBuild: OK: Tests
Time: 1m 0.146s
```

ffulin:
Thanks for the changes! These have been integrated and will be in the next release (v1.03). 

In the future, I'll look into adding an additional build step that can merging the x64 and ARM binaries we can now build into "Universal 2" binaries, but I'm not sure with which release I can start shipping those.

ffulin:
I made a small followup change which restores the sleep test to 10ms for non-ARM binaries. (I added a small TODO to look into why the granularity for Apple Silicon is so poor - Could be that we need to change the code Sleep is calling, or it could be that this is somehow a deliberate change by Apple)  (https://github.com/fastbuild/fastbuild/commit/6464b5109e937b4719323bb017b3e399729810be)

PR #791, opened 2020-10-25
Linux: Set thread stack size always above PTHREAD_STACK_MIN.

uucidl:
# Description:

On AArch64 (ARM64) on Linux, `Thread::CreateThread` would fail when calling
`pthread_attr_setstacksize` returned EINVAL, with stackSize=65536.

This fixes it by honoring the documented minimum limit: PTHREAD_STACK_MIN.

# Checklist:

The pull request:
- [ ] **Is created against the Dev branch**
- [ ] **Is self-contained**
- [ ] **Compiles on Windows, OSX and Linux**
- [ ] **Has accompanying tests**
- [ ] **Passes existing tests**
- [ ] **Keeps Windows, OSX and Linux at parity**
- [ ] **Follows the code style**
- [ ] **Includes documentation**

PR #792, opened 2020-10-25
Core: adapt to Aarch64 on Linux (ARM64)

uucidl:
# Description:

This makes the necessary changes in Core to support ARM64 in Linux.

I chose to use the __ARM64__ define that was introduced recently for ARM on OSX, rather than
use a gcc specific define.

# Checklist:

The pull request:
- [ ] **Is self-contained**

### DIFF
--- a/Code/Core/Env/Assert.h
+++ b/Code/Core/Env/Assert.h
@@ -13,7 +13,11 @@
 #elif defined( __APPLE__ )
     #define BREAK_IN_DEBUGGER __builtin_trap();
 #elif defined( __LINUX__ )
-    #define BREAK_IN_DEBUGGER __asm__ __volatile__("int $3")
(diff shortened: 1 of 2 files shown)


uucidl:
I've got a private branch that compiles on Linux/AArch64 against Linux/AArch64. I bootstrapped it using a Makefile first, then
made adaptation locally to fbuild.bff, then built succesfully FASTbuild with the boostrapped fbuild.

This #792 together with #791 were the only fixes of importance.

ffulin:
This seems like a good change. I don't have plans to natively support ARM linux at this time, but it's great to have the option, particularly for so little change.

ffulin:
This change will be part of the source for the next release (v1.03). 

Thanks!

PR #794, opened 2020-10-26
Fix minor doc typo

matklad:
# Description:

Please provide details for the change or fix:
 - (For changes or new functionality) Detail the motivation/use-case behind the change and any context required to understand it.
 - (For fixes) Describe the bug/problem being fixed.
 - (If appropriate) Link to any associated issues.

# Checklist:

The pull request:
- [ ] **Is created against the Dev branch**
- [ ] **Is self-contained**
- [ ] **Compiles on Windows, OSX and Linux**
- [ ] **Has accompanying tests**
- [ ] **Passes existing tests**
- [ ] **Keeps Windows, OSX and Linux at parity**
- [ ] **Follows the code style**
- [ ] **Includes documentation**

PR #797, opened 2020-10-31
Revert "Fix minor doc typo"

ffulin:
Reverts fastbuild/fastbuild#794


ffulin:
I applied this change, but then realized it was created against the master branch. I've reverted the changes and manually applied the fix to the dev branch. This documentation fix will be part of the next release (v1.03).

Thanks for the fix!

PR #811, opened 2021-01-13
Add Intellisense support for MSVC /external:I paths

jblazquez:
# Description:

This change makes the FASTBuild Intellisense logic understand the new `/external:I` MSVC options for system include paths. These options are basically like Clang/GCC's `-isystem` when it comes to compiler warning behavior.

See this blog post for details on these new MSVC options: https://devblogs.microsoft.com/cppblog/broken-warnings-theory/#proposed-solution 

# Checklist:

The pull request:
- [X] **Is self-contained**


ffulin:
This looks like a good change. As well as improving project generation for Intellisense, it also adds support for /external:I style include paths for the LightCache caching mode.

Looking into this compiler feature a bit more, it seems /extenral:I is incompatible with FASTBuild's distributed compilation mechanism.

There are several issues, which I've reported on the Developer Community forums: https://developercommunity2.visualstudio.com/t/experimental:external-does-not-work-wit/1313253

I don't think that would prevent this PR being merged, but it does mean that if you start using /external:I, you won't be able to use distributed compilation at the same time :(

ffulin:
(A related issue https://github.com/fastbuild/fastbuild/issues/764 and PR https://github.com/fastbuild/fastbuild/pull/765)

ffulin:
Thanks for the improvement. This change has been integrated and will be in the next release (v1.05).

For reference, these two compiler bugs will need to be fixed before /external:I can effectively be used with FASTBuild distributed compilation:

**/external:I requires directory to exist, inhibiting distributed compilation (VS2019 16.8.3)**
 - https://developercommunity2.visualstudio.com/t/external:I-requires-directory-to-exist/1340525
 
**/experimental:I incorrectly ignores directories in #line**
 - https://developercommunity2.visualstudio.com/t/experimental:I-incorrectly-ignores-dire/1306689

PR #817, opened 2021-02-13
Fix response file argument computation

Ace17:
Here's what you get if you try to use response files with clang under GNU/Linux:

```
4> clang: error: no such file or directory: '@"/tmp/_fbuild.tmp/0x8a3452e1/core_4/args.rsp"'
clang: error: no input files
```

The problem here is the double quotes.

The command-line using a response file should look like this:
```
 [...] "@/path/to/my/reponse/file/with spaces/in/it"  [...]
```

But it actually is like this ( the '@' sign and leading double-quote are swapped ).
```
 [...] @"/path/to/my/reponse/file/with spaces/in/it"  [...]
```

This will prevent the double-quotes to be interpreted by the shell, which lead the above argument to be seen as two arguments:
```
@"/path/to/my/reponse/file/with
spaces/in/it"
```

More problematic: it breaks clang (at least under GNU/linux), because clang doesn't know what to do with double-quotes (they should be interpreted by the shell, but they shouldn't be part of the argv received by the application).

For reference, here's the clang code that reads the response files in the command line :  https://github.com/llvm/llvm-project/blob/e3f02302e318837d2421c6425450f04ae0a82b90/llvm/lib/Support/CommandLine.cpp#L1139 

# Checklist:

The pull request:
- [X] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/Args.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/Args.cpp
@@ -171,7 +171,7 @@ bool Args::Finalize( const AString & exe, const AString & nodeNameForError, Args
     }
 
     // Create new args referencing response file
-    m_ResponseFileArgs = "@\"";
+    m_ResponseFileArgs = "\"@";
     m_ResponseFileArgs += m_ResponseFile.GetResponseFilePath();
     m_ResponseFileArgs += "\"";
 
--- a/Code/Tools/FBuild/FBuildTest/Data/TestArgs/ResponseFile/fbuild.bff
+++ b/Code/Tools/FBuild/FBuildTest/Data/TestArgs/ResponseFile/fbuild.bff
@@ -1,0 +1,35 @@
+//
+// TestArgs - ResponseFile
+//
+// Check the invocation command-line
+//
+//------------------------------------------------------------------------------
+
+// Use the standard test environment
+//------------------------------------------------------------------------------
+#include "../../testcommon.bff"
+Using( .StandardEnvironment )
+Settings {}
+{
+    // Common settings
+    .CompilerInputPath          = 'Tools/FBuild/FBuildTest/Data/TestArgs/ResponseFile/'
+    .UnityOutputPath            = '$Out$/Test/ObjectList/ResponseFile/'
+    .CompilerOutputPath         = '$Out$/Test/ObjectList/ResponseFile/'
+
+    Compiler( 'ClangAlwaysUsingResponseFiles' )
+    {
+      .Root                     = '$Clang10_BasePath$'
+      .Executable               = '$Root$/clang'
+      .ForceResponseFile        = true
+    }
+
+    ObjectList( 'ResponseFilePath' )
+    {
+        .Compiler = 'ClangAlwaysUsingResponseFiles'
+    }
+}
+
+Alias( 'Test' )
+{
+    .Targets = { 'ResponseFilePath' }
+}
--- a/Code/Tools/FBuild/FBuildTest/Data/TestArgs/ResponseFile/file.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Data/TestArgs/ResponseFile/file.cpp
@@ -1,0 +1,4 @@
+// simple valid file
+void Function()
+{
+}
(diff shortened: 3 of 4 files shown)


ffulin:
Thanks for you PR. This fix seems sensible, and I am working to clear the backlog of PRs.

Although this fix seems trivial, this code path is shared with non-Clang compilers and is a key part of FASTBuild so some regression tests will be needed. If you are able to add those tests (checking clang and non-clang compilers when using response file args) that would make it possible to integrate sooner, assuming that doesn't uncover anything unexpected.

Ace17:
> If you are able to add those tests (checking clang and non-clang compilers when using response file args) 

The README says there are "unit tests" and "functional tests".

I believe "checking clang and non-clang compilers"doesn't fit into "unit tests",  as it requires actually invoking the compilers. Especially when the unit tests for the response files do this:

```
    // Prevent actual write to disk as this depends on running in a WorkerThread
    args.DisableResponseFileWrite();
```

So I'm looking for the "functional tests".

I ran "fbuild Exes Tests".

The build was successful, however 3 tests failed on the current "dev" branch (I'm using GNU/Linux, maybe they pass under Windows?).

The only mentions of "response file" I could find in "tmp/x64Linux-Debug/Tools/FBuild/FBuildTest/TestOutput.txt" seems to be related to unit tests.

The "TestFramework" directory only contains source files with "UnitTest" in their names.

Where are the "functional tests" ?
Do we currently do integration testing with clang and non-clang compilers? If yes, where is it?

Thanks !

ffulin:
*> The README says there are "unit tests" and "functional tests".*

The terminology used to describe the test in various places (including some class names in the TestFramework) is not correct. The terms "unit test" and "functional test" are used interchangeably and incorrectly in some cases.

Poor naming aside, it's possible to implement tests as needed. For most libraries there is an accompanying test project. It is perfectly fine (and necessary) to invoke compilers in tests, and those generally live at the FBuildTest level (as they test the FBuild) functionality.

Tests are maintained and run on all platforms and msvc, clang and gcc are tested depending on the platform. TestArgs is the class where response file management is currently tested. It would be appopriate to extend that with functional tests. Other test cpp files in the same directory have example of compiler invocations that could be used as a reference (for example TestObjectList::CompilerInputFilesRoot is a fairly minimal test that ensures invocation passes without error).

Ace17:
Thanks for the info.

I've added a test that shows the issue, here's what it looks like (without the fix, of course).

```
 - Test 'ResponseFile_IfNeeded_Long' in 0.023s : PASSED
 - Test 'ResponseFile_Always_Short'
 - Test 'ResponseFile_Always_Short' in 0.000s : PASSED
 - Test 'ResponseFile_Always_Long'
 - Test 'ResponseFile_Always_Long' in 0.021s : PASSED
 - Test 'ResponseFile_CommandLineQuoting'
Obj: /home/ace/source/fastbuild/tmp/Test/ObjectList/ResponseFile/file.o
Failed to build Object. Error: 1 (0x01) Target: '/home/ace/source/fastbuild/tmp/Test/ObjectList/ResponseFile/file.o'
PROBLEM: /home/ace/source/fastbuild/tmp/Test/ObjectList/ResponseFile/file.o
clang: error: no such file or directory: '@"/tmp/_fbuild.tmp/0x7beeea45/core_1/args.rsp"'
clang: error: no input files
--- Most Expensive ----------------------------------------------
Time (s)  Name:
0.053     /home/ace/source/fastbuild/tmp/Test/ObjectList/ResponseFile/file.o

--- Summary -----------------------------------------------------
                                 /----- Cache -----\
```

Ace17:
OK, seems like I could use some help here.

For this test, I want to `ForceResponseFile=true`. This is specified at "Compiler( ...)" level.
Ideally , what I'd like to do is to locally declare a new Compiler which would be "Clang + ForceResponseFile=true".
If possible, without polluting the test config with OS-specific stuff.

Is it possible to declare a "Compiler()" that would somehow "inherit" from another "Compiler()" ?

leinlin:
When to fix this problem,I meet this bug too when us macos and ndk clang++ to link a dll
.NDKToolchains       = '$NDKRoot$\toolchains'
#if __WINDOWS__
.NDKToolchainClangRoot = '$NDKToolchains$\llvm\prebuilt\windows-x86_64\bin'
#endif
#if __OSX__
.NDKToolchainClangRoot = '$NDKToolchains$\llvm\prebuilt\darwin-x86_64\bin'
#endif

        .LinkerForceResponseFile = true
        .LinkerType = 'clang-orbis'
        .Linker             = '$NDKToolchainClangRoot$\clang++'

ffulin:
Thanks for the fix and the test. This has been integrated and will be in the next release (v1.09)

I made some minor changes to the test to make it agnostic to Clang version: https://github.com/fastbuild/fastbuild/commit/58265b3944f650f85ac8a66d7ad4074e0e47410a

PR #835, opened 2021-05-17
Add workerlimit to modify the number of workers

NguyetLu:
* Add parameter `workerlimit <number>`
* User can modify the number worker limit if they want. This will take full advantage of its productivity if they have more workers.
* Default value if don't have this parameter still 15 (like default)
* We tested on a project which has the .so file size 333 MB
  * 15 workers: Time: 10m 26.904s
  * 20 workers: Time: 9m 28.330s


NguyetLu:
Hello, 
We saw **pull_request_template.md,** pull_request will be created against the Dev branch. However, the dev_branch doesn't have **GCC_CI.bff**, it is in main branch. Could you please help us check?

NguyetLu:
> Hello,
> We saw **pull_request_template.md,** pull_request will be created against the Dev branch. However, the dev_branch doesn't have **GCC_CI.bff**, it is in the main branch. Could you please help us check?

And we also saw some set up for visual 2019 and gcc9 are not correcy. Could you please check again?

ffulin:
*> the Dev branch doesn't have GCC_CI.bff, it is in the main branch. And we also saw some setup for visual 2019 and GCC9 are not correct.*

Travis CI is no longer supported (they effectively removed support for open source projects) so the _CI files have been removed in the dev branch.

ffulin:
I just want to be sure you know that the worker limit can already be configured in the bff file in the Settings(). Is the option you have added still needed?

Code/Tools/FBuild/FBuildCore/FBuildOptions.h
```diff
@@ -70,6 +70,9 @@ class FBuildOptions
     uint32_t    m_CacheTrim                         = 0;
     int32_t     m_CacheCompressionLevel             = -1; // See Compresssor.h
 
+    // Worker Connection Limit
+    int         m_workerlimit                       = 15;
```
ffulin, line 74: Can you please make this a uint32_t to adhere to the code style, as well as making it unsigned (since negative numbers make no sense in this case)

Can you also please rename the variable to "m_WorkerLimit" to adhere to the code style?

Code/Tools/FBuild/FBuildCore/FBuildOptions.cpp
```diff
@@ -584,6 +604,7 @@ void FBuildOptions::DisplayHelp( const AString & programName ) const
             " -cacheinfo        Output cache statistics.\n"
             " -cachetrim <size> Trim the cache to the given size in MiB.\n"
             " -cacheverbose     Emit details about cache interactions.\n"
+            " -m_WorkerLimit <number_worker>\n"
```
ffulin, line 607: This documentation update seems to list the member variable name instead of the command line option name.

The style also lists all existing options in all lower-case.

Code/Tools/FBuild/FBuildCore/FBuildOptions.cpp
```diff
@@ -178,6 +178,26 @@ FBuildOptions::OptionsResult FBuildOptions::ProcessCommandLine( int argc, char *
...
+            {
+                const int sizeIndex = (i + 1);
+                PRAGMA_DISABLE_PUSH_MSVC(4996) // This function or variable may be unsafe...
+                PRAGMA_DISABLE_PUSH_CLANG_WINDOWS("-Wdeprecated-declarations") // 'sscanf' is deprecated: This function or variable may be unsafe...
+                if ((sizeIndex >= argc) ||
+                     (sscanf(argv[sizeIndex], "%i", &m_workerlimit) != 1))
+                PRAGMA_DISABLE_POP_CLANG_WINDOWS // -Wdeprecated-declarations
+                PRAGMA_DISABLE_POP_MSVC // 4996
+                {
+                    OUTPUT("FBuild: Error: Missing or bad <worker> for '-workerlimit' argument\n");
```
ffulin, line 191: This error message refers to "bad <worker>" which is not the name of the parameter (number_worker)

ffulin (changes requested):

Code/Tools/FBuild/FBuildCore/FBuildOptions.cpp
NguyetLu: Yes, I'll correct it.

Code/Tools/FBuild/FBuildCore/FBuildOptions.h
NguyetLu: Yes, I'll rename this variable.

Code/Tools/FBuild/FBuildCore/FBuildOptions.cpp
NguyetLu: Could I modify the output?
`OUTPUT("FBuild: Error: Missing <number_worker> for '-workerlimit' argument\n")`

NguyetLu:
> I just want to be sure you know that the worker limit can already be configured in the bff file in the Settings(). Is the option you have added still needed?

Yes and I think this will be more convenient for us and users

NguyetLu (commented):
I have updated the news. Could you please check them?

NguyetLu:
Hi,
Could you help to review and merge this?

ffulin:
Thank you for the PR.

Unfortunately, this change cannot be integrated as it doesn't function correctly. The change alters the default WorkerConnectionLimit, which means it will only ever work once, the first time FASTBuild is invoked. After that, the SettingsNode is saved to the FASTBuild .fdb file, and the command line will be ignored.

Some extra logic will be required to make this command line function as an override.

Just in case you missed it, I asked before if you were aware that you can already set the WorkerConnectionLimit inside the bff file, which might negate the need for this feature entirely. FASTBuild will try to use the workers upto that limit, but there is little overhead for extra workers that can't be fed with data, so you can usually increase this for all users if your codebase can benefit from more workers. Given that, do you think a command line override is still useful/needed?

NguyetLu:
Hi,

Thank you for your reply.
If we can modify it on bff, so we can skip this change.
However, I would like to ask for more information, could you help me clear them?
- Could you share with me the configuration on bff for worker limit?
- As you said before, this is only the config at the first time if I used the command-line => So, is it the same with configuration on bff or which priority between bff and database?

ffulin:
The Settings function is briefly documented here: https://fastbuild.org/docs/functions/settings.html

It can be used as follows:
```
Settings
{
    .WorkerConnectionLimit = 20
}
```

The .bff files are parsed when:
a) there is no .fdb (the first ever build)
OR
b) when a .bff file changes

In this way, the WorkerConnectionLimit will always reflect what's specified in the .bff.

NguyetLu:
Hi,
Thank you for your explanation. It's clear for me now. So I'll close this Pull Request.

PR #837, opened 2021-06-01
Update ObjectNode.cpp

yuhongchi:
# Description:

 - The source file is saved in utf8-bom. When compiling Chinese strings in distributed mode, c2001: newline in constant error will be reported. However, the temporary file is saved in utf8. After adding utf8-bom, the compilation passes

# Checklist:

The pull request:
- [ ] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -2052,6 +2052,23 @@ bool ObjectNode::WriteTmpFile( Job * job, AString & tmpDirectory, AString & tmpF
             return NODE_RESULT_FAILED;
         }
     }
+    {
+        // add UTF8-BOM to source files
+        const char * lastDot = fileName.FindLast( '.' );
+        if ( ( lastDot != nullptr ) && ( lastDot[1] != '\0' ) )
+        {
+            AStackString<> extension( lastDot + 1 );
+            if ( ( extension == "cpp" ) || ( extension == "cc" ) || ( extension == "cxx" ) || ( extension == "c++" || extension == "cp" || extension == "CPP" || extension == "C" || extension == "inc" ) )
+            {
+                unsigned char c1 = 0xEF;
+                unsigned char c2 = 0xBB;
+                unsigned char c3 = 0xBF;
+                tmpFile.WriteBuffer(&c1, 1);
+                tmpFile.WriteBuffer(&c2, 1);
+                tmpFile.WriteBuffer(&c3, 1);
+            }
+        }
+    }
     if ( tmpFile.Write( dataToWrite, dataToWriteSize ) != dataToWriteSize )
     {
         job->Error( "Failed to write to temp file. Error: %s TmpFile: '%s' Target: '%s'", LAST_ERROR_STR, tmpFileName.Get(), GetName().Get() );


ffulin:
Unfortunately this change cannot be integrated because it's not safe to assume that all compilers or use-cases support UTF-8:
 - The source file may already have some other encoding
 - Some compilers simply do not support UTF-8 at all, or may not accept a UTF-8 file with a BOM
 - Compilers can be any executable, including user created tools with any number of additional restrictions or requirements

Additionally, there are some other problems with this PR:
 - **created against the wrong branch**  PRs are only accepted against the "dev" branch.
 - **no tests** Functional changes, particularly global changes like this require accompanying tests.
 - **Breaks worker compatibility** Preprocessing like this would have to be done on the client side before sending over the network to make sure it works reliably with existing workers.

Based on the error you mention (C2001) it seems you are using the MSVC compiler. I tried to reproduce the problem you mention with distributed compilation for text that contains Chinese characters but didn't have any luck. Things seem to work as expected. I am using Visual Studio 2019 v16.10 (compiler v14.29.30037)

Are you using the /utf-8 command line option? (https://docs.microsoft.com/en-us/cpp/build/reference/utf-8-set-source-and-executable-character-sets-to-utf-8?view=msvc-160)

If that doesn't solve things for you, then if you can provide some more information and ideally a repro-case, then I might be able to suggest a path forward. Given the nature of UTF-8, it's quite possible that only very specific code points will reproduce this problem. Knowing which compiler version you are using would also be helpful.

yuhongchi:
main.cpp is saved as utf8-bom
I use visual studio's build: I can compile it successfully, however if I changed to utf8, it will report C2001.
As the same, When I use fastbuild, I can compile it if I don't use -dist. but it will report C2001 when I open -dist
[test.zip](https://github.com/fastbuild/fastbuild/files/6607940/test.zip)

ffulin:
Thank you for the test file. I tried to reproduce the problem, using that file, but was unable to do so. I tried VS2015, 2017 and 2019 and all of them worked fine whether compiled locally or remotely.

Ideally, if you can create a stand-alone bff file to repro that problem, that would also be helpful.

The following information would also be useful:
 - the Visual Studio version
 - the FASTBuild version

yuhongchi:
[test.zip](https://github.com/fastbuild/fastbuild/files/6613278/test.zip)
the Visual Studio version: VS2013
the FASTBuild version: v1.0.5
I have add fbuild.bff to test.zip
in fbuild.bat, failed if open -dist, but it's succeed without -dist

ffulin:
I tried with VS2013 and was able to compile without any errors:
```
Obj: C:\test\utf8\Debug\main.obj
-> Obj: C:\test\utf8\Debug\main.obj <REMOTE: xxx.xxx.xxx.xxx>
FBuild: OK: action_0```

I had to change my TEMP & TMP directories in the bff to a location that doesn't require admin rights, but that was the only local change.

The Visual Studio 2013 compiler version I have available is:
```Microsoft (R) C/C++ Optimizing Compiler Version 18.00.40629 for x86```

ffulin:
The only idea I have is that this is somehow related to the installation locale of Visual Studio. I don't have the installation media available for 2013 to be able to try and install the locale you are using (2052). Are you able to try switching VS to English to see if the problem goes away?

ffulin:
As noted above, this change cannot be integrated as is and I was not able to reproduce the issue this was intended to address. More information about this problem would be required to be able to propose a fix.

PR #843, opened 2021-06-26
TestProjectGeneration: Expect quoted paths in IntellisenseAndCodeSense test

dummyunit:
# Description:

This patch fixes the `IntellisenseAndCodeSense` test in case when the full path to the repository checkout contains any character that forces `XCodeProjectGenerator` to write quoted include paths to the `.pbxproj` file (see `XCodeProjectGenerator::ShouldQuoteString()`).

Previously the test didn't account for possibility of a quoted path and the trailing quote was left in paths saved to the includes array.

# Checklist:

The pull request:
- [x] **Is self-contained**
- [ ] **Compiles on Windows, OSX and Linux** (I tested it only on Linux x86_64.)


ffulin:
Thank you!

PR #845, opened 2021-06-28
Fix Compress() and Decompress() in MultiBuffer

dummyunit:
# Description:

Calls to `Replace()` in `Compress()` and `Decompress()` in `MultiBuffer` worked correctly only when `Compressor::GetResultSize()` from the second argument was called before `Compressor::ReleaseResult()` from the first argument.

But the order of evaluation of function arguments is unspecified in C++, so the code worked with some compilers (eg. GCC 10) and produced incorrect results (zero data size in `m_ReadStream` or `m_WriteStream`) with others (eg. Clang 12).

In the end this problem was caught by several tests:
* TestBuildFBuild.BuildCleanWithCache
* TestCache.ConsistentCacheKeysWithDist
* TestPrecompiledHeaders.TestPCHWithCache

# Checklist:

The pull request:
- [x] **Is self-contained**
- [ ] **Compiles on Windows, OSX and Linux** (I tested it only on Linux x86_64.)

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/MultiBuffer.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/MultiBuffer.cpp
@@ -162,7 +162,8 @@ void MultiBuffer::Compress( int32_t compressionLevel )
     c.Compress( m_WriteStream->GetData(), m_WriteStream->GetSize(), compressionLevel );
 
     // Transfer compressed results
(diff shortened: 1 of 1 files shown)


ffulin:
Great fix - this makes total sense. Thanks!

I saw this fail on OSX recently, but hadn't had a chance to dig into the cause. Somewhat surprisingly, the tests pass with Clang on Windows.

PR #846, opened 2021-06-29
Use MSVC headers in x64Clang configuration

dummyunit:
# Description:

This fixes build on Windows with Clang using VS2015 or VS2019 and new WinSDK versions.
Currently the build fails with errors about not found include files `excpt.h` and `vcruntime.h`, both are included in WinSDK headers:
```
C:\Program Files (x86)\Windows Kits\10/Include/10.0.19041.0/um\windows.h(167,10): fatal error: 'excpt.h' file not found
C:\Program Files (x86)\Windows Kits\10/Include/10.0.19041.0/ucrt\corecrt.h(10,10): fatal error: 'vcruntime.h' file not found
```

This was observed with Clang 11 (installed via Visual Studio installer) and WinSDK 10.0.19041.0.

We need to specify the include path to MSVC headers using `/imsvc` because otherwise the build fails due to warnings in MSVC headers reported Clang. So the `.VSIncludePaths` variable couldn't be reused.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/External/SDK/Clang/Windows/Clang11.bff
+++ b/External/SDK/Clang/Windows/Clang11.bff
@@ -80,6 +80,7 @@ Compiler( 'Compiler-Clang11-NonCL' )
 
                                     // Include paths
                                     + ' -I"./"'
+                                    + .VSIncludePaths_ClangCl
 
                                     // x64
                                     + ' -m64'
--- a/External/SDK/Clang/Windows/Clang7.bff
+++ b/External/SDK/Clang/Windows/Clang7.bff
@@ -72,6 +72,7 @@ Compiler( 'Compiler-Clang7-NonCL' )
 
                                     // Include paths
                                     + ' -I"./"'
+                                    + .VSIncludePaths_ClangCl
 
                                     // x64
                                     + ' -m64'
--- a/External/SDK/Clang/Windows/Clang8.bff
+++ b/External/SDK/Clang/Windows/Clang8.bff
@@ -80,6 +80,7 @@ Compiler( 'Compiler-Clang8-NonCL' )
 
                                     // Include paths
                                     + ' -I"./"'
+                                    + .VSIncludePaths_ClangCl
 
                                     // x64
                                     + ' -m64'
--- a/External/SDK/Clang/Windows/Clang9.bff
+++ b/External/SDK/Clang/Windows/Clang9.bff
@@ -81,6 +81,7 @@ Compiler( 'Compiler-Clang9-NonCL' )
 
                                     // Include paths
                                     + ' -I"./"'
+                                    + .VSIncludePaths_ClangCl
 
                                     // x64
                                     + ' -m64'
--- a/External/SDK/VisualStudio/VS2015.bff
+++ b/External/SDK/VisualStudio/VS2015.bff
@@ -37,6 +37,7 @@ Compiler( 'Compiler-VS2015-x64' )
 
     // Paths
     .VSIncludePaths                 = ' -I"$VS2015_BasePath$/VC/include/"'
+    .VSIncludePaths_ClangCl         = ' /imsvc "$VS2015_BasePath$/VC/include/"'
     .VSLibPaths                     = ' /LIBPATH:"$VS2015_BasePath$/VC/lib/amd64"'
     .VCPackagesPath                 = '$VS2015_BasePath$/VC/vcpackages'
 
--- a/External/SDK/VisualStudio/VS2017.bff
+++ b/External/SDK/VisualStudio/VS2017.bff
@@ -85,6 +85,7 @@ Compiler( 'Compiler-VS2017-x64' )
 
     // Paths
     .VSIncludePaths                 = ' -I"$VS2017_ToolchainPath$/include/"'
+    .VSIncludePaths_ClangCl         = ' /imsvc "$VS2017_ToolchainPath$/include/"'
     .VSLibPaths                     = ' /LIBPATH:"$VS2017_ToolchainPath$/lib/x64"'
     .VCPackagesPath                 = '$VS2017_BasePath$/../Common7/IDE/VC/vcpackages'
 
--- a/External/SDK/VisualStudio/VS2019.bff
+++ b/External/SDK/VisualStudio/VS2019.bff
@@ -85,6 +85,7 @@ Compiler( 'Compiler-VS2019-x64' )
 
     // Paths
     .VSIncludePaths                 = ' -I"$VS2019_ToolchainPath$/include/"'
+    .VSIncludePaths_ClangCl         = ' /imsvc "$VS2019_ToolchainPath$/include/"'
     .VSLibPaths                     = ' /LIBPATH:"$VS2019_ToolchainPath$/lib/x64"'
     .VCPackagesPath                 = '$VS2019_BasePath$/../Common7/IDE/VC/vcpackages'
 


ffulin:
Interestingly I'm using Clang 11 installed directly from installers available here: https://releases.llvm.org/download.html and I don't have any compile errors. It's curious that the Visual Studio install would be different in a way that would cause this problem.

I'll try to find some time to install via the Visual Studio UI and see I can reproduce this.

ffulin:
I spent some time digging into this, and wasn't able to reproduce the problems you mentioned.

I switched FASTBuild to the latest Windows SDK (https://github.com/fastbuild/fastbuild/commit/a6e5b46c1821995e7c1cad60bb45fdcd5467ea58). I ran into some static analysis issues (which I resolved in that change). With VS2015 I uncovered an existing issue that I also fixed (https://github.com/fastbuild/fastbuild/commit/1c20a8218239bb0709328fcd43b8f4f7e84a6c83)

I didn't run into any problems that required me to add additional include paths however.

I also installed Clang from Visual Studio (as opposed to using the installers directly) just to make sure that wasn't a factor.

It's not obvious to me what might be different between our setups that would explain the include path problems. Are you able to try with the changes I made and let me know if you still have issues? If so, any additional details would be helpful.

dummyunit:
> It's not obvious to me what might be different between our setups that would explain the include path problems you saw. Are you able to try again with the changes I made and let me know if you still have issues? If so, any additional details would be helpful.

I have retested this issue again on latest commit from dev (c7576204) on three machines:
1. Win10; VS 2019 Professional with toolchains for VS 2019 (14.29.30037) and VS 2017 (14.16.27023); Clang 10.0.0, installed via official installer (not VS installer); WinSDK 10.0.19041.0.
2. Win 7 VM; VS 2019 BuildTools with toolchains for VS 2019 (14.29.30037) and VS 2017 (14.16.27023); Clang 12.0.0, installed via official installer (not VS installer); WinSDK 10.0.19041.0.
3. GitHub Runner; VS 2019 Enterprise with toolchains for VS 2019 (14.29.30037) and VS 2017 (14.16.27023); Clang 11.0.0, installed VS installer; WinSDK 10.0.19041.0.

In all cases builds with `USING_VS2019` (and with `USING_VS2015` on GitHub) fail and builds with `USING_VS2017` succeed.

Further investigations also revealed some problems in `USING_VS2017` builds.
When `clang-cl` is run by `FBuild` as part of the build it includes `vcruntime.h` from the VS 2017 toolchain (`C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.16.27023\include\vcruntime.h`), this is according to `procmon`. But if we run the same command manually (`-showcmds` and copy-pasting the command) it includes `vcruntime.h` from the VS 2019 toolchain (`C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30037\include\vcruntime.h`). This is surprising and concerning. The only reasonable explanation for this (that I can come up with) is different environment variables in this cases (`Settings` in `fbuild.bff` adds some VS paths to `PATH`).

But lets look on this issue from another angle:
* FASTBuild code includes `vcruntime.h` (transitively via `stdio.h` and `corecrt.h`).
* `vcruntime.h` exists in two places on the filesystem: `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.16.27023\include\` and `C:\Program Files (x86)\Microsoft Visual Studio\2019\Professional\VC\Tools\MSVC\14.29.30037\include\`.
* We pass neither of these paths via `-I` (or similar) options to `clang-cl`.

So according to all usual rules for include files the compiler wouldn't be able to find `vcruntime.h`.

Also how should `clang-cl` know which one of these two files it should include? (even if it somehow knows about their locations).

It seems there is some undocumented black magic in `clang-cl` that adds path to _some_ (possibly random) VC includes directory to its include search path. It think we shouldn't rely on it and specify the correct path explicitly.

As for why it works for you, I recommend you compile `Core-x64Clang-Debug` while running `procmon` and see which one of `vcruntime.h` files gets read.
When `USING_VS2019` build fails on my machine, `procmon` says that `clang-cl` tries to open `vcruntime.h` from the following locations before giving up:
* `C:\fastbuild\Code\vcruntime.h`
* `C:\Program Files\LLVM\lib\clang\10.0.0\include\vcruntime.h`
* `C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\vcruntime.h`
* `C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\um\vcruntime.h`
* `C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\shared\vcruntime.h`
* `C:\Program Files (x86)\Windows Kits\10\Include\10.0.19041.0\ucrt\vcruntime.h`  <- not a type, it tries this path 2 times
* `C:\fastbuild\Code\Core\Env\vcruntime.h`
* `C:\fastbuild\tmp\x64Clang-Debug\Core\vcruntime.h`

Maybe this list is somehow different on your machine, maybe there is a stray `vcruntime.h` in one of these places.

ffulin:
**> It seems there is some undocumented black magic in clang-cl** 

There absolutely is secret magic in clang-cl. Even though clang++.exe and clang-cl.exe are binary identical executables, invoking clang-cl.exe alters how includes are parsed and in at least one case clang.exe injects its own headers to completely replace what is normally included (intrinsics.h). I can understand why they did this, but it's a frustrating choice. As an example, in my experience, it's not possible to get clang++.exe to work the same as clang-cl.exe no matter what options are used.

Ultimately I still wasn't able to repro the issues you saw exactly as you saw it, but I used procmon as suggested and found the cause. The header was actually being found as follows:

|C:\p4\depot\Code\vcruntime.h|NAME NOT FOUND|
|C:\p4\depot\External\SDK\Clang\Windows\11.0.0\lib\clang\11.0.0\include\vcruntime.h	NAME NOT FOUND|
|C:\p4\depot\External\SDK\Windows\10\Include\10.0.19041.0\ucrt\vcruntime.h|NAME NOT FOUND|
|C:\p4\depot\External\SDK\Windows\10\Include\10.0.19041.0\um\vcruntime.h|NAME NOT FOUND|
|C:\p4\depot\External\SDK\Windows\10\Include\10.0.19041.0\shared\vcruntime.h|NAME NOT FOUND|
|C:\Program Files\Microsoft Visual Studio\2022\Preview\VC\Tools\MSVC\14.30.30401\include\vcruntime.h|SUCCESS|

So in my case clang-cl.exe decided that VS2022 should be used for some reason despite not being present anyway in any include path (system or otherwise). Very frustrating that arbitrary include paths are being checked by clang-cl.exe like this.

**> I think we shouldn't rely on it and specify the correct path explicitly.**

I agree. We definitely don't want arbitrary include paths being used as seems to be the case currently.

Now that I see what's going on, your fix makes absolute sense.

(I think we also need to add it to the "NonCL" flavors of Clang)

One thing that would be ideal is if we could prevent clang from compiling at all without this fix on a system where various versions of VS are installed that hide the problem. I experimented with some options (like -nostdinc and -nobuiltininc) but those just cause compilation to fail in unexpected ways, seemingly changing how clang behaves in non-obvious ways. If you have any ideas on this front, I'd love to hear them (though that's not a blocker for adopting this fix at all)

ffulin:
This change has now been integrated. Thanks for the fixes!

dummyunit:
> There absolutely is secret magic in clang-cl.

I dug around in LLVM sources and found this: https://github.com/llvm/llvm-project/blob/llvmorg-12.0.0/clang/lib/Driver/ToolChains/MSVC.cpp#L1268
This function seems to be responsible for adding system include paths to the include search list, including paths that are found by the means of this "magic" (which happens to be quite extensive - clang tries to use environment variables, registry keys and even some COM API to find VS toolchain installation).

> I experimented with some options (like -nostdinc and -nobuiltininc) but those just cause compilation to fail in unexpected ways

After looking at the code above this seems logical:
* `-nostdinc` disables addition of *all* system include paths, even paths that are added explicitly via `/imsvc` options.
* `-nobuiltininc` disables addition of path to internal clang headers. (the path is based on what `clang -print-search-dirs` displays in `libraries:` line).

> If you have any ideas on this front, I'd love to hear them

Again, after looking at the code above, it looks like we need the `-nostdlibinc` option - it will disable only the addition of paths found by the "magic". Note that `clang-cl` accepts it only like that: `/X` or `-X` (but in "default" compiler driver and internally it is called `-nostdlibinc`).

After I added `/X` to `.CommonCompilerOptions` in `.ToolChain_Clang_Windows` and `-nostdlibinc` to `.CommonCompilerOptions` in `.ToolChain_ClangNonCL_Windows` the build with `USING_VS2017` started to fail (w/o this PR), which means that path that was found by the "magic" wasn't added to the search list. And with this PR and `/X` the build succeeds. I didn't run the tests though, so maybe usage of this option will cause some problems with them.

PR #847, opened 2021-06-30
TestObject: Fix ordering of actions in some tests

dummyunit:
# Description:

* In ClangExplicitLanguageType test an `FBuildOptions` instance was modified after it was passed to `FBuild` constructor. As a result additional options that were forcing remote compilation were ignored.
* In ClangDependencyArgs test `FBuild::Initialize()` was called after a `Server` instance was created. `FBuild::Initialize()` temporarily places `SmallBlockAllocator` into single-threaded mode, but this is not safe as `Server` objects creates other threads that can try to allocate memory. A similar problem was fixed in the past: 7bd20da821 (#606)

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestObject.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestObject.cpp
@@ -512,14 +512,15 @@ void TestObject::ClangExplicitLanguageType() const
         // Init
         FBuildTestOptions options;
         options.m_ConfigFile = configFile;
-        FBuild fBuild( options );
-        TEST_ASSERT( fBuild.Initialize() );
 
         // Force remote
         options.m_AllowDistributed = true;
         options.m_NoLocalConsumptionOfRemoteJobs = true;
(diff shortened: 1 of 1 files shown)


ffulin:
Nice work. 

Regarding the thread safety issue,: The gain from the "single-threaded mode" was significant, but it's unfortunate that problems like this can creep back in.  How did you see the failure manifest? Was it triggering the safety asserts, breaking in asan configs or something else?

dummyunit:
> How did you see the failure manifest? Was it triggering the safety asserts, breaking in asan configs or something else?

Assert in `SmallBlockAllocator::Free()` was triggered several times on GitHub runners: https://github.com/dummyunit/fastbuild/runs/2947383724#step:6:650 https://github.com/dummyunit/fastbuild/runs/2947618779#step:6:650

ffulin:
Good to know that the assert was at least catching this situation.

Thanks for the fix.

PR #848, opened 2021-06-30
Add CI via GitHub Actions

dummyunit:
Since Travis CI became unusable and was removed, FASTBuild is in need of another CI solution as bugs that could be easily caught by CI are starting to creep in: #845, #846, #847.

# Description:

The first two commits are from PR #846 and #847, they are fixing issues that ware found during the setup of CI.
 
Third commit fixes test failure that occurred only on CI runners and I wasn't able to reproduce locally (so it is not clear if that change should go in if this PR is not accepted).

The last commit adds GitHub Action configs to perform checking of PRs and pushes. It also extends VS autodetection with install location of VS on GitHub Windows runners.

Currently the following jobs are defined:
* Linux (Build & Tests, GCC 10, Clang 12)
  Checks compilation with newest compilers (available out of the box), and runs tests (using `Tests` target).
* Linux (Build, GCC 7, Clang 9)
  Checks compilation with oldest compilers (available out of the box).
* Linux (ASan, GCC 10, Clang 12)
  Runs tests compiled with Clang in Address Sanitizer mode.
* Linux (MSan, GCC 10, Clang 12)
  Runs tests compiled with Clang in Memory Sanitizer mode.
* Linux (TSan, GCC 10, Clang 12)
  Runs tests compiled with Clang in Thread Sanitizer mode.
* OSX (Build & Tests)
  Checks compilation with Apple Clang 12 (the only version available out of the box), and runs tests (using `Tests` target).
* Windows (Build & Tests, VS 2019)
  Checks compilation with VS 2019 and Clang 11 and runs tests (using `Tests` target).
* Windows (Build, VS 2017)
  Checks compilation with VS 2017 and Clang 11.
* Windows (Build, VS 2015)
  Checks compilation with VS 2015 and Clang 11.
* Windows (Analyze, VS 2019)
  Builds `All-x64-Analyze` target with VS 2019 performing static analysis.

There are currently some problems that aren't solved and for which I added workarounds:
* ThreadSanitizer found a race in `TestDistributed.TestLocalRace` test ~~once~~twice: https://github.com/dummyunit/fastbuild/runs/2947618917#step:9:1209 https://github.com/dummyunit/fastbuild/runs/2964252136#step:9:1209 I didn't investigate further ~~as this happened only once so far~~.
   As a workaround I marked the corresponding job with `continue-on-error: true` which hopefully should keep overall Ci status green when that job fails.
* Test `TestTestTCPConnectionPool.TestConnectionCount` failed once in `x64ClangLinux-MSan` due to to timeout: https://github.com/dummyunit/fastbuild/runs/2953262398#step:8:320 I didn't investigate further as this happened only once so far.
   The workaround is the same as in the previous case.
* MSVC static analysis fails when WinSDK 10.0.19041.0 is used, the failure is caused by errors reported by analyzer in `WS2tcpip.h`: https://github.com/dummyunit/fastbuild/runs/2945765753#step:7:35
   ~~As a workaround the previous available WinSDK version is used in the corresponding job.~~ There is now a proper fix for this: a6e5b46c

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/.github/workflows/Linux.yml
+++ b/.github/workflows/Linux.yml
@@ -1,0 +1,100 @@
+name: Linux CI
+
+on: [push, pull_request]
+
+defaults:
+  run:
+    shell: bash
+    working-directory: Code
+
+jobs:
+  linux-ci:
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - cfg: Build & Tests
+            gcc: 10
+            clang: 12
+            os: ubuntu-20.04
+          - cfg: Build
+            gcc: 7
+            clang: 9
+            os: ubuntu-18.04
+          - cfg: ASan
+            gcc: 10
+            clang: 12
+            os: ubuntu-20.04
+          - cfg: MSan
+            gcc: 10
+            clang: 12
+            os: ubuntu-20.04
+            can-fail: true
+          - cfg: TSan
+            gcc: 10
+            clang: 12
+            os: ubuntu-20.04
+            can-fail: true
+
+    name: Linux (${{ matrix.cfg }}, GCC ${{ matrix.gcc }}, Clang ${{ matrix.clang }})
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.can-fail || false }}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: ./.github/actions/DownloadFBuild
+
+      - name: Configure fbuild.bff
+        env:
+          GCC: gcc-${{ matrix.gcc }}
+          GXX: g++-${{ matrix.gcc }}
+          CLANG: clang-${{ matrix.clang }}
+          CLANGXX: clang++-${{ matrix.clang }}
+        run: |
+          for x in ${GCC} ${GXX} $(${GCC} -print-prog-name=cc1) $(${GCC} -print-prog-name=cc1plus) ${CLANG} ${CLANGXX}; do
+            echo $x; which $x && readlink -f $(which $x)
+            echo
+          done
+          ${GCC} --version
+          ${CLANG} --version
+          # Inject "#define CI_BUILD" into root configs to activate CI logic.
+          sed -i -e "1i\\
+          #define CI_BUILD
+          " fbuild.bff Tools/FBuild/FBuildTest/Data/testcommon.bff
+          # Put full paths to GCC binaries into config files.
+          sed -i -e "
+          s:GCC_BINARY:$(which ${GCC}):
+          s:GXX_BINARY:$(which ${GXX}):
+          s:CC1_BINARY:$(${GCC} -print-prog-name=cc1):
+          s:CC1PLUS_BINARY:$(${GCC} -print-prog-name=cc1plus):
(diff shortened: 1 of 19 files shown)


.github/workflows/Linux.yml
```diff
@@ -0,0 +1,102 @@
...
+    name: Linux (${{ matrix.cfg }}, GCC ${{ matrix.gcc }}, Clang ${{ matrix.clang }})
+    runs-on: ${{ matrix.os }}
+    continue-on-error: ${{ matrix.can-fail || false}}
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: ./.github/actions/DownloadFBuild
+        with:
+          version: 1.03
```
ffulin, line 48: Should this use the current version? (v1.05)?

Would it be better to keep the version inside the DownloadFBuild action so that it can be updated in only one place when new versions are released?

External/SDK/Clang/Linux/Clang_CI.bff
```diff
@@ -0,0 +1,63 @@
+// Clang
```
ffulin, line 1: Seems pretty reasonable to bring these *_CI files back for Linux and OSX until they can be merged so they auto-detect like the Windows bffs do now.

External/SDK/VisualStudio/VS2017.bff
```diff
@@ -40,11 +41,21 @@
                 .VS2017_Version         = '14.16.27023'
             #else
                 //
-                // Failed
+                // Use Visual Studio 2019 Enterprise installation (used by GitHub Actions)
                 //
```
ffulin, line 45: This is the 2017.bff, but this refers to VS 2019 (the comment and the path)?

ffulin:
The work you have done here is very much appreciated.

External/SDK/VisualStudio/VS2017.bff
dummyunit: That is correct.
The latest VC toolchain from VS 2017 (14.16) can be installed as part of VS 2019, in that case it goes to `C:\Program Files (x86)\Microsoft Visual Studio\2019\EditionName\VC\Tools\MSVC`. This is the way it is installed on Windows runners for GitHub Actions - VS 2019 Enterprise is installed there and toolchains from VS 2017 and VS 2015 are installed as its separate components.

.github/workflows/Linux.yml
dummyunit: > Should this use the current version? (v1.05)?

I just copied the version from `.travis.yml`. Will try with 1.05.

> Would it be better to keep the version inside the DownloadFBuild action so that it can be updated in only one place when new versions are released?

I agree. It seems I was too busy playing with syntax. Added a default value to `version` in `DownloadFBuild/action.yml` and removed `with` sections in workflow files.

External/SDK/VisualStudio/VS2017.bff
ffulin: Thanks.

ffulin:
This change has now been integrated. Thanks again for all your work on this!

PR #850, opened 2021-07-03
Fix crash on wrong literal type in boolean evaluator

dummyunit:
# Description:

This fixes 2 problems:
* A crash found by BFFFuzzer in `Operand::Ensure()` on code that refers to a non-existing variable: `.A='' If (.A>.B) {}`
* Continuation of parsing after error about incorrect type of the RHS argument of operator `in`. For example this code printed "True": `.A=1 If (('x' in .A) || true) { Print("True") }`

Changes:
* All token validation steps are moved to `Operand` and `OperandOf` constructors. This provides a nice invariant: if `IsValid()` is false then there was an error, for which an error message was already emitted, and the parsing should be aborted.
* `EnsureValid()` is removed in favor of always checking `IsValid()` after constructing `OperandOf` (except in cases where the token validity is already checked beforehand).
* Added handling of non-variable tokens in the code reporting error 1050 for incorrect operand type.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/BFF/BFFBooleanExpParser.cpp
+++ b/Code/Tools/FBuild/FBuildCore/BFF/BFFBooleanExpParser.cpp
@@ -117,7 +117,7 @@ struct ArrayOfString
     typedef Array<AString> Type;
     typedef const Array<AString> & ReturnType;
     typedef Array<AString> & OutParamType;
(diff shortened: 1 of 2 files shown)


ffulin:
Thanks for the fix! The IsValid() vs EnsureValid() is a good improvement too.

This fix will be in the next release (v1.06)

PR #851, opened 2021-07-03
Update BFFFuzzer

dummyunit:
# Description:

* Reset `FBuild` state before parsing to prevent leaking data between runs.
  Leaking some of the state (environment, `BFFFileExists`) can affect the parsing results and make found crashes non reproducible as a result.
  Leaking `BFFUserFunctions`, on the other hand, results in bogus crashes because it holds pointers to deleted `BFFToken` objects, which were created for some previous input data.
* Allow specifying size of the hash table in `NodeGraph` and use much smaller one (2^8 vs 2^16) during fuzzing. This doubles the fuzzing speed as previously most of the time was spent allocating and initializing the huge (512 KiB) `m_NodeMap` array.
* Fix usage of `USING_CLANG_*` defines .bff files in parts related to sanitizer and fuzzer options. Previously `USING_CLANG_6` was used to enable use of flags intended for Clang >= 6, as a result flags intended for Clang 3.* were used with Clang 10 (`USING_CLANG_10`).
* Add some new keywords/operators/etc. to bff.dict.
* Don't disable output via `FBuildOptions` as setting `Tracing` callback is enough to prevent output. Add comment explaining why we want output from FBuild disabled.
* Update BFFFuzzer readme - update build instructions for libFuzzer and libc++, add section about removing slow test cases.

# Checklist:

The pull request:
- [x] **Is self-contained**


Code/Tools/FBuild/BFFFuzzer/Main.cpp
```diff
@@ -52,7 +43,8 @@ extern "C" int LLVMFuzzerTestOneInput( const uint8_t * data, size_t size )
     memcpy( str.Get(), data, size );
     str.Get()[ size ] = 0;
 
-    NodeGraph ng;
+    FBuild fbuild( env.options );
+    NodeGraph ng( 10 ); // Use smaller hash table to speed up fuzzing.
```
ffulin, line 47: I'm curious about the nature of the speedup gained by having a smaller hash table. Do you have any info you can share about that?

dummyunit: This is a great question. It prompted me to actually profile code in its current state (instead of going from what I remembered about profiling BFFFuzzer 2-3 years ago). As a result I found another bottleneck (which I actually added in this PR by recreating `FBuild`), after fixing it the speedup is now x4 instead of x2 (and x7 for an empty corpus, but that isn't a very useful case).

### The current version (dev branch)
Running `bfffuzzer` in "reproducible" mode (fixed number of iterations (100000), fixed seed for RNG, empty corpus) results in the following statistics:
```
$ ../tmp/x64ClangLinux-ASan/Tools/FBuild/BFFFuzzer/bfffuzzer -seed=123 -runs=100000 -print_final_stats=1
<<< snip >>>
Done 100000 runs in 55 second(s)
stat::number_of_executed_units: 100000
stat::average_exec_per_sec:     1818
```
If we run it under `perf`:
```
$ perf --call-graph dwarf ../tmp/x64ClangLinux-ASan/Tools/FBuild/BFFFuzzer/bfffuzzer -seed=123 -runs=20000
$ perf report
```
we see that 40,5% of all time is attributed to `LLVMFuzzerTestOneInput` function, but only 3,5% from these 40,5% are spent on actual parsing, while `NodeGraph` "eats" almost 35%:
```
- 40,53%  LLVMFuzzerTestOneInput
  + 18,75% ~NodeGraph (inlined)
  + 16,54% NodeGraph (inlined)
  + 3,68% BFFParser::ParseFromString
    0,58% BFFParser (inlined)
```

### First version of this PR
Running  `bfffuzzer` in "reproducible" mode yelds:
```
<<< snip >>>
Done 100000 runs in 39 second(s)
stat::number_of_executed_units: 100000
stat::average_exec_per_sec:     2564
```
So there is some speedup, but it is countered by a new problem which we see in `perf` report:
```
- 71,21%  LLVMFuzzerTestOneInput
  - 34,19% FBuild (inlined)
    + 21,23% FBuildStats (inlined)
    + 8,85% Function::Create
    + 2,78% FBuildOptions (inlined)
      0,74% FileIO::GetCurrentDir
  - 29,59% ~FBuild (inlined)
    + 23,72% ~FBuildStats (inlined)
    + 4,28% Function::Destroy (inlined)
    + 0,73% LightCache::ClearCachedFiles
  + 4,20% BFFParser::ParseFromString
  + 1,21% NodeGraph (inlined)
  + 0,56% ~BFFParser (inlined)
```
Now majority of time is spent on creating/destroying `FBuild`, mainly on `FBuildStats`.

### Second version of this PR
Running  `bfffuzzer` in "reproducible" mode yelds:
```
<<< snip >>>
Done 100000 runs in 8 second(s)
stat::number_of_executed_units: 100000
stat::average_exec_per_sec:     12500
```
So now its 7 faster (but it is on an empty corpus, on a populated one it "only" 4 time faster). Running under `pref` confirms that no majority of time is spent on parsing:
```
- 35,31% LLVMFuzzerTestOneInput
  + 18,47% BFFParser::ParseFromString
  + 6,94% NodeGraph (inlined)
  + 2,84% ~NodeGraph (inlined)
  + 2,55% ~BFFParser (inlined)
  + 1,45% BFFParser (inlined)
    1,26% ~BFFStackFrame (inlined)
    0,58% FBuildForFuzzing::ResetState
    0,51% Alloc
```

ffulin: The freeing of the Node hashmap is a single memory deallocation, so I'm a little surprised it's so expensive. I assume the cost comes from the sanitizer updating their shadow memory for what is a relatively large allocation?

For the new Reset() function, I have a small concern that this could break in the future as changes are made to FASTBuild. I think the code in question is rarely changes, so it's certainly ok for now, but I think this is something we'll need to keep an eye on.

ffulin:
These changes have been integrated and will be part of the next release (v1.09). Thanks as always!

PR #852, opened 2021-07-05
Allow to manually enable asserts in non-debug builds

dummyunit:
# Description:

* Some code assumed that `ASSERTS_ENABLED` is defined when `DEBUG` is defined and vice versa. This made it difficult to temporary enable asserts in non-debug builds by defining `ASSERTS_ENABLED` as such code started to produce warnings and/or errors.
* Always use `__builtin_trap()` to break into debugger on Linux.
   `__builtin_trap()` generates `ud2` instruction on x86. Both `ud2` and `int3` stop the program execution when the program is run under `gdb`, but execution of `ud2` generates `SIGILL` instead of `SIGTRAP` and the latter isn't intercepted by libFuzzer runtime. So with `int3` asserts just terminate BFFFuzzer instead of handling the crash (saving the input, displaying stack trace and restarting fuzzing).

The goal of these changes is to be able to check test corpus generated by BFFFuzzer for possible asserts that it might trigger, for now by manually defining `ASSERTS_ENABLED` and rebuilding BFFFuzzer.

The pull request:
- [ ] **Is self-contained** (There are two independent changes that could have been in different PRs, so I guess the PR is not self-contained. But they are thematically related and work towards the same goal so maybe it is OK to have them in one PR. I could split the PR if needed.)

### DIFF
--- a/Code/Core/Env/Assert.cpp
+++ b/Code/Core/Env/Assert.cpp
@@ -116,6 +116,9 @@ bool IsDebuggerAttached()
         }
         return true; // break execution
     #else
+        (void)message;
+        (void)file;
+        (void)line;
         return true; // break execution
     #endif
 }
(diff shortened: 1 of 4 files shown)


ffulin:
I'm not sure it will always be desirable to support compiling Release with asserts enabled, but with your changes is seems like there's not much work required right now so having that ability is nice.

Issue #398, opened 2018-02-27
Could I get fbuild binary for CentOS officially?


BrianMSon:
Could I get fbuild binary for CentOS officially?
I got errors on gcc 4.8.2.

_fbuild: /lib64/libc.so.6: version `GLIBC_2.15' not found (required by fbuild)_

_fbuild: /lib64/libc.so.6: version `GLIBC_2.14' not found (required by fbuild)_

_$ gcc --version_
_gcc (GCC) 4.8.2 20131017 (Red Hat 4.8.2-1)_

Thank you.

ffulin:
What version of CentOS are you using, and what version of glibc does it make available?

(For reference, this is the same issue described here: https://github.com/fastbuild/fastbuild/issues/225)

BrianMSon:
OS version : CentOS release 6.5 (Final)
Required glibc versions : 2.14 and 2.15

I cannot install external library because I'm not super user for this system.

ffulin:
I've made the necessary changes (https://github.com/fastbuild/fastbuild/commit/db78492682d8a9619208dc4c8880df5cfe6bb9ab) such that the next version will only require glibc v2.6. This change will be in the next release (v0.96). I'll leave this issue open until that version is released.

ffulin:
A version containing a fix for this issue has now been released (v0.96).

Issue #225, opened 2017-03-25
Use older glibc on Linux


Leandros:
FASTBuild requires a glibc 2.15 or higher, which is often not available for slightly older systems (as an example, Debian 7 only has 2.14, and was only released in 2013).
Likely related to #66.

ffulin:
I'm certainly open to supporting older kernel versions, but I've not seen a way to target older kernel versions from newer versions of Linux. It does seem that statically linking glibc might be a way to increase compatibility, but it's not something I've had a chance to try.

Leandros:
Since I've stumbled upon this again, one rather easy solution would be to statically link musl, instead of dynamically linking glibc (glibc has stopped supporting static linking).

ffulin:
I'm not super excited by the idea of depending on a third party library to avoid this problem. It seems (as per #66, the best fix might be to switch the official builds to an older Linux version. The question then becomes "how far back is reasonable to support/worth supporting?"

Leandros:
I wouldn't consider Musl to be a third-party library. Musl is the standard libc for numerous Linux distributions. It's one of the major contenders to glibc.

ffulin:
> I wouldn't consider Musl to be a third-party library.

Fair enough :) I guess "external" library might be a better way to put it.

ffulin:
Looking into this a little bit more, it seems we only really depend on anything meaningful in glibc 2.6, with the exception of these two functions:

 - memcpy
 - __fdelt_chk

I think both of those dependencies could be avoided with some effort (implementing or --wrap magic for the memcpy and undefining the symbols enabling the "fortify source" runtime checks).

Leandros:
You could get around linking glibc dynamically (or musl statically) by loading `memcpy` via `dlsym`. Since `dlsym` is ignoring all the embedded symbol versioning, you should be able to build on a linux with a newer glibc and still deploy to older glibc's.

dummyunit:
> memcpy
> __fdelt_chk
> I  think both of those dependencies could be avoided with some effort  (implementing or --wrap magic for the memcpy and undefining the symbols  enabling the "fortify source" runtime checks).

I have tried to remove these dependencies and came up with this:
* Dependency on `__fdelt_chk` can be removed by building with `-U_FORTIFY_SOURCE`.
* We can force compiler to link to the old `memcpy` implementation by injecting this line into all translation units: `__asm__(".symver memcpy,memcpy@GLIBC_2.2.5");`

[...]

PR #854, opened 2021-07-27
Avoid dependency on glibc 2.33 and 2.17 when building against them

dummyunit:
# Description:

Since the compatibility with old glibc versions was addressed previously (db78492682, #225, #398) it has regressed several times:
* In 0.97 fbuild started requiring glibc 2.10 due to usage of `accept4()`. That function first appeared in glibc 2.10 so this dependency is not easy to avoid. Also requiring glibc 2.10 shouldn't be a problem as no supported distribution is using glibc older than that.
* In 1.03 fbuild started requiring glibc 2.17 due to usage of `clock_gettime()`. Since that function was used well before 1.03, the dependency was most likely caused by building on a newer distribution than before.
* Very soon fbuild might start requiring glibc 2.33 due to usage of `stat()`, `lstat()` and `fstat()`, as the way they are implemented was changed in that version.

Fixing dependency on glibc 2.17 is easy - we use the `__asm__(".symver")` trick to link to the old implementation.
According to glibc 2.17 release notes (https://sourceware.org/legacy-ml/libc-announce/2012/msg00001.html) the new version of `clock_gettime()` was introduced to improve performance of singe-threaded programs, so we can continue to use the old version.

Fixing dependency on glibc 2.33 is harder as implementation of `stat()`, `lstat()`, etc. was changed significantly.
Previously they were implemented as inline wrappers directly in `sys/stat.h` and were calling internal implementation functions (like `__xstat()`). And executables were then linking to those internal functions.
In 2.33 this indirection was removed and now executables are linked to real `stat()`, `lstat()`, etc. in libc.so.6.
So to stay compatible with old glibc versions we have to implement those wrappers ourselves when we are building with glibc 2.33 or above.

# Checklist:

The pull request:
- [x] **Is self-contained**
   Adding tests for this is an interesting idea, we can even test this on CI building against the latest glibc. But I think this is better to be done in a different PR (and maybe discussed beforehand).

### DIFF
--- a/Code/Core/Env/glibc_compat.h
+++ b/Code/Core/Env/glibc_compat.h
@@ -8,8 +8,51 @@
 //------------------------------------------------------------------------------
 #undef _FORTIFY_SOURCE
 
-// Use older memcpy
+// Use older implementations of glibc functions
 //------------------------------------------------------------------------------
+__asm__( ".symver clock_gettime,clock_gettime@GLIBC_2.2.5" );
 __asm__( ".symver memcpy,memcpy@GLIBC_2.2.5" );
+
+//------------------------------------------------------------------------------
+
+// Use stat,lstat,fstat implementations from older versions of glibc for better Linux distro compatibility
+//------------------------------------------------------------------------------
+#include <sys/stat.h>
+
+// In glibc 2.33 implementation of various stat functions was changed - usage of proxy functions (__xstat, etc.) was removed.
+// As a result code built with glibc 2.33 headers links to stat@GLIBC_2.33 instead of __xstat@GLIBC_2.2.5.
+// But in this case we can't use the __asm__(".symver") trick to use old versions of those functions,
+// because there are no compatible versions - those proxy functions have a different signature.
+// Thus we must use a more involved hack to force usage of old versions to avoid depending on a new glibc version.
+
+#if defined( __GLIBC__ ) && ( __GLIBC__ * 1000 + __GLIBC_MINOR__ ) >= 2033
+
+    #ifdef __cplusplus
+        extern "C" {
+    #endif
+
+    int __xstat( int ver, const char * path, struct stat * buf );
+    __extern_inline int stat( const char * path, struct stat * buf )
+    {
+        return __xstat( 0 /*_STAT_VER*/, path, buf );
+    }
+
+    int __lxstat( int ver, const char * path, struct stat * buf );
+    __extern_inline int lstat( const char * path, struct stat * buf )
+    {
+        return __lxstat( 0 /*_STAT_VER*/, path, buf );
+    }
+
+    int __fxstat( int ver, int fildes, struct stat * buf );
+    __extern_inline int fstat( int fildes, struct stat * buf )
+    {
+        return __fxstat( 0 /*_STAT_VER*/, fildes, buf );
+    }
+
+    #ifdef __cplusplus
+        }
+    #endif
+
+#endif
 
 //------------------------------------------------------------------------------
--- a/Code/Tools/FBuild/FBuildTest/Data/TestUnity/ClangStaticAnalysis/fbuild.bff
+++ b/Code/Tools/FBuild/FBuildTest/Data/TestUnity/ClangStaticAnalysis/fbuild.bff
@@ -43,9 +43,8 @@ ObjectList( 'Compile' )
 ObjectList( 'Compile-InjectHeader' )
 {
     #if __WINDOWS__ // Using clang-cl on Windows
-        .CompilerOptions                = ' "/FICore/Env/glibc_compat.h" $CompilerOptions$'
-                                        + ' -Wno-reserved-id-macro'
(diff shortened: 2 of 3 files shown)


Code/Core/Env/glibc_compat_stat.h
```diff
@@ -0,0 +1,39 @@
...
+// because there are no compatible versions - those proxy functions have a different signature.
+// Thus we must use a more involved hack to force usage of old versions to avoid depending on a new glibc version.
+
+#if defined( __GLIBC__ ) && ( __GLIBC__ * 1000 + __GLIBC_MINOR__ ) >= 2033
+
+    extern "C" int __xstat ( int ver, const char * path, struct stat * buf );
+    extern "C" __inline int stat ( const char * path, struct stat * buf )
+    {
+        return __xstat( 0 /*_STAT_VER*/, path, buf );
+    }
```
dummyunit, line 23: This is more or less corresponds to what glibc does: https://sourceware.org/git/?p=glibc.git;a=blob;f=io/sys/stat.h;h=ce014d03a588ef5e2064b25594c69b9746c37976;hb=3de512be7ea6053255afed6154db9ee31d4e557a#l452

Code/Core/Env/glibc_compat_stat.h
```diff
@@ -0,0 +1,39 @@
+// glibc_compat_stat
```
dummyunit, line 1: Named this file in camel_case to match existing glibc_compat.h.
But maybe I should have named it PascalCase to match other header files? (since it is included directly)

dummyunit:
> _Very soon_ fbuild might start requiring glibc 2.33

I guess this entirely depends on when the OS used to build official Linux binaries will be updated, so it might be "not soon at all" in fact. But I think raising the issue in advance is beneficial.

ffulin:
Thanks for investigating this. The distro used to compile the distributed binaries was updated (new PC) and this almost certainly explains some of the changes you noted.

With respect to the new file you added to shim the fstat functions, would it be better to simply merge this with the existing file that we inject? My concern is that manually including it means it's quite a new use in the future could miss this include and re-introduce a problematic dependency.

dummyunit:
> With respect to the new file you added to shim the fstat functions, would it be better to simply merge this with the existing file that we inject? My concern is that manually including it means it's quite a new use in the future could miss this include and re-introduce a problematic dependency.

Originally I decided to put shim declarations in a separate file to avoid including `<sys/stat.h>` in all translation units. Looking at it again it doesn't look like a big deal (that include only add 250 lines after pre-processing):
```sh
$ echo '#include <sys/stat.h>' | gcc -x c++ -E - | sed -e '/^$/d; /^#/d;' | wc -l
250
```

So I moved everything to `glibc_compat.h` as you have suggested. Now I only need to fix test ClangStaticAnalysis_InjectHeader, as it (ab)uses `glibc_compat.h` for its testing needs and now fails on Windows (no `<sys/stat.h>` on Windows).

> re-introduce a problematic dependency.

Speaking of that, we can add a regression test for this. The check can be easily done in shell:
```sh
new_functions=$(objdump -T tmp/x64Linux-Release/Tools/FBuild/FBuild/fbuild | sed -e '/GLIBC_/! d; /GLIBC_2\.[0-9][. ]/ d; /GLIBC_2\.1[0-2][. ]/ d')
if [[ -n "${new_functions}" ]]; then
	echo "${new_functions}"
	false
fi
```
Do you want such test?  
In what form do you prefer it integrated? Easiest way it to just put this check in CI scripts, but I can try adding it as a `Test` target in .bff files.

dummyunit:
> Now I only need to fix test ClangStaticAnalysis_InjectHeader

Done.

But now there are:
* A timeout in TestConnectionCount test in MSan build:
```
 - Test 'TestConnectionCount'
-------- TEST ASSERTION FAILED --------
/home/runner/work/fastbuild/fastbuild/Code/Core/CoreTest/Tests/TestTCPConnectionPool.cpp(190): Assert: t.GetElapsed() < 30.0f
-----^^^ TEST ASSERTION FAILED ^^^-----
 - Test 'TestConnectionCount' *** FAILED ***
```
* A leak in TestStaleDynamicDeps detected on Windows by `MemTracker.h`:
```
 - Test 'TestStaleDynamicDeps'
 - Test 'TestStaleDynamicDeps' in 0.591s : *** FAILED (Memory Leaks)***
--- DumpAllocations ------------------------------------------------
D:\a\fastbuild\fastbuild\Code\Core\Env\Env.cpp(374): Id 5846 : 170 bytes @ 0x0000018c1d5ca5b0 (Mem: PATH=C:/Program.Files.(x86)/Mic)
--------------------------------------------------------------------
Total: 170 bytes in 1 allocs
--------------------------------------------------------------------
```
I am pretty sure changes in this MR are not responsible for these failures.

ffulin:
Thanks for the fixes! This has been integrated and will be in the next release v1.14.

PR #870, opened 2021-11-26
Compiler family detection improvements

jorgenpt:
# Description:

This makes various improvements to the compiler family detection, to make it more robust, consistent, and simpler. It also adds tests to the compiler family detection. The changes are:

- Strip the directory part of the compiler name before doing any comparisons, so we don't have to deal with which direction the slashes are in
- Strip executable-style extensions from the compiler name before comparing it, so that this behavior is consistent across all compilers (some compilers ignored .cmd, some didn't) -- this also adds support for ignoring `.sh` extensions for people who create wrapper scripts
- Have error 1501 use the input path in the error message, so that the slashes aren't backwards when seeing this error on Linux/Mac
- Add support for compiler versioning of Clang, to match GCC, since Linux distros will ship clang as e.g. `clang-12`
- Add tests to the compiler family detection

Generally speaking this should be at least as permissive as the old detection, except for `arbitrary-prefixed-clang-cl` no longer being detected as `CLANG_CL` -- I don't think there's every a use case for this (and that might be true for other tool chains, but I don't know them well enough to make that change), as clang-cl.exe is typically used from an MSVS distribution, which doesn't ever prefix it.

# Checklist:

The pull request:
- [X] **Is self-contained**


jorgenpt:
(Rebased to get new VS2019 compiler path detection.)

PR #871, opened 2021-11-26
Strip `-ftime-trace` for clang preprocessing

jorgenpt:
# Description:

`-ftime-trace` is used to generate timing information about building a
file, and while it would be ideal to have it include preprocessing *and*
the normal build, it relies on the output file name passed in via `-o`,
so in preprocessor operation it will always write to (and fight over)
`-.json` (since without `-o`, the output path is `-`).

# Checklist:

The pull request:
- [X] **Is self-contained**


ffulin:
This change looks sensible but needs to be accompanied by a test. If you are able to add an accompanying test it'll make it much easier to integrate.

jorgenpt:
(Rebased to get new VS2019 compiler path detection.)

I will look into writing tests!

jorgenpt:
@ffulin I don't see any tests that check the argument generation / commands executed by the build pipeline to use as a starting point -- do you have a suggestion for how to best invoke & intercept this code path in a test?

ffulin:
TestObject::ClangExplicitLanguageType might be a good reference and TestObject.cpp might be a good place to add this test also.

PR #880, opened 2022-01-03
Fix envvar leak

belkiss:
# Description:

Fix the memory leak of environment variable, GetEnvironmentString could be called with already allocated memory which we would lose track of

# Checklist:

The pull request:
- [X] **Is self-contained**


ffulin:
Thanks for the fix! This has been integrated and will be part of the v1.07 release.

I applied the logic modification I mentioned above to clarity the intended a logic a bit to make the reintroduction of a bug like this less likely in the future.

Issue #821, opened 2021-03-05
Share cache between git worktrees


MartyLake:
Hello!
Thanks for making fastbuild, it is really an amazing contribution to my development time everyday :)
I started to take a look into git worktrees, it is the interesting promise of having different branches checked-out at the same time on your computer.

I wish that fastbuild would be able to share cache between git worktrees.

I made my include paths relative, and turned on the fbuild compiler setting .UseRelativePaths_Experimental. This way, the command line hash is the same accross worktrees.

If I try to hash the cpp file (with md5sum.exe for example) on the two worktrees, I have the same results, but fastbuild come to a different source hash. I have compared the two pre-compiled files, and it seems that cl.exe adds "#line" directives with full path when passing "/E".

I have seen that there is a cl.exe preprocessor "/PE" option to remove those "#line" directives, but I wonder if that would interefere with fastbuild. Also, I wonder if making those "#line" directives relative somehow would solve this issue, but I have the intuition that it may have to be done by fastbuild.

Best,

ffulin:
It would be ideal for all caching to be path-agnostic. This would allow caching between paths.

Clang exposes some command line options that allow re-mapping include paths, including fixing up debug info. FASTBuild has two experimental options on the Compiler() that are steps towards fully supporting that:
  - .UseRelativePaths_Experimental
  - .SourceMapping_Experimental
The future for Clang looks promising WRT path-independent caching.

The MCVS compiler on the other hand, at least the last time I checked, embedded full paths into compiled output, even when all paths on the command line and all #line directives were relative. The concern here is that debug info may be incorrect. Some experimentation is needed to see how much of a problem this is for debugging. It's possible that this still works well enough, or that even if it doesn't work seamlessly could be an opt-in choice. Some further experimentation is needed.

MartyLake:
My knowledge in the inner workings of fastbuild are pretty limited right now, but I would be happy to help if I can :)

aganea:
@MartyLake if you apply the following "receipe" with clang-cl, then you'll have workspace-independent .OBJ files: https://blog.llvm.org/2019/11/deterministic-builds-with-clang-and-lld.html

MartyLake:
Interesting, thank you very much @aganea ,that's one more good reason to make our build system compatible with clang-cl !

PR #885, opened 2022-01-10
Allow LightCache to use relative paths

ben-craig:
# Description:

When both .UseLightCache_Experimental and .UseRelativePaths_Experimental are enabled on the compiler, LightCache will use relative paths to compute cache hashes, rather than absolute paths.  The file contents and compiler command line are still hashed, just as they were before.  This means that users should provide relative paths on the compiler command line if they want caching to work across machines / worktrees.

This enables (or at least improves) the ability for LightCache to cache results across machines with different path setups, as well as caching across git worktrees [#821].  Files that confuse LightCache because of recursive includes and conditional macros will still use the old fallback, and are unlikely to hit the cache.

# Checklist:

The pull request:
- [x] **Is self-contained**
- [TBD] **Compiles on Windows, OSX and Linux**

This relies on existing (minimal) documentation in compiler.html and cache.html.  This feature skews to Windows since this is a LightCache feature.


ben-craig:
Commentary on debugging experience (since that came up in the linked issue):

Let's consider 4 cases...
1. No fancy acceleration
2. Distributed build
3. Cached build
4. Distributed and cached build

In case 1, when you hit an assert or crash or whatever that is directly in a .cpp file and view the file in Visual Studio, the debugger takes you to the correct file without any extra effort.

In case 2, the debugger is likely going to require you to do some work to get to the correct file, because the TU was built on a different machine in a temporary path.  You can usually coax the debugger into using the "real" copy of the file, but it's annoying.

Prior to this change, case 1 and case 3 acted the same.  After this change, case 3 may cause the debugger to open the wrong copy of the file.  The "wrong" copy would be in another worktree (if that worktree still exists).  Chances are good that the contents are still useful, but it's not guaranteed.  If the user then modifies this wrong copy and tries to rebuild, then the user is likely to get confused.

In case 4, all the confusing and bad things could happen.

Since I'm posting the PR, I clearly believe that these issues aren't deal breakers.  They are all also opt-in.

ffulin:
As part of recent LightCache work, support for UseRelativePaths_Experimental was added here (https://github.com/fastbuild/fastbuild/commit/a7c1c15d053953f1aa49eb712c2dc41ee172f07a)

This change is similar to yours and should give you the desired behavior (caching between different directories)

This change will be in the next release (v1.19). I'll keep this PR open until then.

ffulin:
v1.19 has now been released and contains changes to allow the LightCache to support relative paths.

PR #890, opened 2022-02-02
Unity supports alias

Domain:
# Description:

CompilerInputUnity supports alias

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectListNode.cpp
@@ -9,6 +9,7 @@
 #include "Tools/FBuild/FBuildCore/BFF/Functions/FunctionObjectList.h"
 #include "Tools/FBuild/FBuildCore/FBuild.h"
 #include "Tools/FBuild/FBuildCore/FLog.h"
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the PR. Unfortunately, this needs to be fixed in a slightly different way. Aliases can be recursive, which with this change would result in silent failures/missing dependencies (or an assert in debug builds).

The code you modified is a bit of a remnant of the old/manual way of gathering dependencies for nodes, which involves manual handling of aliases. This code could be changes to handle aliases recursively, but there is a new and preferred way to handle that which eliminates the need for nodes to deal with aliases at all, and that is to use property-level metadata. Unfortunately, this exact piece of metadata to handle this case (non-files) doesn't exist right now so the core fix would be to add that and simplify this code.

It's worth noting that this bug exists in other properties/nodes, so that deeper fix (adding new metadata) could be applied to several places, simplifying them as well as making them work better.

I'll leave this PR open for now to continue tracking this issue while I continue to think about the best way to add this metadata.

ffulin:
I'd still like to address this problem. I'm ok with keeping this PR open as a reminder. Alternatively we can raise an issue to track it.

PR #896, opened 2022-02-11
Add support for arrays as unary expressions

jorgenpt:
# Description:

Add support for `If( .Array )` to check if an array has any elements.
Useful for constructs like

```
Alias("Foo") { .Targets = .SomeArrayThatMightBeEmpty }
```

As `.Targets` cannot be empty for an `Alias` -- now you can do:

```
If(.SomeArrayThatMightBeEmpty) {
    Alias("Foo") { .Targets = .SomeArrayThatMightBeEmpty }
}
```

 - (For changes or new functionality) Detail the motivation/use-case behind the change and any context required to understand it.
 - (For fixes) Describe the bug/problem being fixed.
 - (If appropriate) Link to any associated issues.

# Checklist:

The pull request:
- [X] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/if.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/if.html
@@ -36,6 +36,7 @@ The following types of expressions are supported:
   <li>Boolean Values</li>
   <li>Comparisons</li>
   <li>Presence in ArrayOfStrings</li>
+  <li>Array is non-empty</li>
   <li>Compound boolean expressions of above three types</li>
 </ul>
 <b>NOTE:</b> Literal of ArrayOfStrings is not supported so far. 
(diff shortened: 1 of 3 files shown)


ffulin:
Thanks for the PR. The implementation you've provided seems sensible/correct at a glance, but I'm concerned about adding implicit conversions such as this. I'd prefer to err on the side of more explicit checks, making the syntax stricter where possible to avoid ambiguity and user-error.

Do you have an example you can provide of your use-case for this that might help guide my thoughts around this?

jorgenpt:
@ffulin Pretty much the case I listed in the original ticket. Slightly more concretely, this is what we have:

```
ForEach(.Toolchain in .SupportedToolchains) {
    Using(.Toolchain)
    .UnrealDependenciesToolTargets = {}


    If(.SomeFlagFromToolchain) {
        ^UnrealDependenciesToolTargets + { 'Some value', 'Other value' }
    }

    If(.OtherFlagFromToolchain) {
        ^UnrealDependenciesToolTargets + { 'Third value' }
    }

    // etc

    If(.UnrealDependenciesToolTargets) {
        Alias('UnrealDependencies-$ToolchainName$-Tools') {
            .Targets = .UnrealDependenciesToolTargets
        }
    }
}
```

Xorcerer:
If the implicit conversion is not welcomed, can we do```.Array == {}```?

PR #908, opened 2022-05-17
Fixed compatibility with VS 2022 (17.2.32505.173). The IDE requires v…

Xorcerer:
# Description:
VS 2022 (17.2.32505.173) required the full version, e.g. `17.2.32505.173` in the `.sln` file instead of a single number, e.g. `16`.

How to reproduce:
* Update the VS 2022 to the latest version `17.2.32505.173`
* Generate a .sln file with fastbuild OR create a .sln file with IDE.
* (If it was generated by VS) change the content of the file by setting the version string to a single number e.g.`VisualStudioVersion = 17`
* open the file with VS 2022, please EXPECT an error from the IDE 
image

*Note*: `16.0.28701.123` in the changes came from https://docs.microsoft.com/en-us/visualstudio/extensibility/internals/solution-dot-sln-file?view=vs-2022, feel free to change it to any validate one.
* Updated the content of VS2022.bff, using the full version for `VS_SolutionVersion`.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/External/SDK/VisualStudio/VS2022.bff
+++ b/External/SDK/VisualStudio/VS2022.bff
@@ -173,7 +173,7 @@ Compiler( 'Compiler-VS2022-x64' )
 [
     .VS_Version                     = .VS2022_Version
     .VS_Version_HumanReadable       = '2022'
-    .VS_SolutionVersion             = '16'
+    .VS_SolutionVersion             = '16.0.28701.123'
     .MSC_VER                        = .VS2022_MSC_VER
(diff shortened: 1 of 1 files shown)


Xorcerer:
Weird, I don't think the one-line change of `VS2022.bff` impacted the Linux build.

ffulin:
Thanks for the fix. I have reproduced this with the latest VS2022 install. Strange that they would change this behavior in such a way.

The Linux failure is unrelated to your changes - it appears to be a rare bug that i will investigate separately.

PR #914, opened 2022-06-21
Adds `.Environment` option to the `Exec` function

ZeroErrors:
# Description:

Adds `.Environment` option to the `Exec` function. Its implemented the same as how its implemented on the `Compiler`, `DLL`, `Executable`, `Library` and `Test` functions.

Includes a unit test that provides environment variables via this option to an executable that dumps the values via stdout and then validates that the values returned are the overridden values.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Data/TestExec/exec.bff
+++ b/Code/Tools/FBuild/FBuildTest/Data/TestExec/exec.bff
@@ -165,4 +165,3 @@ Exec( "ExecCommandTest_OneInput_WrongOutput_ExpectFail" )
     .ExecReturnCode = 0
     .ExecUseStdOutAsOutput = false
 }
-
--- a/Code/Tools/FBuild/FBuildTest/Data/TestExec/execenv.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Data/TestExec/execenv.cpp
@@ -1,0 +1,40 @@
+//
+// An simple executable to test the environment variable FB_ENV_TEST_VALUE is set
+//
+#if defined( __WINDOWS__ )
+    #include <Windows.h>
+#elif defined( __LINUX__ ) || defined( __APPLE__ )
+    #include <stdlib.h>
+#else
+    #error Unknown platform
+#endif
+
+#include <string.h>
+
+int main(int, char *[], char *[])
+{
+    #if defined( __WINDOWS__ )
+        char envVar[15];
+        DWORD ret = GetEnvironmentVariable( "FB_ENV_TEST_VALUE", envVar, 15 );
+
+        // variable does not exist or is longer than expected
+        if ( ret == 0 || ret > 9 )
+        {
+            return 1;
+        }
+    #elif defined( __LINUX__ ) || defined( __APPLE__ )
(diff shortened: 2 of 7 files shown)


Code/Tools/FBuild/FBuildTest/Data/TestExec/execenv.cpp
```diff
@@ -0,0 +1,30 @@
+//
+// An simple executable to run
+//
+#include <iostream>
+#include <fstream>
+#include <string>
```
ffulin, line 6: FASTBuild doesn't use C++ standard libraries at all. It uses C standard libraries only.

This will make this test significantly slower to compile, but also introduces problems with C++ runtime libs (which even when statically linked have various issues on all platforms)

I think this test could be simplified to check that retrieving the env vars are the exact ones we know the test sets and returning 0 or 1 as a return value. This would avoid the need for fstream, iostream and string (removing all c++ standard library use)

Core/Env/Env.cpp has environment variable access code that is free of C++ standard library use that could be used as a reference.

Code/Tools/FBuild/FBuildTest/Data/TestExec/exec.bff
```diff
@@ -166,3 +195,21 @@ Exec( "ExecCommandTest_OneInput_WrongOutput_ExpectFail" )
...
+// Test with a no input file and using stdOut as output
+// In this case:
+// - The command will generate "Env_StdOut.txt.out" containing
+//   the value of "FB_ENV_TEST_VALUE" and "FB_ENV_TEST_EMPTY"
+//   environment variables.
+// - Return code will be 2 because 2 argument is passed in
+Exec( "ExecEnvCommandTest_StdOut" )
+{
+    .ExecExecutable = .EnvHelperExecutableName
+    .ExecInput = '$OutPath$\dummy_file_does_not_exist'
```
ffulin, line 208: ExecInput is optional and doesn't need to be set to a fake file.

Code/Tools/FBuild/FBuildTest/Tests/TestExec.cpp
```diff
@@ -299,6 +335,45 @@ void TestExec::Build_ExecCommand_ExpectedFailures() const
...
+    f.Close();
+
+    // Split lines
+    Array< AString > lines;
+    buffer.Replace( '\r', '\n' );
+    buffer.Tokenize( lines, '\n' ); // Will discard empty lines
+
+    TEST_ASSERT( lines.GetSize() == 2 );
+    TEST_ASSERT( lines[0] == "FB_ENV_TEST_VALUE=FASTbuild" );
+    TEST_ASSERT( lines[1] == "PATH=" );
```
ffulin, line 374: I think the two new test functions could be collapsed together and most of the logic in this function removed if we based the test passing entirely on the return code as I mention above.

ffulin (approved):

ffulin:
Thanks for the improvement! This change has been integrated and will be in the next release (v1.09)

(I've made some small followup tweaks to the test in https://github.com/fastbuild/fastbuild/commit/51dfa6ab663957db5c0c2f4eb88b3b9a7eb3cb20)

PR #915, opened 2022-06-29
Fix warnings when building in Clang 14 on Linux

dummyunit:
# Description:

Clang 14 started to emit `-Wunused-value` for the left side of the comma operator even in constexpr/template contexts: https://godbolt.org/z/aP6r313v3
This change in behavior breaks the build via this warning:
```
./Core/Containers/Forward.h:11:84: error: left operand of comma operator has no effect [-Werror,-Wunused-value]
template < class T, class U > struct ForwardType< T &, U >      { static_assert( ( sizeof( T ), false ), "can not forward rvalue as lvalue" ); };
```

I also suspect that `#pragma`s disabling `-Wcomma` around this code might not be needed after this change. I can't check that as the commit in which they were added doesn't mention which version of Clang produced `-Wcomma` there.

# Checklist:

The pull request:
- [x] **Is self-contained**
- [ ] **Has accompanying tests** N/A

### DIFF
--- a/Code/Core/Containers/Forward.h
+++ b/Code/Core/Containers/Forward.h
@@ -8,7 +8,7 @@ template < class T, class U > struct ForwardType< T &, U & >    { using type = T
     __pragma( clang diagnostic push )
     __pragma( clang diagnostic ignored "-Wcomma" )
 #endif
-template < class T, class U > struct ForwardType< T &, U >      { static_assert( ( sizeof( T ), false ), "can not forward rvalue as lvalue" ); };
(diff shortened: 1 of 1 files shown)


ffulin:
I'm not certain exactly which versions of Clang emit the -Wcomma warning that is being suppressed, but at I was able to verify that with Clang 11 on Windows the warning is emitted and indeed your cast to void eliminates the need to disable the warning so I think it can be safely removed after your changes are integrated (which I will do)

ffulin:
I've integrated the fix - thanks!

I've also removed the -Wcomma disabling pragmas in a followup here: https://github.com/fastbuild/fastbuild/commit/f6dc293a365c70097dd1d9db435888847ea39aab

PR #916, opened 2022-06-30
Support caching of .gcno files

dummyunit:
# Description:

GCC's instrumentation for code coverage (also known as gcov coverage) relies on `.gcno` files generated alongside `.o` files during compilation.
Previously FASTBuild didn't store these files in the cache, this effectively made caching unusable for such build configurations as object files retrieved from cache would miss corresponding `.gcno` file (or that file could be stale, which probably is even worse).

To solve this problem we now store `.gcno` files in the cache, similarly to `.nativecodeanalysis.xml` files that are produced by MSVC.

Also distribution is now disabled for object files that use options enabling generation of `.gcno` files. Distributed compilation in that case produces barely usable results:
1. `.gcno` files contain incorrect build root path (working directory on the worker, instead of the local working directory).
2. `.gcda` files (which are created at runtime by instrumented code) are created in wrong directories, namely in the directory where object files were stored on the worker.

While these issues are not critical (`.gcda` files can be moved manually to correct places, build root path stored in `.gcno` files can be overridden by `gcov` options) enabling distribution would also require bumping the version of distribution protocol, and it is not worth it for a barely usable results.

# Checklist:

The pull request:
- [x] **Is self-contained**
    I have added a test that verifies that `.gcno` file is created when object file is retrieved from the cache.
    I have also added a similar test for `.nativecodeanalysis.xml` files, it probably duplicates some part of `PrecompiledHeaderCacheAnalyze_MSVC` test, but it was too easy to generalize such tests. I hope that is OK.
    Added a mention that distribution is disabled for GCC/Clang when code coverage options (`-ftest-coverage` or `--coverage`) are used.

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/features/distribution.html
+++ b/Code/Tools/FBuild/Documentation/docs/features/distribution.html
@@ -51,7 +51,7 @@ settings, it will automatically build the impacted object(s) locally. A build ca
 <p>
 <b>GCC/SNC/Clang</b>
 <ul>
-  <li>No restrictions.</li>
+  <li>-ftest-coverage (and --coverage) cannot be distributed: .gcno files would contain wrong paths, .gcda files would be created in a wrong place.</li>
 </ul>
 <b>MSVC</b>
 <ul>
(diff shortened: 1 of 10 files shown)


ffulin:
Thanks for the improvements! These have been integrated and will be in the next release (v1.08)

I made some small follow up changes (https://github.com/fastbuild/fastbuild/commit/d3933b01809abcff65ac4a109c4ed2f209653617), the most notable of which are:
a) Incremented cache version. This is so that a new fbuild.exe doesn't crash if it retrieves a cached item built by an older fbuild.exe that did not store the additional file.
b) Removed code related to distributed compilation since distributed compilation is disabled when using gcov.

Let me know if any of that doesn't look right.

PR #919, opened 2022-07-31
Avoid dependency on glibc 2.34 when building against it

dummyunit:
# Description:

This PR continues to proactively fix issues that prevent building FASTBuild with new glibc versions and have it run on system with older glibc installed. Previous such PR is #854 

Glibc 2.34 broke forward compatibility again in two entirely new ways.

The first issue is that in glibc 2.34 the way in which global initialization and destruction functions are called was changed:
https://sourceware.org/git/?p=glibc.git;a=commitdiff;h=035c012e32c11e84d64905efaf55e74f704d3668

Previously each executable contained functions `__libc_csu_init` and `__libc_csu_fini` that iterated through all global initialization and destruction functions and the program entry point (`_start`) passed their addresses to the main startup code (`__libc_start_main`) implemented in `libc.so`.

In glibc 2.34 `__libc_csu_init` and `__libc_csu_fini` are removed, `_start` passes null pointers to `__libc_start_main` and `__libc_start_main` finds and executes global initialization and destruction functions by itself.
As this new behavior of `_start` is incompatible with `__libc_start_main` present in old `libc.so`, `__libc_start_main` got a new symbol version and now programs built with glibc 2.34 link to `__libc_start_main@@GLIBC_2.34`.

To be able to continue to use any glibc to build executables that will run on older systems we undo all of that:
1. We implement our own `__libc_csu_init` and `__libc_csu_fini` when building with glibc >= 2.34.
2. We hook calls `__libc_start_main` by using `--wrap=` linker option.
3. In our hook for `__libc_start_main` we call the real old `__libc_start_main`.
4. If we are building with glibc >= 2.34 we pass addresses of our `__libc_csu_`* implementations to `__libc_start_main`.

The second issue is that in glibc 2.34 libpthread, libdl and librt were merged into libc. Unversioned symlinks for these shared libraries are no longer installed and instead dummy static libraries are provided.
This means that now we can't link with these libraries in a usual way (`-ldl`, etc.), if we try we will instead link with dummy static
libraries, and as a result our binary won't load on system with older glibc due to unresolved symbols.

To ensure that we are linked to all needed libraries we now link directly to versioned shared library files.

Also many functions got new symbol versions, so we add corresponding `__asm__(".symver")` lines to `glibc_compat.h`.

# Checklist:

The pull request:
- [x] **Is self-contained**
- [ ] **Has accompanying tests** Discussion about tests is here: https://github.com/fastbuild/fastbuild/pull/854#issuecomment-991695813

### DIFF
--- a/Code/Core/CoreTest/CoreTest.bff
+++ b/Code/Core/CoreTest/CoreTest.bff
@@ -69,6 +69,7 @@
             #endif
             #if __LINUX__
                 .LinkerOptions                  + ' -pthread -lrt'
+                                                + ' -Wl,--wrap=__libc_start_main'
             #endif
         }
         Alias( '$ProjectName$-$Platform$-$BuildConfigName$' ) { .Targets = '$ProjectName$-Exe-$Platform$-$BuildConfigName$' }
--- a/Code/Core/Env/glibc_compat.h
+++ b/Code/Core/Env/glibc_compat.h
@@ -8,8 +8,26 @@
 //------------------------------------------------------------------------------
 #undef _FORTIFY_SOURCE
 
-// Use older memcpy
+// Use older implementations of glibc functions
 //------------------------------------------------------------------------------
+__asm__( ".symver dlclose,dlclose@GLIBC_2.2.5" );
+__asm__( ".symver dlerror,dlerror@GLIBC_2.2.5" );
+__asm__( ".symver dlopen,dlopen@GLIBC_2.2.5" );
+__asm__( ".symver dlsym,dlsym@GLIBC_2.2.5" );
 __asm__( ".symver memcpy,memcpy@GLIBC_2.2.5" );
+__asm__( ".symver pthread_attr_setstacksize,pthread_attr_setstacksize@GLIBC_2.2.5" );
+__asm__( ".symver pthread_create,pthread_create@GLIBC_2.2.5" );
+__asm__( ".symver pthread_detach,pthread_detach@GLIBC_2.2.5" );
+__asm__( ".symver pthread_join,pthread_join@GLIBC_2.2.5" );
+__asm__( ".symver pthread_mutexattr_init,pthread_mutexattr_init@GLIBC_2.2.5" );
+__asm__( ".symver pthread_mutexattr_settype,pthread_mutexattr_settype@GLIBC_2.2.5" );
+__asm__( ".symver pthread_timedjoin_np,pthread_timedjoin_np@GLIBC_2.3.3" );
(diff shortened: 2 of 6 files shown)


ffulin:
I've finally gotten around to looking at this issue (and the related https://github.com/fastbuild/fastbuild/pull/854) and it seems like perhaps it's worth talking about an alternative approach for maintaining compatibility as it seems you've spent a lot of time on this and because I'm not being thorough enough about which version binaries are built on its subject to continual breakage.

One option might be to switch the canonical shipped builds to always be built on a specific distro which is pinned to a specific version of glibc. We could chose a reasonable version to support and perhaps add a verification build step to ensure the requirement doesn't accidentally change.

The only downside of that is that anyone building their own version from source will need to do so in an environment that has a glibc supported by any machine they wish to deploy on, but that doesn't seem unreasonable to me.

What do you think about this?

I'm also open to other ideas.

ffulin:
I've *finally* been able to get to looking at this properly, understanding your changes, and I've been thinking about the maintenance of glibc compatibility. So far, i've come up with the following strategy:

**Use a distro with no later than glibc 2.31 for binary builds**
Using a host with glibc avoids issues with 2.33 and 2.34. Ubuntu 20.04 LTS can be used for now.
I've integrated your [other PR](https://github.com/fastbuild/fastbuild/pull/854) which means we can likely bump this to 2.33 in the future.

**Choose glibc 2.10 as a minimum target**
There are now some dependencies on accept4 (2.10) and utimensat (2.6).
Reliance on newer things has been addressed by manually integrating the symver parts of this PR and your [other PR](https://github.com/fastbuild/fastbuild/pull/854) in its entirety.
glibc 2.10 is ~15 years old so this still provides compatibility quite a way back (anything older than that is likely end of life already)
We will certainly be able to increase this in the future, and may also be able to remove some .symver entries that pushe some symbols back earlier than 2.10 if newer (but still <= 2.10 versions exist)

**Automate detection of breakages**
I've created a helper script (https://github.com/fastbuild/fastbuild/blob/dev/Scripts/CheckGLIBCCompat.py) which detects reliance on unexpected glibc versions, allowing us to list explicitly those that are supported.
It should be possible to add this to the Linux CI.

----

The next version (v1.14 should regain compatibility with glibc 2.10. I'll keep this issue open until that version is released.

I'd be curious to hear your thoughts.

ffulin:
v1.14 has now been released.
In addition to the above, this also includes:
 - removal of dependency on glibcxx
 - automated testing of compatibility in the CI by running the CheckGLIBCCompat.py script

PR #922, opened 2022-08-02
Remove extra files leftover after v1.07 sync

dummyunit:
# Description:

Presence of these files breaks build (from main branch) as `UnitTest.h` defines `TEST_ASSERT`, etc. which are now defined in `TestGroup.h`.

# Checklist:

The pull request:
- [ ] **Is created against the Dev branch** This PR is deliberately made against main branch as it fixes issue specific to the main branch.
- [x] **Is self-contained**
- [ ] **Compiles on Windows, OSX and Linux** I have tested on Linux only (no CI in the main branch).
- [ ] **Has accompanying tests** N/A
- [ ] **Passes existing tests** I have tested on Linux only (no CI in the main branch).


ffulin:
Thanks! I'll update the release tag to include this change (the fastbuild.org zip already contains the correct files and doesn't need to be regenerated)

PR #923, opened 2022-08-03
get limit from system configuration

Domain:
# Description:

get arg limit from system configuration

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Helpers/Args.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/Args.cpp
@@ -14,6 +14,8 @@
 // system
 #if defined( __OSX__ )
     #include <limits.h>
+#elif defined( __LINUX__ )
+    #include <unistd.h>
 #endif
 
 // CONSTRUCTOR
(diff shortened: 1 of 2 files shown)


ffulin:
I remember now why I didn't implement it this way and there is a code comment about it being problematic.

It seems that _SC_ARG_MAX will return the maximum size of space available for both arguments and environment. With that in mind, since there is always some amount of environment data, it looks as though attempting to use args upto the value returned by _SC_ARG_MAX will start to fail at some relatively arbitrary point depending on the complexity of the environment.

This leaves us with a bit of a dilema: For things that don't support response args, this change is better (as it'll work more often) but for things that do support response files, this change is worse as at some point things will fail when they would not have before.

The only idea I have to make this reliable is to calculate the "available space" from this actual limit by subtracting the environment length, but it seems some reverse engineering of how that is allocated would be needed and it's hard to be confident that would be reliable.

ffulin:
After giving this some more thought, I think even though using _SC_ARG_MAX alone comes with some new issues, it's probably better overall than the existing hard-coded value.

I ran an experiment, launching gcc with args that are too long and it fails with an error that makes the failure somewhat obviously command line length related, so i think the edge case where the args approach _SC_ARG_MAX but the full allocation can't be used (due to the size of the env block for example), will hopefully be somewhat obvious.

This change also is a step towards any more complex logic we might add in the future to try and handle the environment block size complexities.

ffulin:
Thanks for the PR - this change will be in the next release (v1.11)

PR #933, opened 2022-09-19
Fix NodeGraph::CleanPath() for paths ending with \. or /.

dummyunit:
# Description:

Previously, in paths ending with `\.` or `/.` the trailing dot was not removed. This made setting properties to the working directory unintuitive.

For example setting `.ProjectBasePath` to `'.'` didn't work, strings in `VSProjectGenerator::m_BasePaths` had a dot at the end, `BeginsWithI()` check against absolute paths to project files never succeeded and all files were dumped into the project without directory structure. Of course setting `.ProjectBasePath` to `'./'` did work, but that is not very intuitive. At least I had to dig through the sources to find that workaround.

# Checklist:

The pull request:
- [x] **Is self-contained**


ffulin:
Thanks for the fix. This change has been integrated and will be in the next release (v1.09)

Issue #200, opened 2017-01-20
Folder permissions to wild


Leandros:
The created folders by FASTBuild are created with 777 (rwx:rwx:rwx), which is far from ideal.
Folders should be created with 755 (rw-:r-x:r-x) permissions. While non-executable files should have 644 (rw-:r--:r--) and executable files 755 (rwx:r-x:r-x).

Tested on FASTBuild v0.92 on macOS 10.11.5.

ffulin:
There are also similar issues on Linux.  Here is what I think is the definitive list of cases that need to be checked:

```C++
/*static*/ bool FileIO::FileCopy( const char * srcFileName, const char * dstFileName,
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(180): int dest = open( dstFileName, O_WRONLY | O_CREAT | O_TRUNC, 0644 ); // TODO:LINUX Check args for FileCopy dst
  
/*static*/ bool FileIO::DirectoryCreate( const AString & path )  
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(410): mode_t mode = S_IRWXU | S_IRWXG | S_IRWXO; // TODO:LINUX TODO:MAC Check these permissions
  
/*static*/ bool FileIO::SetReadOnly( const char * fileName, bool readOnly )    
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(627): const bool currentlyReadOnly = !( ( s.st_mode & S_IWUSR ) == S_IWUSR ); // TODO:LINUX Is this the correct behaviour?
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(637): s.st_mode &= ( ~S_IWUSR ); // TODO:LINUX Is this the correct behaviour?  
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(642): s.st_mode |= S_IWUSR; // TODO:LINUX Is this the correct behaviour?
  
/*static*/ bool FileIO::GetReadOnly( const char * fileName )  
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(676): return ( ( s.st_mode & S_IWUSR ) == 0 );// TODO:LINUX Is this the correct behaviour?
  
/*static*/ bool FileIO::SetExecutable( const char * fileName )
  C:\p4\depot\Code\Core\FileIO\FileIO.cpp(687):  // rwxr-x--x (751) TODO:LINUX TODO:MAC Is this correct?
  
[...]

PR #956, opened 2023-02-01
Fix Linux and MacOS file permissions:

PJayB:
# Description:

Fix Linux and MacOS file permissions:

* `SetExecutable`: Today, existing permissions are destroyed. The fix preservss any existing read-write permissions and only adds `x`.
* `SetReadOnly`: Generally, a file should only be writable for the user that created it. Thus, when removing the flag, remove `w` from everyone. When adding the flag, only set for the user.
* `mkdir`: Create directories with 0755 permissions. By default, directories shouldn't be writable by anyone but the user that created it.

Fixes issue https://github.com/fastbuild/fastbuild/issues/200.

# Checklist:

The pull request:
- [X] **Is self-contained**
- [X] **Compiles on Windows, OSX and Linux** (TODO - will test on Linux & Windows, but will need help with MacOS)
- [X] **Has accompanying tests** (I figured the existing coverage would be sufficient, but please let me know if you want to see more detailed tests)
- [X] **Includes documentation** (fixes for existing feature)

### DIFF
--- a/Code/Core/FileIO/FileIO.cpp
+++ b/Code/Core/FileIO/FileIO.cpp
@@ -500,7 +500,7 @@
         }
     #elif defined( __LINUX__ ) || defined( __APPLE__ )
         umask( 0 ); // disable default creation mask // TODO:LINUX TODO:MAC Changes global program state; needs investigation
-        mode_t mode = S_IRWXU | S_IRWXG | S_IRWXO; // TODO:LINUX TODO:MAC Check these permissions
+        mode_t mode = (S_IRWXU | S_IRGRP | S_IXGRP | S_IROTH | S_IXOTH);
         if ( mkdir( path.Get(), mode ) == 0 )
         {
             return true; // created ok
@@ -888,7 +888,8 @@
         {
             return true; // can't even get the attributes, treat as not read only
         }
-        const bool currentlyReadOnly = !( ( s.st_mode & S_IWUSR ) == S_IWUSR ); // TODO:LINUX Is this the correct behaviour?
+        const unsigned int anyWriteBits = S_IWUSR | S_IWGRP | S_IWOTH;
+        const bool currentlyReadOnly = ( ( s.st_mode & anyWriteBits ) == 0 );
         if ( readOnly == currentlyReadOnly )
         {
             return true; // already in desired state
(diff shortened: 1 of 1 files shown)


Code/Core/FileIO/FileIO.cpp
```diff
@@ -898,12 +898,12 @@
         if ( readOnly )
         {
             // remove writable flag
-            s.st_mode &= ( ~S_IWUSR ); // TODO:LINUX Is this the correct behaviour?
+            s.st_mode &= ~( S_IWUSR | S_IWGRP | S_IWOTH );
```
ffulin, line 901: If setting writeable only adds the user flag, it seems that removing writable should also only remove thwe user writable flag?

ffulin:
Thanks for taking a look at this. I think these changes make sense, with the exception of SetReadOnly behavior not being symmetrical which doesn't seem correct to me.

PJayB:
If you want to think of it symmetrically, one way to think about it is that SetReadOnly is doing the most conservative operation in either direction.

When marking something as readonly, it wouldn't make sense to only mark the file as RO for just the User that created it, but leave write available to the wider Group or Other. Doing anything else invalidates the purpose.

Conversely, when setting write permissions, typically only the User has write permissions unless that user explicitly grants Group or Other access. So we only want to set the writeable bit in this case.

PJayB:
@ffulin I've incorporated some MacOS build fixes from @Leandros. He confirmed over on #200 that FBuild builds and tests pass on MacOS with this change.

Does the above comment about the read-only bits make sense?

Code/Core/Env/Assert.cpp
```diff
@@ -85,7 +85,7 @@ bool IsDebuggerAttached()
         const uint32_t BUFFER_SIZE( 4096 );
         char buffer[ BUFFER_SIZE ];
         #if defined( __APPLE__ ) || defined( __LINUX__ )
-            sprintf( buffer,
+            snprintf( buffer, BUFFER_SIZE,
```
ffulin, line 88: These snprintf changes seem ok but unrelated to file permissions?

PJayB: Removed. @Leandros needed these changes to build for OSX, but I'll let you guys work that out.

ffulin:
Thanks for the fixes! These have been integrated and will be in the next release (v1.10)

PR #966, opened 2023-02-25
Fix collision errors on chained ObjectLists 

### LINKED ISSUE #964: ObjectList inputs do not respect base path from input ObjectLists.
Hey!

Recently when I was updating a few of my scripts to handle ObjectList inputs to another ObjectList. I've run into an issue where the first ObjectList outputs are stored in sub-folders (be it due to matching in sub-directories or base root path), which allow you to have the same file name without collisions. Ex.: subA/file.shader -> subA/file.preprocessed_shader, subB/file.shader -> subB/file.preprocessed_shader

Now when we push the outputs of this ObjectList into another ObjectList the files are added without a basepath, resulting in collisions as the base path is not provided.
Looking at the code, it seems the only change needed is to set the base path for the created DynamicObjectNode to the output compiler path of the input object list.

```cpp
            const ObjectListNode * objListNode = dep.GetNode()->CastTo< ObjectListNode >();
            // ...
            
// Old
                // create the object that will compile the above file
                if ( CreateDynamicObjectNode( nodeGraph, n->GetName(), AString::GetEmpty() ) == false )
                
// New
                // create the object that will compile the above file
                if ( CreateDynamicObjectNode( nodeGraph, n->GetName(), objListNode->m_CompilerOutputPath ) == false )
```

I don't have right now any pull request or ability to do so, but if you see anything wrong with this change, please let me know.

Dandielo:
# Description:
- Introduced a new GetCompilerPath method.
- Dynamic object nodes created from chained ObjectLists are now using the source ObjectList compiler output path. This properly handles sub-directories if the source list also had sub-directories in its outputs.
- Added a new file in a subdirectory to "Chained Object Lists" test. The file name is exactly the same as the file at the root, failing the test with a collision if a regression in behavior appears.

Closes #964

# Checklist:

The pull request:
- [x] **Is self-contained**


ffulin:
This change has been integrated and will be in the next release (v1.10) Thanks for the fix :)

PR #973, opened 2023-04-21
Get JSON files produced by -time-trace from cache and worker

Lovali:
# Description:

The `-ftime-trace` clang flag produces Chrome Tracing `.json` file next to the `.obj` (see [this article](https://aras-p.info/blog/2019/01/16/time-trace-timeline-flame-chart-profiler-for-Clang/)). With this commit we wanted to be able to store and retrieve json files from cache if it compiles with the `-ftime-trace` flag in the same time as the `.obj` file, and also we wanted to get this file from the worker after the distributed compilation.
In addition to this commit we needed [this commit](https://github.com/fastbuild/fastbuild/pull/871).

# Checklist:

The pull request:
- [x] **Is self-contained**
- [ ] **Compiles on Windows, OSX and Linux** (did not have the chance to test)


Code/Tools/FBuild/FBuildCore/Protocol/Protocol.h
```diff
@@ -31,7 +31,7 @@ namespace Protocol
     enum : uint16_t { PROTOCOL_PORT = 31264 }; // Arbitrarily chosen port
 
     // Protocol Version
-    enum : uint32_t { PROTOCOL_VERSION_MAJOR = 22 };    // Changes here make workers incompatible
+    enum : uint32_t { PROTOCOL_VERSION_MAJOR = 23 };    // Changes here make workers incompatible
```
ffulin, line 34: The protocol generally needs to remain compatible between versions because many users rely on this (sharing workers between various projects on varying versions). The major protocol version would ideally never be bumped, but exists for should a change be needed in the future that is impossible to keep compatible.

The minor version can be used to allow workers and clients to operate in a forwards or backwards compatible fashion. There isn't an example of this exact type of operation, but I think GetDistributableJobToProcess could be modified to pas in the minor version of the worker requesting the job which it could use to know if that worker can support a job with this flag.

I hope to do further cleanup in the future that would make this logic more easily exist in the CompilerDriver which would make a change list this simpler.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -1670,6 +1692,14 @@ void ObjectNode::GetExtraCacheFilePaths( const Job * job, Array< AString > & out
         outFileNames.Append( xmlFileName );
     }
 
+    // JSON files produced by -ftime-trace option
+    if ( objectNode->m_CompilerFlags.IsUsingClangTimeTrace() )
+    {
+        AStackString<> jsonFileName;
+        objectNode->GetTimeTraceJsonName( jsonFileName );
+        outFileNames.Append( jsonFileName );
```
ffulin, line 1700: This will require the cache version to be bumped, otherwise older workers will break when retrieving older cache entries that don't have a 5th file in the result they retrieve.

It's ok to simply increment the cache version to fix this.

ffulin:
I think this changes is reasonable, but it's worth noting that it may result in some confusing results. Specifically, the same object could be built remotely or stored to the cache by a machine with significantly different performance characteristics to the local host, which means comparison of one object file to another becomes more tricky.

Performance between objects on the local host can also vary considerably in a multi-threaded environment so this isn't an entirely new problem I suppose, but it may cause some surprises.

ffulin:
Changes like this require accompanying tests. I can write those, but it means it'll be much longer before I can integrate this change. If you're able to provide tests, that'll make it possible to integrate sooner.

Lovali:
> I think this change is reasonable, but it's worth noting that it may result in some confusing results. Specifically, the same object could be built remotely or stored to the cache by a machine with significantly different performance characteristics to the local host, which means comparison of one object file to another becomes more tricky.
> 
> Performance between objects on the local host can also vary considerably in a multi-threaded environment so this isn't an entirely new problem I suppose, but it may cause some surprises. (The most accurate result would require a single-threaded, no-cache and no distribution clean build)

We are aware of that but we still think that it can help spot when compilation time is abnormally high for a file we are working on for example, and if we want to confirm it or if we want to have an accurate compile time report we can still disable cache and distributed compilation.

Lovali:
Hi, all checks have passed and I added the tests as requested. Do you think it could be merged ?

Code/Tools/FBuild/FBuildCore/Protocol/Protocol.h
```diff
@@ -32,7 +32,7 @@ namespace Protocol
 
     // Protocol Version
     enum : uint32_t { PROTOCOL_VERSION_MAJOR = 22 };    // Changes here make workers incompatible
-    enum : uint8_t  { PROTOCOL_VERSION_MINOR = 2 };     // Changes must be forwards and backwards compatible
+    enum : uint8_t  { PROTOCOL_VERSION_MINOR = 3 };     // Changes must be forwards and backwards compatible
```
ffulin, line 35: This changes the minor version of the protocol, but none of the logic in this PR uses that version to ensure backwards/forwards compatibility.

If a client was built with this code, and connected to a worker of a previous version it would crash trying to extract the json file from the payload in Client.cpp as it would not be present in the payload.

Similarly, a worker built with this version should not attach the new file for an older client when it doesn't expect that file to be attached.

Lovali: It seems that the server never gives its protocol version to the client so it will be difficult to ensure the backwards/forwards compatibility for the worker. Do you have any ideas on how to do this ?

Lovali: Hi, we still have not found on how to do this. Maybe you have an idea ?

ffulin:
Sorry for the slow reply. I think we're probably missing a few things in the core connection management/protocol to make this easy to implement robustly.

The first thing that's missing is, as you noted, the Client doesn't know what version the Server is running. I went ahead and added that here: https://github.com/fastbuild/fastbuild/commit/99ff043e809aae0b8d81346486ed7cd72730f612

That should allow us to add necessary logic in the Client to not pass jobs to workers when they use this new flag, unless they are using a new protocol version that allows them to return a valid result.

There will likely be other considerations, but we should be able to make some progress now hopefully.

PR #976, opened 2023-05-03
Fix docs to mark various function properties as optional that were incorreclty marked as required

harrisont:
# Description:

Fix docs to mark various function properties as optional that were incorrectly marked as required:
* `DLL`/`Executable`: `Libraries2`
    * [LinkerNode.cpp](https://github.com/fastbuild/fastbuild/blob/c78009ad8b2a3c11427735dbe7cf912521f9675f/Code/Tools/FBuild/FBuildCore/Graph/LinkerNode.cpp#L40) shows that it's optional. `DLLNode` and `ExeNode` use `LinkerNode` as a base.
* `XCodeProject`: `XCodeOrganizationName`
    * [XCodeProjectNode.cpp](https://github.com/fastbuild/fastbuild/blob/85f8e54077b81f29c165040c2c0702265b5734c7/Code/Tools/FBuild/FBuildCore/Graph/XCodeProjectNode.cpp#L42) shows that it's optional.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/functions/library.html
+++ b/Code/Tools/FBuild/Documentation/docs/functions/library.html
@@ -37,7 +37,7 @@ Builds a statically linked library.
   ; options for compilation
   .Compiler                 ; Compiler to use
   .CompilerOptions          ; Options for compiler
-  .CompilerOutputPath       ; (optional) Path to store intermediate objects
+  .CompilerOutputPath       ; Path to store intermediate objects
   .CompilerOutputExtension  ; (optional) Specify the file extension for generated objects (default .obj or .o)
   .CompilerOutputPrefix     ; (optional) Specify a prefix for generated objects (default none)
 
@@ -49,7 +49,7 @@ Builds a statically linked library.
                             ; Default is 'auto' (use the librarian executable name to detect)
   .LibrarianOutput          ; Output path for lib file
   .LibrarianAdditionalInputs; (optional) Additional inputs to merge into library
-  .LibrarianAllowResponseFile ; (optional) Allow response files to be used if not auto-detected (default: false)
+  .LibrarianAllowResponseFile ; (optional) Allow response files to be used if not auto-detected (default: false)  
   .LibrarianForceResponseFile ; (optional) Force use of response files (default: false)
 
   ; Specify inputs for compilation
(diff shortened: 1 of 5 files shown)


Code/Tools/FBuild/Documentation/docs/functions/library.html
```diff
@@ -37,7 +37,7 @@ <h1>Library</h1>
   ; options for compilation
   .Compiler                 ; Compiler to use
   .CompilerOptions          ; Options for compiler
-  .CompilerOutputPath       ; Path to store intermediate objects
```
ffulin, line 40: CompilerOutputPath is conditionally optional. It's only optional in the special case of using ObjectList or Library to compile a precompiled header separately from the objects that consume it.

MetaOptional is used in the code and the emptiness is checked as a special case to manage this.

I think either way you look at it, this one-liner documentation can't communicate that special case effectively, so will always be wrong one way or another. I think it could be argued that listing it as required since that is the most common case might be preferable.

Code/Tools/FBuild/Documentation/docs/functions/objectlist.html
```diff
@@ -37,7 +37,7 @@ <h1>ObjectList</h1>
   ; options for compilation
   .Compiler                 ; Compiler to use
   .CompilerOptions          ; Options for compiler
-  .CompilerOutputPath       ; Path to store intermediate objects
```
ffulin, line 40: Same comment here as for Library above

Code/Tools/FBuild/Documentation/docs/functions/objectlist.html
```diff
@@ -54,15 +54,15 @@ <h1>ObjectList</h1>
   .CompilerInputUnity          ; (optional) Unity to build (or Unities)
   .CompilerInputAllowNoFiles   ; (optional) Don't fail if no inputs are found
   .CompilerInputObjectLists    ; (optional) ObjectList(s) whos output should be used as an input
-  
+
```
ffulin, line 57: It would be preferable to avoid unrelated whitespace changes if possible. There are quite a few in this review.

ffulin:
Thanks for the fixes. All but one of these look like silly errors on my part.

Code/Tools/FBuild/Documentation/docs/functions/library.html
harrisont: Sounds good. I undid the change to `CompilerOutputPath`, both in `library.html` and in `objectlist.html`.

Code/Tools/FBuild/Documentation/docs/functions/objectlist.html
harrisont: Sounds good. I undid the change to `CompilerOutputPath`, both in `library.html` and in `objectlist.html`.

Code/Tools/FBuild/Documentation/docs/functions/objectlist.html
harrisont: 👍I undid the whitespace changes.

ffulin:
Thanks again for the fixes. These have been integrated and will be part of the v1.10 release.

PR #985, opened 2023-06-09
Cache store slice added on profiling

Lovali:
# Description:

Add a `Cache Store` slice to profiling to have a better view of the whole caching process.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1534,6 +1534,9 @@ void ObjectNode::WriteToCache_FromDisk( Job * job )
     WriteToCache_FromUncompressedData( job,
                                        buffer.GetData(),
                                        buffer.GetDataSize() );
+
+    job->GetBuildProfilerScope()->SetStepName( "Cache Store" );
+
 }
 
 // WriteToCache_FromUncompressedData


ffulin:
It's desirable to improve the utility of the profile output, but there are a few problems with this change:

**1 - It obfuscates the type of compilation operation that was performed.**
The various states related to "racing" a remote worker are overwritten.

**2 - It doesn't cover all paths**
There are two paths for storing to the cache:
a) (the path you annotated) when compiling locally, retrieving the written results from disk and storing them
b) when receiving a completed job from a remote worker

The second case is handled when retrieving the result from a remote worker so is a bit more challenging to associate with an item on the timeline. Without this though, the profile would present an incomplete picture of what was stored.

---

I think with these things in mind, this change as-is is probably not desirable.

The structure of the chrome tracing format limits exactly what we can do, but here are some ideas about how we might proceed:

1 - Adding additional information (like we already do with the filename) so that when you click on the item it can show if we stored to the cache or not. It could include timing or other information as well. This is quite flexible and doesn't impact the existing data being reported, but isn't as visible as your proposed change. It might be possible to structure this data so that chrome pro

2 - Add another "track" showing total cache stores, hits and misses through the build.
Not useful for seeing whether individual items were stored, but would show an overall picture of things.

3 - Experiment with other styles of formatting
Chrome tracing supports some other types of markup which might be possible to leverage in a useful way.

---

Just in case it's not known, if you're looking at debugging something, if you build from source, the Profile configuration of the code has a far more detailed set of macros that can be used for deeper instrumentation (as opposed to the -profile feature which is end-user facing)

Issue #814, opened 2021-02-01
Loopback IP issue with Windows 10 and Hyper-V


nil-ableton:
If you have enabled once Hyper-V (VM support). FASTbuild 1.0.4 (at least) starts to send traffic to the virtual switch ip address for localhost as if it was a remote. I.e. with this:

```
> ipconfig
Windows IP Configuration
Ethernet adapter vEthernet (Default Switch):
   Connection-specific DNS Suffix  . :
   Link-local IPv6 Address . . . . . : fe80::c67:437e:a552:f809%27
   IPv4 Address. . . . . . . . . . . : 172.28.96.1
   Subnet Mask . . . . . . . . . . . : 255.255.240.0
   Default Gateway . . . . . . . . . :
Ethernet adapter Ethernet:
   Connection-specific DNS Suffix  . :
   IPv6 Address. . . . . . . . . . . : 2a01:e0a:216:e840:f179:aca9:e62b:6d4a
   Temporary IPv6 Address. . . . . . : 2a01:e0a:216:e840:f9bb:272d:6f3d:6c40
   Link-local IPv6 Address . . . . . : fe80::f179:aca9:e62b:6d4a%10
   IPv4 Address. . . . . . . . . . . : 192.168.1.9
   Subnet Mask . . . . . . . . . . . : 255.255.255.0
   Default Gateway . . . . . . . . . : fe80::72fc:8fff:fe51:288c%10
                                       192.168.1.254
```
I see FASTbuild sending remote work to self/localhost:
```
-> Obj: C:\repos\Unity_Src_cpp-25.obj <REMOTE: 172.28.96.1>
-> Obj: C:\repos\Unity_Src_cpp-2.obj <REMOTE: 172.28.96.1>
```

ffulin:
I'm guessing you are running an FBuildWorker inside the VM and running FBuild on the host? If so, this is expected behavior. This is how you can test changes to distributed compilation for example.

It might be possible to detect this situation and explicitly ignore those hosts, but I think that would be a new feature would need to be off by default.

Let me know if I've misunderstood the problem.

nil-ableton:
Sorry I wasn't clear. No, I'm not running any VM at the moment. Both FBuildWorker and FBuild run on the host.

Rinn:
[...]

PR #986, opened 2023-06-10
Add option and envvar for overriding the FBuildWorker IP Address

dinibu:
# Description:

 - See [issue 814](https://github.com/fastbuild/fastbuild/issues/814) _Loopback IP issue with Windows 10 and Hyper-V_
 - Under Windows, FBuildWorker will often choose the Hyper-V virtual ethernet address, if e.g. WSL is installed, and cause local addresses to be used as the IP address advertised to FBuild instances checking for workers in a brokerage path, where successful connections cannot be made. 
 - The added command line option (`-ipaddress`) and environment variable (`FASTBUILD_WORKER_IP_ADDRESS`) overrides the IP address discovered by FBuildWorker, so that it can advertise a desired IP address on the worker brokerage. The IP address is also is shown in the UI window title alongside the host name used on the brokerage.

# Checklist:

The pull request:
- [ v ] **Is created against the Dev branch**
- [ v ] **Is self-contained**
- [ ? ] **Compiles on Windows, OSX and Linux**
  - Windows compiles successfully.
  - Linux under Windows with WSL seems to run into a link issue with the Fuzzer library binary, at least on my Ubuntu 20.04 container. I've had similar issues with 22.04, although I had to hack to use Clang-11 with `clang-10` not being available. This also happens without my change, so it looks like there's an issue with my Linux compiler set up.
  - Unfortunately I have no access to a Mac OSX machine to verify this change under OSX either. 
  - The change itself does not introduce any usages of platform-specific code to my knowledge, the existing platform abstractions are used for e.g. getting the environment variable value. 
- [ X ] **Has accompanying tests** 
  - There are no existing tests that test FBuildWorker and the Brokerage classes. Testing the `WorkerBrokerageServer` class would require decoupling from e.g. Network API calls that call real system APIs for e.g. `getaddrinfo()`. This would be a much larger architectural change, and weighing up against size of the changes to the brokerage class to add an IP address string override, I would suggest such a change would be more suited for a separate future change to improve its testability. 
  - This change does not otherwise modify any of the other FBuild core libraries.
- [ v ] **Passes existing tests**
- [ ? ] **Keeps Windows, OSX and Linux at parity**
  - Build issues prevent this from being verified, but given the use of platform abstractions for e.g. retrieving environment variables, this change should maintain parity across those platforms.
- [ v ] **Follows the code style**
- [ v ] **Includes documentation**

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/environmentvariables.html
+++ b/Code/Tools/FBuild/Documentation/docs/environmentvariables.html
@@ -44,6 +44,10 @@
     <td>Set location of the Brokerage Path for distributed compilation.</td>
   </tr>
   <tr>
+    <td><a href="#FASTBUILD_WORKER_IP_ADDRESS">FASTBUILD_WORKER_IP_ADDRESS</a></td>
+    <td>Set the IP address for FBuildWorker to advertise as.</td>
+  </tr>
+  <tr>
     <td><a href="#FASTBUILD_WORKERS">FASTBUILD_WORKERS</a></td>
     <td>Set the list of workers explicitly.</td>
   </tr>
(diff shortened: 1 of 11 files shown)


dinibu:
Oh, I just realised the change breaks when the IP address is not overridden, will post a new revision when I next have a chance.

ffulin:
These changes seem fairly reasonably, but it seems like having to pass the IP address manually would be rather inconvenient (either via the command line or an env var). I wonder if it would be possible to make the IP address lookup more intelligent to just have it work better by default.

**> Linux under Windows with WSL seems to run into a link issue with the Fuzzer library binary,**

Linking the fuzzer requires some additional setup. Some documenation exists here: Tools\FBuild\BFFFuzzer\README.md
I don't think running that is necessary for your tests. A work-around to allow everything to run to completion even if that fails is to use -nostoponerror

GTez:
I’m not sure if there’s a good way to “bump” an issue. We’ve been finding a lot of our users’ PCs don’t get correctly added to the worker pool, and the files added are showing their HyperV IPs from WSL

Drigax:
Is this change good to check in? I'm looking forward to having a solution to "force" a worker to advertise they're at a specific IP

PR #988, opened 2023-06-23
Upgraded sort code to use QuickSort instead of ShellSort.

rmitton:
Jobs in FASTBuild are sorted using _ShellSort_, which has execution time of O(n^2). When 'n' is large, which can happen on builds containing thousands of jobs, the time taken to perform the sorting quickly becomes notable, sometimes taking over 30 seconds to complete.

Data can be sorted quickly using an O(n log n) algorithm such as _QuickSort_ (C. A. R. Hoare, 1961). QuickSort uses a divide-and-conquer strategy to recursively split the problem into smaller tasks. This implementation uses an explicit stack to avoid function-call overhead and bound stack depth in case of adverse inputs (degrading to a slower sort under the rare case of overflow), chooses median-of-three as its partition function to maintain recursion balance, and handles repeated equal values well.

Other options such as deferring to the _libc qsort_ or _STL std::sort_ were not considered due to the pre-existing coding style of the FASTBuild codebase. Switching algorithms to a distributive sort such as Radix Sort was discounted to maintain API compatibility.

### DIFF
--- a/Code/Core/Containers/Sort.h
+++ b/Code/Core/Containers/Sort.h
@@ -30,37 +30,80 @@ public:
     }
 };
 
-// ShellSort
+template < class T > 
+static inline void   SortSwap( T & a, T & b )      { T tmp( a ); a = b; b = tmp; }
+static inline size_t SortMin( size_t a, size_t b ) { return a < b ? a : b; }
+
+// QuickSort
 //------------------------------------------------------------------------------
 template < class T, class COMPARE >
-void ShellSort( T * begin, T * end, const COMPARE & compare )
+void QuickSort( T * begin, T * end, const COMPARE & less )
 {
-    const size_t numItems = (size_t)( end - begin );
-    size_t increment = 3;
-    while ( increment > 0 )
+    struct Stack { T *arr, *end; } stack[ 64 ], range = { begin, end };
+    int depth = 0;
+
+    for (;;)
     {
-        for ( size_t i = 0; i < numItems; i++ )
+        // Sort small runs with O(n^2) insertion sort, large with O(log n) quick sort.
+        size_t count = size_t( range.end - range.arr );
+        if ( count <= 16 || depth >= 63 )
         {
-            size_t j = i;
-            T temp( begin[ i ] );
-            while ( ( j >= increment ) && ( compare( temp, begin[ j - increment ] ) ) )
+            for ( size_t i = 1; i < count; i++ )
             {
-                begin[ j ] = begin[ j - increment ];
-                j = j - increment;
+                T tmp( range.arr[ i ] );
+                size_t j = i;
+                while ( j > 0 && less( tmp, range.arr[ j - 1 ] ) )
+                {
+                    range.arr[ j ] = range.arr[ j - 1 ];
+                    j--;
+                }
+                range.arr[ j ] = tmp;
             }
-            begin[ j ] = temp;
-        }
-        if ( increment / 2 != 0 )
-        {
-            increment = increment / 2;
-        }
-        else if ( increment == 1 )
-        {
-            increment = 0;
-        }
-        else
-        {
-            increment = 1;
+
+            // Pop new range off stack.
+            if ( depth <= 0 )
+                break;
+            range = stack[ --depth ];
+        } else {
+            // Pick median-of-3 as pivot.
+            T *a = range.arr, *b = range.arr + ( count >> 1 ), *c = range.end - 1;
+            if ( less( *c, *a ) ) SortSwap( *a, *c );
+            if ( less( *b, *c ) ) SortSwap( *c, *b );
+            if ( less( *c, *a ) ) SortSwap( *a, *c );
+            T pivot(*c);
+
+            // Partition into (eq, lt, gt, eq). (Bentley/McIlroy, "Engineering a Sort Function", 1993)
+            size_t i = 0, j = count - 1, ieq = 0, jeq = count - 1;
+            for (;;)
+            {
+                while ( i <= j && !less( pivot, range.arr[ i ] ) )
+                {
+                    if ( !less( range.arr[ i ], pivot ) )
+                        SortSwap( range.arr[ ieq++ ], range.arr[ i ] );
+                    i++;
+                }
+                while ( j >= i && !less( range.arr[ j ], pivot ) )
+                {
+                    if (!less( pivot, range.arr[ j ] ) )
+                        SortSwap( range.arr[ jeq-- ], range.arr[ j ] );
+                    j--;
+                }
+                if (i > j)
+                    break;
+                SortSwap( range.arr[ i++ ], range.arr[ j-- ] );
+            }
+
+            // Move elements equal to the pivot back into the middle.
+            size_t s = SortMin( ieq, i - ieq );
+            for (size_t l = 0, h = i - s; s; s--)
+                SortSwap( range.arr[ l++ ], range.arr[ h++ ] );
+            s = SortMin( jeq - j, count - 1 - jeq );
+            for (size_t l = i, h = count - s; s; s--)
+                SortSwap( range.arr[ l++ ], range.arr[ h++ ] );
+
+            // Save one half for later and recurse into the other.
+            stack[ depth++ ] = { range.arr + count - ( jeq - j ), range.end };
+            range = { range.arr, range.arr + ( i - ieq ) };
         }
     }
 }
--- a/Code/Core/Containers/Array.h
+++ b/Code/Core/Containers/Array.h
@@ -53,10 +53,10 @@ public:
     void                        Swap( Array< T > & other );
 
     // sorting
-    void                        Sort() { ShellSort( m_Begin, m_Begin + m_Size, AscendingCompare() ); }
-    void                        SortDeref() { ShellSort( m_Begin, m_Begin + m_Size, AscendingCompareDeref() ); }
+    void                        Sort() { QuickSort( m_Begin, m_Begin + m_Size, AscendingCompare() ); }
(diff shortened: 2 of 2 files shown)


Code/Core/Containers/Sort.h
```diff
@@ -30,37 +30,80 @@ class AscendingCompareDeref
     }
 };
 
-// ShellSort
```
missmah, line 33: Wondering why `ShellSort` is removed here - presumably there are no current uses for it, but it could potentially find uses in the future? Or even be used in a unit test of the new quicksort implementation?

ffulin:
Changing the default sort algorithm, particularly to a hybrid one might make sense. That approach is the standard on most stl implementations for good reason. I certainly want to avoid pathological edge cases and have the default Sort() be broadly performant.

**Background on current implementation**

To add some background, the existing sort algorithm was chosen over quicksort originally for a few reasons:
 * purported better or equivalent worst-case performance ([n log n vs n^2](https://en.wikipedia.org/wiki/Sorting_algorithm), or [n^2 vs n^2](https://www.geeksforgeeks.org/time-complexities-of-all-sorting-algorithms/) depending on the source) (at the expense of worse average time)
 * better space complexity
 * implementation simplicity
Naturally, big-O complexity notation is not the only factor in actual real world performance. The gap sequences, cost of comparisons and cost of exchanges or copies are all factors, so performance was profiled and I've not seen any performance issues even on large codebases with hundreds of projects and dozens of build configurations.

**Questions about your use case**

I'm curious to learn a bit more about exactly which situation you saw this being slow and what order of magnitude the number of jobs was.

For example, the QueueJobs function that pushes the main thread queue into the job system, combining it with the existing queue has 2 sorts in it. One to sort the new jobs and a second on to sort the merged lists. Do you see slow sorts in the first or second sort? If it was the second one for example, we could do better than re-sorting the list my performing a merge of the already two sorted lists.  If it was the first sort, it might be good to limit the size of that queue, and flush it periodically (which would have some other benefits to jobs processing)

Just to be clear, those changes may be complimentary to changing the sort algorithm.

**Tests**
Various tests appear to be failing on Windows, hitting some asserts and then ultimately resulting in the application being terminated with a "fail fast" exception (likely indicating some sort of buffer overflow). It's strange that the failures occur on Windows only.

The Array class does implement some Sort() tests but perhaps the coverage is not adequate. (It's also possible the failure is not in the sort itself but perhaps some existing issue has been uncovered).

**Thoughts about Moves**
Looking at both the existing sort implementation and the proposed new one, I think there is an opportunity to improve sort performance by utilizing the Move helper (FASTBuild's std::move equivalent). Many performance critical sorts in FASTBuild use SortDeref to avoid deep copies so wouldn't benefit, but some sorts are using Sort() and would benefit from using move semantics.

Code/Core/Containers/Sort.h
ffulin: I'd be ok with deleting it if we determine it should never be used. I'd also be ok with moving it into the tests to have a simple way of verifying the sort/sorts we chose to keep (though it may be of limitted use since the sorts we're using are unstable and may not produce the same results for lists with "equal" items)

ffulin:
I spent some time constructing some build graphs with very large amounts of jobs and manipulating them so as to cause large amounts of re-sorting when combining the list of new jobs with existing jobs. I managed to craft some extreme cases that caused 10+ seconds worth of sorting when adding new job batches into the queue.

In my scenario, the majority of the cost was spent inside the re-sorting of the combined lists, so I went ahead and replaced that with a merge (https://github.com/fastbuild/fastbuild/commit/979f6054e557d1628f62b8f532614cb027dec668) as I mentioned might be nice to do in my previous comment.

In my testing it all but eliminated the 10+ second stall of my forced bad case and was an nice improvement in all other real-world scenarios i measured.

I'd be interested to hear if this change improves the situation you were seeing.

I still think the sort itself is still worth considering for further optimization and these changes are complimentary.

I did also experiment with earlier flushing of job batches as I mentioned before, and while that yielded some interesting results due to how it impacts job scheduling, it improved some cases while making others worse. There could be some heuristic to apply here that might still make this useful, but for now this isn't an obvious win.

MinisculeGirraffe:
> 

Alexa. Write me a petty response please.

jan-vandenberg:
Could this actually be a response by Donald Knuth? I doubt it.

It doesn't read like him (i.e. I don't think he would call it 'Donald Shell's idea'). Also you don't have to print and mail Knuth. Yes, he doesn't use e-mail, but his staff does. I know this, because I myself have interacted this way with Knuth before.

Overall there is no other 'proof' that this is actually a response by Knuth.

rmitton:
I would like to take this time to thank the members of the open-source community for their kindness, warmth, and general goodwill. It's been nice to see folks take such a keen, spirited interest in the history of computer science and the practical details of sorting algorithm implementation.

As there seems little interest in pursuing this pull request any further however, I have elected to close it and move on with other things.

I wish the fine, daring algorithm designers over at FASTBuild the best of luck in future with their bold independent research.

ffulin:
@rmitton I'm not sure where you got the impression that I didn't want to fix this issue or that I wasn't open to merging your changes.

My initial response to you was:

> **Changing the default sort algorithm, particularly to a hybrid one might make sense. That approach is the standard on most stl implementations for good reason. I certainly want to avoid pathological edge cases and have the default Sort() be broadly performant.**

I wanted to understand your use-case better to make sure any fix was holistic and that we could take advantage of other Further to that, I suggested additional complimentary changes.

I requested additional information from you to help drive this work.

You did not respond.

I mentioned your PR was failing the tests, possibly due to stack corruption.

You did not respond.

While waiting for your response, I went ahead and generated some test cases to emulate my best guess at the issue you were seeing. I eliminated the need for some of the sort calls as I suggested and saw significant uplift in my tests.

I released a new version and asked if this improved your use-case at all, with the expectation that we'd continue talking about the core change you proposed.

You did not respond.

Finally, you responded, but not to any of my queries. Instead, you mentioned that there was that apart from a sub-optimal algorithm being used there was likely also an implementation error. You did this in a rather unpleasant way, but honestly I was kind of amused at the hyperbole - it's hard to believe a function with a silly error in it and an inelegant expression is the worst code any programmer has ever seen.

I would still like to fix this issue for you and for others.

I plan to:
 - fix error in the existing Sort() function as a likely cheap improvement
 - look to replace it with a hybrid sort algorithm as you originally proposed

If you'd still like to help, I'd welcome you to either reopen this PR and perhaps help investigate the crash it causes, or perhaps raise a new PR if that's not possible (or if you prefer).

I am still curious if the changes in the latest version have improved your use-case at all, ideally on top of your changes. I would expect them to still be complimentary.

If you are no longer interested in helping, I'll create a new issue to track this problem and continue fixing it.

MinisculeGirraffe:
> I would like to take this time to thank the members of the open-source community for their kindness, warmth, and general goodwill. It's been nice to see folks take such a keen, spirited interest in the history of computer science and the practical details of sorting algorithm implementation.


You didn't like that people called out on your bad behavior, so you're throwing a tantrum. Like a petty child.

ffulin:
I've fixed the gap sequence calculation in the Sort function, replacing the erroneous fixed starting gap of 3 (which was intended to be n / 3) with Ciura's [2001] empirically derived sequence (https://github.com/fastbuild/fastbuild/commit/0edfdbca0a67a31bf85cfaa35d2d89cace9dbe69)

Sorts that previously took tens of seconds now take tens of milliseconds.

I additionally added move semantic support as mentioned previously, though the gains there are not relevant for the job queuing case and most other uses in FASTBuild avoid the need by performing indirect sorts. Nonetheless, some other infrequent sorts are also improved by this change.

Previous changes to avoid sorts entirely by performing merges remain in-tact.

There may still be modest gains to be had by a change in algorithm, but given the gains and relatively low cost for large sets of data, that may no longer be necessary.

This fix will be in the next release (v1.12)

@rmitton If you are able to confirm if this resolves the performance problem you reported, that would be appreciated. If it doesn't, we can continue to investigate an algorithm switch as you proposed.

rmitton:
Dear @ffulin,

Thank you for expressing continued interest in the performance analysis of sorting algorithms. As I previously mentioned, my focus has shifted to other matters. However, given the heightened interest in the shellsort gap sequence from so many unexpected sources, I thought it'd be beneficial to test the code in detail for those keen on the subject.

To that end, I've taken the various sorting algorithm implementations discussed in this thread and created a test harness (available [here](https://gist.github.com/rmitton/fdf84a56fdfa9a79b051f22c980b8729)). This program evaluates the original ShellSort, the improved ShellSort, the provided QuickSort, and the two built-in C++ library sorts. It times sorting random numbers, with the quantity of 'n' varying from one to a million, and graphs the results in a .SVG file:

analysis graph

The tests were conducted on a Core i9 @ 3.5GHz using Microsoft Visual Studio 2022 with the /O3 optimizer flag. (the graph's Y-axis unfortunately required a mixed linear/logarithmic scale to accommodate them all together)

From the results, it's evident that the improved gap sequence offers *clear* notable enhancements as 'n' increases, and competes favourably with other production sorting implementations. I understand that refining the gap sequence for shellsort is an ongoing area of research in computer science, and I commend the developers for their continued and noble efforts.

Thank you to everyone for the constructive feedback. Given the continued importance of open-source projects like FASTBuild in our industry, those willing to work together to improve such software will no doubt find many future opportunities for tackling large-scale engineering challenges through the art of applied computer science.

ffulin:
v1.12 has now been released. It includes a fix for the core sorting function bug at the core of this thread, as well as the other improvements discussed (elimination of a sort entirely during core graph traversal and better leveraging of move semantics in some sorts)

Issue #959, opened 2023-02-09
[Feature Request] Dynamically adjusting thread pool as ram availability increases/decreases


Caffiendish:
A useful feature for FastBuild to have, would be dynamically adjusting the threadpool, so that, if we're at less than 500MB of ram free, for example, we don't spawn more compiler threads, and lock up the system entirely.
Or conversely, if more ram becomes available, we could launch more threads, and blast through some extra tasks.

I was trying to compile Unreal from source with FBuild present leads to my local system completely locking up, and Windows failing entirely, because FBuild doesn't take into account the requested ram per action (probably because it doesn't know/get told).

Here's a log excerpt from VS, where (I assume) UnrealBuildTool discerns that terrible choices are about to be made, and course corrects:
> 1>Determining max actions to execute in parallel (16 physical cores, 32 logical cores)
1>  Executing up to 16 processes, one per physical core
1>  Requested 1.5 GB free memory per action, 16.82 GB available: limiting max parallel actions to 11

Strangely, UBT will respect processor limitation requests if FBuild isn't present, but since it doesn't pass along the -j param to FBuild, that's something to work out with UBT.

If I actually could use FBuild's distributed building for Unreal, having all threads firing would be really useful for shader compilation, but locking up every machine because I forgot that I'm about to build an absolute ram demon (or the correct variables aren't being sent to FBuild) is slightly less helpful, and it'd be incredibly useful if Fastbuild would take mercy upon us plebs in these cases.

ffulin:
I don't believe a dynamic throttle would be very effective because the peak memory usage of a particular process is not known up-front and varies significantly from invocation to invocation and during the lifetime of the process. At the point where we would decide to spawn another process or not the total memory usage would often be very low. After the processes run for a while the memory use would become too high, and a that point it's too late. The start of the build being the most obvious problematic case, but this problem would be an issue throughout builds.

I think explicitly reducing the concurrent counts via -j in situations where you know the combination of code base and hardware is mismatched is the right way to tackle this problem.

I think as you've noted in the Unreal case, since the UBT is used as a wrapper/layer above FASTBuild and already has logic driven around its knowledge of the specifics of the Unreal codebase, it makes sense for the UBT to pass the appropriate -j parameter to match the throttling it would pass to other build systems already.

I don't think it makes sense for FASTBuild to try and emulate that behavior itself either, since it works with a variety of codebase, most of which are much "leaner" than Unreal.

[...]

PR #995, opened 2023-08-23
Use only physical cores

jorgenpt:
# Description:

Default to only using physical cores

On machines with fast I/O, trying to use all the logical cores
(hyperthreading cores) will slow down the build as the compiler
instances get de-scheduled. Looking at a trace in Superluminal,
clang.exe processes spent a lot of time being pre-empted.

On our project, building with `-j<physical cores>` would frequently take
less than half the time of the old behavior. This change

For cases where people are e.g. working with slow drives, there is a new
`-useallcores` command line option which uses the old behavior for
determining number of worker threads.

# Checklist:

The pull request:
- [X] **Is self-contained**


ffulin:
I'm interested to understand more about your setup/use-case, as the behavior your describe isn't what I see.

I generally see anywhere from a 0% to 10% improvement in overall throughput when using a logical thread count as opposed to physical count for CPU bound builds. More importantly though, for builds that are bound on network IO for caching, i see as much as a 50% improvement using logical thread counts instead of physical counts.

With that in mind, as much as your change appears to help your case, it seems it would be a significant degradation to performance to other cases.

The only situations in which I've seen the case you describe (where using physical core counts instead of logic core counts is significantly faster) is when the build performance is memory limited (when there isn't enough physical memory to properly back the compiler instances and cache necessary files. (Along the lines of what's discussed here: https://github.com/fastbuild/fastbuild/issues/959)

I'd like to understand a bit more about why we might be seeing such differing results, so we can perhaps find a way to have improved performance for your case without harming others.

I'm very interesting in finding better defaults or making tweaks that might improve performance in different cases. With that in mind, are you able to tell me a bit more about your setup?
 - which FASTBuild version are you using?
 - which compiler are you using?
 - what hardware (CPU model & physical memory size)
 - are you using Unity files to accelerate compilation?
 - are you using caching? (and if so, are you using the "LightCache")
 - are you using distributed compilation?

jorgenpt:
> I generally see anywhere from a 0% to 10% improvement in overall throughput when using a logical thread count as opposed to physical count for CPU bound builds. More importantly though, for builds that are bound on network IO for caching, i see as much as a 50% improvement using logical thread counts instead of physical counts.
Comparing this to some of my coworkers on different machines, it seems to behave more like you're describing. 
A coworker with a desktop and an i7-10700 sees ~4m28s clean builds with this change and ~3m 10s without it. Another one with a different Intel desktop CPU is seeing ~3m 14s with this change, 2m 36s without this change.

FWIW this change was modeled on how UnrealBuildTool schedules its compiler invocation. (But they are also more conscious about memory overhead, since their compiler invocations tend to be pretty memory hungry.)

> With that in mind, are you able to tell me a bit more about your setup?
> 
> * which FASTBuild version are you using?
We're currently using dev + some minor patches, as represented here: https://github.com/jorgenpt/fastbuild/tree/v1.11-spryfox

> * which compiler are you using?
clang 13

> * what hardware (CPU model & physical memory size)
12th Gen Intel(R) Core(TM) i9-12950HX with 64GB of RAM (on a laptop connected to an AC adapter)

> * are you using Unity files to accelerate compilation?
No

> * are you using caching? (and if so, are you using the "LightCache")
We are not explicitly using any cache -- should we be using it? Is FASTBuild's cache mechanism similar to a `ccache` type caching? 

> * are you using distributed compilation?
We are not, local only.

ffulin:
Thanks for the additional info.

**> i9-12950HX with 64GB of RAM**

It seems like memory available shouldn't be a constraining factor with 64GiB, so I think we can mostly eliminate that. I think this is therefore a different problem to the one I linked to about constraining scheduling based on/considering memory use.

Your CPU is interesting in that it has "efficiency cores". What may hold true for hyperthreaded cores (often a gain but rarely if ever worse than break even) perhaps doesn't hold true for efficiency cores.

With that in mind, perhaps what we should be doing is excluding efficiency cores in the thread count decision logic. Perhaps instead of trying to exclude hyperthreaded cores, we should be excluding "efficiency" cores. Given your CPU has 8 of each, that would result in the same end result for you, but for others with "normal" hyperthreaded CPUs, the behavior would remain unchanged.

I don't have access to a device with efficiency cores to test on unfortunately. Do you know if these can be differentiated in a useful way when iterating the cores?

tru:
We (Ubisoft) have been running into this with new machines that have hybrid core setup as well. We did quite a bit of testing and have come to the following conclusions:

* Your performance will be atrocious if you have hybrid cores and run on Windows 10, this is because Windows 10 doesn't know about the hybrid setup.
* Even on Windows 11, pushing jobs to all threads will cause a performance reduction.

We did a patch internally that uses Intel's [code](https://github.com/GameTechDev/HybridDetect/blob/main/HybridDetectConsole/HybridDetect.h) for hybrid core detection and use the following changes:

* Change Env::GetNumProcessors() to instead of just returning all threads it does the following:
** Always add all performance cores and their threads
** If the OS can handle HybridCores (Windows 11+) add half of the ecores.
** If the OS cannot handle HC don't add the ecores at all.
** On a machine with 8 pcores, 8 ecores on Windows 11 it would add 8 (pcore) + 8 (pcore threads) + 4 (ecores) = 20 jobs.
** On the same machine with Windows 10 instead it would be 8 + 8 = 16 jobs.

On Windows 10 machines we also change the affinity of threads from WorkerThread and Process to only execute on the pcores. On Windows 11 we let the OS figure that out and it seems to be doing well.

With these changes we see the following performance change:

* My machine runs win11 with 8 pcores and 8 ecores it improved build times from 68s to 64s. So not massive, but not pushing on all ecores helps.
* On a machine with 8 ec / 8 PC running win10 it halved build times from 186s to 97s.

The patch isn't very intrusive, but adding HybridDetect.h is a pretty big dependency, but I am not sure I want to re-implement that complex logic for each operating system.

Is it worth me pushing our patch for this @ffulin ?

ffulin:
@tru Thanks for the extra info.

It does seem that we should add some additional logic to better handle efficiency cores.

**For core count,** would it make sense to simply ignore ecores, leaving them for other uses including non-job control threads in FASTBuild?

> If the OS cannot handle HC don't add the ecores at all.

I think we'll also need to handle the case where the CPU has *only* efficiency cores. For example: https://ark.intel.com/content/www/us/en/ark/products/134797.html

**For affinity,** the Windows 10 only logic seems reasonable.

Again the "efficiency cores only" case would need to be handled.

------

I took a look at the HybridDetect header, and use of it as-is seems like a bit of a non-starter as it depends on the C++ runtime, which FASTBuild doesn't use. The code also seems to be Windows only so doesn't provide any benefit over rolling or own Windows-specific functionality to start. It also seems more complex than it necessary for our use-case.

According to this stack overflow thread response by Agner Fog (https://stackoverflow.com/questions/69955410/how-to-detect-p-e-core-in-intel-alder-lake-cpu), the heart of adding detection logic is to use the CPUID instruction and check some of the bits (and this seems to match the HybridDetect logic at a glance). Enumerating all the cores and executing that logic doesn't seem too tricky, apart from it being difficult to test without access to one of these CPUs.

ffulin:
I've added an Env::GetProcessorInfo() function (https://github.com/fastbuild/fastbuild/commit/c7b262e45eea834b0ddf161bbcaeae155627e6ad) to differentiate Performance/Efficiency cores using cpuid based on Intel's guidance:
https://www.intel.com/content/www/us/en/developer/articles/guide/12th-gen-intel-core-processor-gamedev-guide.html

I have an AMD CPU with no efficiency cores, and it appears to behave sensibly:
```
 - Test 'GetProcessorInfo'
Num Cores: 32 (PCores: 32 + ECores: 0)
```

I don't have access to Intel system with ECores, so I can't fully test it works for the intended case.

**If someone with access to such a device could check if it functions properly, that'd be great.**

If it works as expected, we could use that to start controlling the thread pool creation immediately and could also extend it to track CPU masks to allow control of affinity on Window 10 as noted earlier.

The code is Windows only but could probably be extended to Linux fairly easily. OSX will be more challenging; I think no real intel-based macs have ecores, but for arm macs we'd need to find out whether the same sorts of performance issues exist and custom detection logic would be required (assuming its possible).

missmah:
I would caution against too broad a sword here - I saw faster builds in the past using all cores on 13900K - the optimal answer probably depends on many factors (as alluded to above) - hardware, OS/scheduler, the DAG/workload itself, etc. I also wouldn't be surprised if the most important things would be:
1. Keep tail linking jobs off of e-cores (e.g. never let your last few linking jobs end up on e-cores)
2. Keep root dependency jobs on p-cores (things with lots of dependencies are going to be better on p-cores because the children get to start sooner)
3. For compiling jobs, smaller leaf nodes are likely good candidates for e-cores
4. etc...

DAG/Workload aware scheduling seems like the long-term right answer here.

ffulin:
I think we want to leave core assignment as much as possible to the OS. If the OS is doing it's job, it should manage that for us, and from what Tobias is saying, it seems like Windows 11 at least is doing a reasonable job as long as we don't overload the E cores and leave some headroom for both other processes and additional threads in FASTBuild itself.

Having FASTBuild start making decisions on a per-task basis will hopefully not be necessary as I can see that becoming unwieldy, but we can certainly see how this space evolves. As these hybrid cores become more prevalent, the OS behavior and chipset drivers will likely continue to improve as well.

The goal for now is to avoid the significant performance degradation that seems to occur when all P and E cores are fully loaded by having a better default behavior for the majority of users. All P Cores + half E Cores seems like a good place to start (for Windows 11 at least)

As a fail-safe, we can make sure the number of cores can still continue to be explicitly set as is the case now, or additional controls can be exposed, perhaps in the bff file.

missmah:
My counterargument will be that the OS scheduler lacks the requisite information to do this optimally. The DAG contains this information. Again, scheduling of the last few linker jobs is a great example: round-robin/next available core is often likely sub-optimal here, and the OS scheduler doesn't really know that these jobs are of a different relative latency/throughout importance than any other jobs..

ffulin:
From what I've read (in the intel guidance I linked above and other places) Windows 11 works with a custom chipset driver for these hybrid Intel CPUs and that driver has access to more information about what resources a process is consuming than FASTBuild does, including awareness of the cost of context switches for example. 

How well that driver works, I'm not sure. Whether FASTBuild could do something smarter or hint to that process, I'm also not sure. I think we can leave space for those ideas,  but start with something simple.

lastquest:
Hi @ffulin , I just wanted to chime in to inform you that we (Intel Game DevTech) have been in contact with Tobias and Jorgen about this issue and we want to allocate a bit of time to run some test and provide guidance.
Now that I can build fastbuild I can also test your detection code on hybrid CPUs!

A couple of things to consider:

- Win10 is not hybrid aware and so the suggestion that the app should provide needs to be more explicit, on Win11 the experience should be better just out of the box but there are hints that could improve the overall performance. (Also just to complicate things even more... there are issues with current Win11 version and core parking messing with performance especially when overscheduling)

- There are several P/E hybrid configuration, some of them with significant more E cores numbers rather than P cores, finding the right strategy without excluding E cores is important if we don't want to leave performance on the table.

ffulin:
@lastquest 

Thanks for the additional notes - they all make sense and seem like valid considerations.

I'm open to any learnings or advice you have once you've completed your investigation and tests.

lastquest:
Hi there 
I profiled fastbuild runs on Win11 while building fbuild.bbf itself, I'm using the -j to reduce the threads and using _fbuild -clean_ as the cmd line.

**Desktop i9-12900K**
using logical cores  ~23.7s
using physical cores ~24.3s

**Laptop i7-12700H**
using logical cores count  ~45.5s
using physical cores 	  ~47.7s

I'm not able to reproduce the situation described by Jorgen and Tobias, I suspect that the type of workload they are building is exposing the problem, maybe some long compile task remains on e-cores or a long linking phase?

I asked for ETL traces from Jorgen and Tobias to investigate what is exactly happening on their builds, but with the data I see so far I would not recommend just defaulting using the physical core count instead of the logical core count on hybrid, since a workload like fbuild.fbb seems to perform better with the current setup.

I'll report more when I can analyze a case from Jorgen, Tobias, or something similar.

Also, for Win10 the general recommendation is to not use it at all on a hybrid CPU otherwise the app needs to be hybrid aware by itself. Likely to implement a strategy with different thread pools and prefer using the p-cores first and then e-cores.

ffulin:
FASTBuild's code is small and compiles so quickly, and all cores are fully loaded for only the first 10s or so of the build usually (with the longer tasks/stragglers continuing after that). I thik it's not unsurprising that no problematic behavior would be reproducible in such a simple case.

Game projects tend to have code bases which are several orders of magnitude larger. Hopefully some traces will help identify some problems as you've suggested.

ffulin:
*> I would not recommend just defaulting using the physical core count instead of the logical core count on hybrid,*

I think the proposal is to continue to use all logical cores associated with p-cores in all cases, and look to limit the use of e-cores in various ways (depending on the OS version for example)

PR #1001, opened 2023-10-19
Handle not found full path include

avudnez:
If the branch `PathUtils::IsFullPath( include )` is taken, even for a file that does not exist it will return a non-null `IncludeFile*`.
We had a case in our project where this happens, the header was then considered a dependency and always caused a rebuild.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/LightCache.cpp
@@ -716,7 +716,7 @@ void LightCache::ProcessInclude( const AString & include, IncludeType type )
         }
     }
 
-    if ( file == nullptr )
+    if ( file == nullptr || file->m_Exists == false )
     {
         // Include not found. This is ok because:
         // a) The file might not be needed. If the include is within an inactive part of the file


ffulin:
The current behavior was create intentionally. The idea being that a file may not exist, but may begin to exist later, so we want to capture that dependency should the existence of the file change. It continuing to not exist shouldn't cause rebuilds unless something has changed or has broken.

Whether intentional or not, it's possible this behavior is not actually necessary: i'm trying to think thought the different scenarios.

With that in mind, can you please help me understand your exact scenario a bit better?

ffulin:
Thanks for the quick reply. 

The case you mentioned (missing files due to an SDK not being installed) is the sort of case the logic was intending to handle.

I was thinking that the include being recorded was necessary to handle the case where you later install the missing SDK, making the file appear and thus triggering a build.

However, re-reading my comments inside the branch you modified:
```
        // Include not found. This is ok because:
        // a) The file might not be needed. If the include is within an inactive part of the file
        //    such as a comment or ifdef'd for example. If compilation succeeds, the file should
        //    not be part of our dependencies anyway, so this is ok.
        // b) The files is genuinely missing, in which case compilation will fail. If compilation
        // fails then we won't bake the dependencies with the missing file, so this is ok.
```

I think perhaps that case already works correctly and is compatible with your change.

So I'm leaning towards your change being correct, but I think I'd want to do some testing to reacquaint myself with that code and perhaps extend some tests to make sure we're not missing anything.

avudnez:
Thanks, I'm going to work with this for now then, but I'll keep an eye out.

ffulin:
After an absurdly long time, I was finally able to investigate this problem.

I found that ideally some deeper changes should be performed, but those would have implications outside of the LightCache. Given that, and also because this change makes the full path case consistent with the relative path case, I think it's safe to use as-is for now.

I've gone ahead and integrated this change, and submitted some accompanying test coverage: https://github.com/fastbuild/fastbuild/pull/1162

These changes will be in the next release (v1.19). Thanks again for the PR and reporting this issue!

Issue #971, opened 2023-04-18
Support for clang-tidy


avudnez:
Hello,

We're going to start working on supporting `clang-tidy` for distribution as this will greatly increase the speed of our static analysis.
Since FASTBuild already supports msvc's, I figured you would be interested in `clang-tidy` too.

I was thinking of treating `clang-tidy` as a new compiler that would behave very similarly to clang, except producing no output files. Does that sound like a good way to go ?
One thing I'm unsure about is that it seems `clang-tidy` accepts arguments in msvc-style, probably reusing the `clang-cl` command line parser, but it's not explicitly documented. In which case, FBuild could treat it similar to `clang` _or_ `cl.exe`, but we can't really say in advance.

ffulin:
It's worth noting that clang's static analysis is already supported (with clang or clang-cl). So if you just want static analysis, no additional work is required other than configuring your build targets.

If however you specifically want to leverage the superset of behavior provided by clang-tidy, it'd be pretty reasonable to add built-in support for detecting clang-tidy and behaving sensibly. I'm not intimately familiar with clang-tidy, so I can't say for sure if it would make sense to simply detect when it's used and have one of the existing compiler drivers simply handle any differences, or whether it'd be better to add a new CompilerDriver. Ultimately that'd likely be clear when implementing support and would be driven by how many compiler args FASTBuild needs special case logic or handling for.

avudnez:
From my preliminary testing, built-in support might not even be necessary.
There is one issue with `clang-tidy` though that seems to not behave well with preprocessed source files : https://github.com/llvm/llvm-project/issues/62405

This can probably be worked around while waiting for a fix from llvm, I'll try to investigate it.

ffulin:
Thanks for the update. and the bug link.

*> built-in support might not even be necessary.*

Seems like it might be nice to automatically detect clang-tidy at least. A quick look at the clang detection seems like that won't happen right now (CompilerNode::InitializeCompilerFamily)

Lovali:
We worked on using clang-tidy with FASTBuild and we discovered that clang-tidy does not support response files. This is problematic because most of the time the size of our command line arguments exceeds the limit accepted by a process.
To work around we put all the command line arguments in the ExtraArgs field of clang-tidy's configuration file and we create a configuration file for each file.

[...]

PR #1005, opened 2023-10-30
clang tidy support proof of concept

avudnez:
# Description:

As hinted in #971, we did some groundwork to get `clang-tidy` distributed on fbuild.
This is largely based on work from @lovali during her internship.

_This PR is only a proof of concept_.

On huge projects such as Unreal games, analysis time is extremely long, to the point it's not usable by developers.
We're using it in our projects, but support may be lacking, and it's not the best implementation that can be done.
I had hoped that it would be simple, but it turns out `clang-tidy` has kind of bad support for some CLI shenanigans and its user interface is not as flexible as clang's, in our case.

## Big picture
By default, `clang-tidy` is expected to run on one or more files locally, optionally ran via a launcher script (LLVM provides `run-clang-tidy.py` to this effect). It will look for a `compile-commands.json` file in a parent directory, and try to find the compilation arguments associated to the filename it received in its arguments.
It will also look for a `.clang-tidy` file containing its settings (what checks to run, etc), or receive them via CLI.

When distributing, it's more complex to rely on this behaviour and we want to pass everything we can via CLI so that each job is self-contained and does not rely on the environment.
Unfortunately, since `clang-tidy` does not handle response files correctly (it can read its own arguments via rsp, but not pass an rsp to the underlying clang process), it's a bit complicated to go through the usual fbuild behaviour. Our workaround was to have one configuration file (including compilation arguments) pre-existing for each source file to analyze, and transfer it to the worker with the job data. The advantage is that we get the analysis settings by the way.

Another issue is https://github.com/llvm/llvm-project/issues/62405, `clang-tidy` does not handle diagnostics correctly on already preprocessed source files. To workaround that we added some logic in fbuild to fix filenames and line numbers in diagnostics (not in the fastest way, but simple enough).

Other than that, it works and we hope it can be an interesting addition to fbuild, especially if the implementation/usage can be improved.

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -6,6 +6,10 @@
 #include "PathUtils.h"
 #include "Core/Strings/AStackString.h"
 
+#if defined( __WINDOWS__ )
+#include <windows.h>
+#endif
+
 // Exists
 //------------------------------------------------------------------------------
 /*static*/ bool PathUtils::IsFolderPath( const AString & path )
@@ -290,4 +294,21 @@
     outRelativeFileName += pathB;
 }
 
+// GetFullPath
+//------------------------------------------------------------------------------
+/*static*/ AString PathUtils::GetFullPath( const AString & fileName )
+{
+#if defined( __WINDOWS__ )
+    char fullPath[MAX_PATH] = { 0 };
+    uint32_t size = GetFullPathName( fileName.Get(), MAX_PATH, fullPath, nullptr );
+    if (size == 0)
+    {
+        return AString{};
+    }
+
+    return AString{ fullPath, fullPath + size };
+#else
+#error TODO
+#endif
+}
 //------------------------------------------------------------------------------
--- a/Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriverBase.cpp
+++ b/Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriverBase.cpp
@@ -134,6 +134,13 @@ void CompilerDriverBase::Init( const ObjectNode * objectNode,
 {
 }
 
+// AddPreliminaryArgs_Common
+//------------------------------------------------------------------------------
+/*virtual*/ void CompilerDriverBase::AddPreliminaryArgs( bool /*isLocal*/,
+                                                         Args & /*outFullArgs*/ ) const
+{
+}
+
 // StripTokenWithArg
 //------------------------------------------------------------------------------
 /*static*/ bool CompilerDriverBase::StripTokenWithArg( const char * tokenToCheckFor,
--- a/Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.cpp
+++ b/Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.cpp
@@ -1,0 +1,94 @@
+// CompilerDriver_ClangTidy.cpp
+//------------------------------------------------------------------------------
+
+// Includes
+//------------------------------------------------------------------------------
+#include "CompilerDriver_ClangTidy.h"
+
+// FBuild
+#include "Tools/FBuild/FBuildCore/FBuild.h"
+#include "Tools/FBuild/FBuildCore/Graph/CompilerNode.h"
+#include "Tools/FBuild/FBuildCore/Graph/ObjectNode.h"
+#include "Tools/FBuild/FBuildCore/Helpers/Args.h"
+#include "Tools/FBuild/FBuildCore/WorkerPool/Job.h"
+
+// CONSTRUCTOR
+//------------------------------------------------------------------------------
+CompilerDriver_ClangTidy::CompilerDriver_ClangTidy() = default;
+
+// DESTRUCTOR
+//------------------------------------------------------------------------------
+CompilerDriver_ClangTidy::~CompilerDriver_ClangTidy() = default;
+
+// ProcessArg_PreprocessorOnly
+//------------------------------------------------------------------------------
+/*virtual*/ bool CompilerDriver_ClangTidy::ProcessArg_PreprocessorOnly( const AString & /*token*/,
+                                                                       size_t & /*index*/,
+                                                                       const AString & /*nextToken*/,
+                                                                       Args & /*outFullArgs*/ ) const
+{
+    ASSERT(false); // ClangTidy can't be used as a preprocessor
+    return false;
+}
+
+// ProcessArg_Common
+//------------------------------------------------------------------------------
+/*virtual*/ bool CompilerDriver_ClangTidy::ProcessArg_Common( const AString & token,
+                                                             size_t & index,
+                                                             Args & outFullArgs ) const
+{
+    if ( StripToken("--config-file=", token, true) )
+    {
+        return true;
+    }
+
+    // Strip -fdiagnostics-color options because we are going to override them
+    if ( m_ForceColoredDiagnostics )
+    {
+        if ( StripToken( "-fdiagnostics-color", token, true ) ||
+             StripToken( "-fno-diagnostics-color", token ) )
+        {
+            return true;
+        }
+    }
+
+    return CompilerDriverBase::ProcessArg_Common( token, index, outFullArgs );
+}
+
+// AddAdditionalArgs_Preprocessor
+//------------------------------------------------------------------------------
+/*virtual*/ void CompilerDriver_ClangTidy::AddAdditionalArgs_Preprocessor( Args & /*outFullArgs*/) const
+{
+    ASSERT(false); // ClangTidy can't be used as a preprocessor
+}
+
+// AddAdditionalArgs_Common
+//------------------------------------------------------------------------------
+/*virtual*/ void CompilerDriver_ClangTidy::AddAdditionalArgs_Common( bool /*isLocal*/,
+                                                                    Args & outFullArgs ) const
+{
+    if ( m_ForceColoredDiagnostics )
+    {
+        outFullArgs += " -fdiagnostics-color=always";
+    }
+}
+
+// AddPreliminaryArgs
+//------------------------------------------------------------------------------
+/*virtual*/ void CompilerDriver_ClangTidy::AddPreliminaryArgs( bool /*isLocal*/,
+                                                         Args & outFullArgs ) const
+{
+    // clang-tidy accepts two sets of arguments, its linter arguments and optional
+    // compilation arguments forwarded to the underlying clang instance
+    // typical usage is:
+    // $ clang-tidy [linter_arguments...] -- [clang_arguments...]
+    // --config-file being a linter argument we rely on, we make sure it is always
+    // passed before clang arguments
+
+    outFullArgs += "--config-file=";
+    outFullArgs += m_OverrideSourceFile;
+    outFullArgs += ".config.yaml";
+    outFullArgs.AddDelimiter();
+}
+
+//------------------------------------------------------------------------------
--- a/Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.h
+++ b/Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.h
@@ -1,0 +1,40 @@
+// CompilerDriver_ClangTidy.h
+//------------------------------------------------------------------------------
+#pragma once
+
+// Includes
+//------------------------------------------------------------------------------
+#include "CompilerDriverBase.h"
+
+// Core
+#include "Core/Containers/Array.h"
+#include "Core/Env/Assert.h"
+#include "Core/Process/Process.h"
+
+// Forward Declarations
+//------------------------------------------------------------------------------
+
+// CompilerDriver_ClangTidy
+//------------------------------------------------------------------------------
+class CompilerDriver_ClangTidy : public CompilerDriverBase
+{
+public:
+    CompilerDriver_ClangTidy();
+    virtual ~CompilerDriver_ClangTidy() override;
+
+    virtual bool ProcessArg_PreprocessorOnly( const AString & token,
+                                              size_t & index,
+                                              const AString & nextToken,
+                                              Args & outFullArgs ) const override;
+    virtual bool ProcessArg_Common( const AString & token,
+                                    size_t & index,
+                                    Args & outFullArgs ) const override;
+
+    virtual void AddAdditionalArgs_Preprocessor( Args & outFullArgs ) const override;
+    virtual void AddAdditionalArgs_Common( bool isLocal,
+                                           Args & outFullArgs ) const override;
+    virtual void AddPreliminaryArgs( bool isLocal,
+                                     Args& outFullArgs) const override;
+};
+
+//------------------------------------------------------------------------------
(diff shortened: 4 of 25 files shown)


ffulin:
**> This PR is only a proof of concept.**

Thanks for the PR - I took a quick look over it. I think the approach generally makes sense (adding a new compiler driver seems like a good call given that various special cases that are needed).

I think to turn this proof of concept into something that could be integrated we'd need to solve some issues.

**Critical Changes**

1.  Cross-platform
The PR is currently Windows only. We'd need to make it cross-platform. I didn't see anything that would appear to be too problematic there.
2. Portability
Related to the above, the tests assume a very specific installation on Windows using VS2019. We'd need to make that a bit more flexible, esp considering CI builds.
3. Breaking change to network protocol
This change would force a worker updated for all users by breaking the distributed protocol. We'd need to make this change backwards compatible.

**Other Implementation notes**

There are some observations of the current implementation:

1. GetFullPath
I'm not sure why this is needed. Eliminating this would resolve one of the cross-platform issues
2. CompilerDriver interface
The PR adds a new function that automatically adds clang-tidy args. Args that can be managed unconditionally are usually preferred to be set in the bff.
The PR implemented various functions that unconditionally assert, which seems a little odd.
3. Tests
The tests appear to set preprocessor options, but that shouldn't be necessary and makes it look as though the tests themselves would fail (as per the asserts mentioned above)
The tests don't appear to cover caching. They also don't seem to cover non-distributed compilation.
4. Line number fixup
The line number fixup code seems like it would be simplified and seems like it might be unsafe
5. Clang-tidy detection
The detection of use of clang tidy is duplicated and is inconsistent (case sensitive in one place, but not in another)
6. clang-tidy config file/param
Exposing the clang-tidy config as an explicit parameter on all object lists seems like it'd be nice to avoid.
 
**Coding Style**
As a final note, coding style issues/inconsistencies are easily resolved, but the more there are, the more time consuming they are to fix. After any functional changes are locked down, and style conformance work would make the code more easily mergeable.

Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +294,21 @@
     outRelativeFileName += pathB;
 }
 
+// GetFullPath
+//------------------------------------------------------------------------------
+/*static*/ AString PathUtils::GetFullPath( const AString & fileName )
```
ffulin, line 299: FASTBuild uses an output param style for returning complex types (rather than relying on return by value optimizations)
i.e.
```
void PathUtils::GetFullPath( const AString & filename, AString & outFilename )
```

This also allows use of stack types like AStackString to eliminate memory allocations.

(Note that I think we'd probably want to try and eliminate the need for this function entirely if possible, avoiding the portability issue of the unimplemented TODO)

Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriverBase.cpp
```diff
@@ -134,6 +134,13 @@ void CompilerDriverBase::Init( const ObjectNode * objectNode,
 {
 }
 
+// AddPreliminaryArgs_Common
```
ffulin, line 137: Function comment header and various functions are mismatched

Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriverBase.cpp
```diff
@@ -134,6 +134,13 @@ void CompilerDriverBase::Init( const ObjectNode * objectNode,
...
+//------------------------------------------------------------------------------
+/*virtual*/ void CompilerDriverBase::AddPreliminaryArgs( bool /*isLocal*/,
+                                                         Args & /*outFullArgs*/ ) const
+{
```
ffulin, line 141: Generally FASTBuild tries not to automatically add or modify command line args, except where strictly necessary.

Args that are always needed usually are specified in the bff file by the user.

Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.cpp
```diff
@@ -0,0 +1,94 @@
...
+CompilerDriver_ClangTidy::~CompilerDriver_ClangTidy() = default;
+
+// ProcessArg_PreprocessorOnly
+//------------------------------------------------------------------------------
+/*virtual*/ bool CompilerDriver_ClangTidy::ProcessArg_PreprocessorOnly( const AString & /*token*/,
+                                                                       size_t & /*index*/,
+                                                                       const AString & /*nextToken*/,
+                                                                       Args & /*outFullArgs*/ ) const
+{
+    ASSERT(false); // ClangTidy can't be used as a preprocessor
```
ffulin, line 30: Seems odd to implement these functions but assert.

The tests also seem to set preprocessor options, which seems at odds with this.

Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.h
```diff
@@ -0,0 +1,40 @@
...
+#pragma once
+
+// Includes
+//------------------------------------------------------------------------------
+#include "CompilerDriverBase.h"
+
+// Core
+#include "Core/Containers/Array.h"
+#include "Core/Env/Assert.h"
+#include "Core/Process/Process.h"
```
ffulin, line 12: Doesn't seem like Array, Assert or Process includes are needed.

Code/Tools/FBuild/FBuildCore/Graph/FileNode.cpp
```diff
@@ -66,6 +69,205 @@ void FileNode::HandleWarningsClangCl( Job * job, const AString & name, const ASt
...
+                        {
+                            ++cursor;
+                        }
+                        const char* filenameEnd = cursor;
+
+                        filename.Assign(filenameBegin, filenameEnd);
+                    }
+                }
+
+				startOfLine = false;
```
ffulin, line 141: Seems like tab characters have snuck in a few places

Code/Tools/FBuild/FBuildCore/Graph/FileNode.cpp
```diff
@@ -66,6 +69,205 @@ void FileNode::HandleWarningsClangCl( Job * job, const AString & name, const ASt
...
+
+			++cursor;
+        }
+
+        sourceLoc.lineNum = sourceLoc.lineNum - (lastIncludeLine + 1) + numberOfLastInclude;
+        PathUtils::FixupFilePath( filename );
+        sourceLoc.filename = filename;
+    }
+
+    return;
```
ffulin, line 152: return redundant here

Code/Tools/FBuild/FBuildCore/Graph/FileNode.cpp
```diff
@@ -66,6 +69,205 @@ void FileNode::HandleWarningsClangCl( Job * job, const AString & name, const ASt
...
+{
+    if ( data.IsEmpty() )
+    {
+        return;
+    }
+
+    AString newWarningMessage;
+    // Avoid the client who wants to print remote work to process again the error line number
+    // This part of the function workaround this issue : https://github.com/llvm/llvm-project/issues/62405
+    if ( ( job->GetDistributionState() == Job::DIST_NONE || job->GetDistributionState() == Job::DIST_BUILDING_LOCALLY ) )
```
ffulin, line 167: double () here

Probably intended:
```
if ( ( job->GetDistributionState() == Job::DIST_NONE ) ||
    ( job->GetDistributionState() == Job::DIST_BUILDING_LOCALLY ) )
```

Code/Tools/FBuild/FBuildCore/Graph/FileNode.cpp
```diff
@@ -66,6 +69,205 @@ void FileNode::HandleWarningsClangCl( Job * job, const AString & name, const ASt
...
+
+            if (diagnosticBegin != nullptr)
+            {
+                fixupLineIndices.Append(i);
+
+                // Skip column number
+                const char* cursor = diagnosticBegin - 1; // skip starting ':'
+                while (*cursor >= '0' && *cursor <= '9')
+                {
+                    --cursor;
```
ffulin, line 198: I'm not sure if this is possible, but if the first list began with ": error:" (or similar) this could read before the start of the buffer.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -77,6 +78,7 @@ REFLECT_NODE_BEGIN( ObjectNode, Node, MetaNone() )
     REFLECT( m_AllowDistribution,                   "AllowDistribution",                MetaOptional() )
     REFLECT( m_AllowCaching,                        "AllowCaching",                     MetaOptional() )
     REFLECT_ARRAY( m_CompilerForceUsing,            "CompilerForceUsing",               MetaOptional() + MetaFile() )
+    REFLECT( m_ClangTidyConfigurationFile,          "ClangTidyConfigurationFile",       MetaOptional() + MetaFile() )
```
ffulin, line 81: This will add a string to every object file, whether using clang tidy or not.

I wonder if there's a reasonable way to avoid this.

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -1937,6 +1948,26 @@ void ObjectNode::TransferPreprocessedData( const char * data, size_t dataSize, J
...
+            return;
+        }
+        extraBuffer.SetLength( (uint32_t)configFile.GetFileSize() );
+        if ( configFile.ReadBuffer( extraBuffer.Get(), extraBuffer.GetLength() ) != extraBuffer.GetLength() )
+        {
+            FLOG_ERROR( "Error: could not read file at '%s'", clangTidyConfigureFilePath.Get() );
+            return;
+        }
+        job->SetExtraDataSize( extraBuffer.GetLength() );
+        configFile.Close();
```
ffulin, line 1969: explicit close isn't needed here (FileStream closes when it falls out of scope)

Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -2343,6 +2419,23 @@ bool ObjectNode::CompileHelper::SpawnCompiler( Job * job,
...
+    if ( m_HandleOutput )
+    {
+        if (job->IsLocal())
+        {
+            isCompilerClangTidy = compilerNode->GetCompilerFamily() == CompilerNode::CompilerFamily::CLANG_TIDY;
+        }
+        else
+        {
+            const AString& file = job->GetToolManifest()->GetFiles()[0].GetName();
+            if (file.EndsWith("clang-tidy") || file.EndsWith("clang-tidy.exe"))
```
ffulin, line 2432: This duplicates detection of clang tidy elsewhere and is inconsistent (it's case sensitive here)

Code/Tools/FBuild/FBuildCore/Protocol/Protocol.h
```diff
@@ -31,7 +31,7 @@ namespace Protocol
     enum : uint16_t { PROTOCOL_PORT = 31264 }; // Arbitrarily chosen port
 
     // Protocol Version
-    enum : uint32_t { PROTOCOL_VERSION_MAJOR = 22 };    // Changes here make workers incompatible
+    enum : uint32_t { PROTOCOL_VERSION_MAJOR = 22 + 1 };    // Changes here make workers incompatible
```
ffulin, line 34: This will force all users to update all their workers, which is burdensome for some users with large build farms and multiple dev teams operating on different versions.

A change like this needs to be made in a backwards compatible. way.

Code/Tools/FBuild/FBuildCore/ExeDrivers/Compiler/CompilerDriver_ClangTidy.cpp
ffulin: I see later on that we're using a different compile for the preprocessor.

I think we'll need to make sure this works without needing to do that. If that's required for distributed compilation (assuming clang-tidy can't function as a preprocessor?) we'll have to come up with a good way to make it clear that extra manual config is needed if you want distributed clang-tidy.

Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
```diff
@@ -373,6 +375,35 @@ void TestDistributed::WarningsAreCorrectlyReported_Clang() const
...
+{
+    FBuildTestOptions options;
+    options.m_ConfigFile = "Tools/FBuild/FBuildTest/Data/TestDistributed/WarningsAreCorrectlyReported/fbuild.bff";
+    options.m_AllowDistributed = true;
+    options.m_NumWorkerThreads = 1;
+    options.m_NoLocalConsumptionOfRemoteJobs = true; // ensure all jobs happen on the remote worker
+    options.m_AllowLocalRace = false;
+    options.m_ForceCleanBuild = true;
+    options.m_EnableMonitor = true; // make sure monitor code paths are tested as well
+    options.m_DistributionPort = Protocol::PROTOCOL_TEST_PORT;
```
ffulin, line 390: PROTOCOL_TEST_PORT is already set to this so doesn't need to be set again.

Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
```diff
@@ -373,6 +375,35 @@ void TestDistributed::WarningsAreCorrectlyReported_Clang() const
...
+    options.m_ShowCommandLines = true;
```
ffulin, line 391: don't need ShowCommandLines for normal execution (left over from debugging?)

Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
```diff
@@ -373,6 +375,35 @@ void TestDistributed::WarningsAreCorrectlyReported_Clang() const
...
+//------------------------------------------------------------------------------
+void TestDistributed::WarningsAreCorrectlyReported_ClangTidy() const
+{
+    FBuildTestOptions options;
+    options.m_ConfigFile = "Tools/FBuild/FBuildTest/Data/TestDistributed/WarningsAreCorrectlyReported/fbuild.bff";
+    options.m_AllowDistributed = true;
+    options.m_NumWorkerThreads = 1;
+    options.m_NoLocalConsumptionOfRemoteJobs = true; // ensure all jobs happen on the remote worker
+    options.m_AllowLocalRace = false;
+    options.m_ForceCleanBuild = true;
```
ffulin, line 388: ForceCleanBuild not needed - this test doesn't save build state so is always clean.

Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
```diff
@@ -373,6 +375,35 @@ void TestDistributed::WarningsAreCorrectlyReported_Clang() const
     TEST_ASSERT( GetRecordedOutput().Find( "warning: unused variable 'x' [-Wunused-variable]" ) );
 }
 
+// WarningsAreCorrectlyReported_Clang
+//------------------------------------------------------------------------------
```
ffulin, line 379: This seems to only test disitributed compilation, but doesn't test local compilation.

avudnez:
Thanks for the review. I will try to improve the PR, maybe when v1.12 comes out and we upgrade to it.

ffulin:
Although this change is pretty well focused already, one thing we could do to try and make progress on this is to see if we can break apart the change.

The only big opportunity for this I see is to set aside distributed compilation support for the initial integration. That would allow the complexities of the network protocol backwards compatibility to be tackled independently.

**> when v1.12 comes out**
v1.12 was released just a few days ago

Issue #1015, opened 2024-01-18
Error in clang-cl cache


poljak181:
Hi!
It seems that I've found a bug while using cache with the **Clang-cl** compiler.
So, I've prepared a minimal reproducible example to demonstrate this error.
I used FastBuild 1.08, Clang version 15.0.7, on Windows.

Steps to reproduce:
1. Unzip the [fastbuild_clangcl_cache_pch_obj.zip](https://github.com/fastbuild/fastbuild/files/13975852/fastbuild_clangcl_cache_pch_obj.zip) into your workspace.
2. Open the **fbuild.bff** file and update the variables at the beginning according to your environment.
3. Open cmd in the directory containing the **fbuild.bff** file.
4. Do not forget to set the **FASTBUILD_CACHE_PATH** environment variable.
5. Run the command **FBuild.exe -cache -cacheverbose -showcmds -clean**. It should be OK.
6. Remove object files from the **out** directory.
7. Run the command **FBuild.exe -cache -cacheverbose -showcmds -clean** again and see something like this: `lld-link: error: could not open 'C:\workspace\test\fastbuild_clangcl_cache_pch_obj/out/main_pch.pch.obj': no such file or directory.`

This is possible fix: [fix_clang_cl_cache.patch](https://github.com/fastbuild/fastbuild/files/13975874/fix_clang_cl_cache.patch)

Pull request: https://github.com/fastbuild/fastbuild/pull/1016

**NOTE:**
There is an ability to break the build even after the fix is applied. If we only remove a.exe and main_pch.pch.obj in step 6 from **out** directory and run the command without the -clean option: `FBuild.exe -cache -cacheverbose -showcmds` we will get the same error as in step 7.
As I understand it, this happens because main_pch.pch still exists, and FastBuild doesn't check the existence of main_pch.pch.obj.

ffulin:
Thanks for the reports and the linked PR. I've added some notes on the PR.

poljak181:
I have updated the PR

ffulin:
I've applied your fix in https://github.com/fastbuild/fastbuild/commit/990be48d701edd058d9a4e878abba280bc131107

This will be in the next release (v1.13). I'll keep this issue open until that version is released.

Thanks again!

ffulin:
v1.13 has now been released and includes this fix. Thanks again for reporting the bug.

PR #1016, opened 2024-01-18
Fix clang-cl cache according to https://github.com/fastbuild/fastbuild/issues/1015

poljak181:
Fix cache for clang-cl compiler

# Description:

This is a fix of the problem described in https://github.com/fastbuild/fastbuild/issues/1015

# Checklist:

The pull request:
- [x] **Is self-contained**

**NOTE:** Compilation checked on Windows only

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/ICache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/ICache.cpp
@@ -16,7 +16,7 @@
                                     AString & outCacheId )
 {
     // cache version - bump if cache format is changed
-    const char cacheVersion( 'E' );
+    const char cacheVersion( 'F' );
 
     // format example: 2377DE32AB045A2D_FED872A1_AB62FEAA23498AAC-32A2B04375A2D7DE.7
     outCacheId.Format( "%016" PRIX64 "_%08X_%016" PRIX64 "-%016" PRIX64 ".%c",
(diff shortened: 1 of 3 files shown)


ffulin:
This seems pretty reasonable, but also requires an increment to this:
```
    const char cacheVersion( 'E' );
```
to ensure existing cache entries are not assumed to include this file.

ffulin:
It would be ideal to extend the tests (such as TestPCHClangWindowsWithCache) to ensure they cover this case.

poljak181:
Thanks for the comment, the fix added!

ffulin:
I realized your PR is against the main branch instead of the dev branch (main is the stable release, while dev is the development branch). Since the change was fairly simple, I re-applied your fix (rather than asking you to create a new PR): https://github.com/fastbuild/fastbuild/commit/990be48d701edd058d9a4e878abba280bc131107

In that commit, I also extended the test coverage a bit, including making the test actually link the object files built into an executable, which without your fix gives a link error (so we have another way to detect any future regression).

Thanks for the fix! This change will be in the next release (v1.13)

poljak181:
> I realized your PR is against the main branch instead of the dev branch (main is the stable release, while dev is the development branch). Since the change was fairly simple, I re-applied your fix (rather than asking you to create a new PR): [990be48](https://github.com/fastbuild/fastbuild/commit/990be48d701edd058d9a4e878abba280bc131107)
> 
> In that commit, I also extended the test coverage a bit, including making the test actually link the object files built into an executable, which without your fix gives a link error (so we have another way to detect any future regression).
> 
> Thanks for the fix! This change will be in the next release (v1.13)

That was my mistake to create PR based on main branch. Thanks for re-applying fix!

PR #1029, opened 2024-03-26
Files which are each other's substrings are sorted in Unity builds

mkawalec:
# Description:

Without this change Unity builds will not sort files where one is a prefix of another. For instance for the following set of files
```
/file.hpp
/file.h
/file
```
The order will be indeterminate as Unity builds would only check for equality up to the length of the shortest filename when comparing pairs of filenames.

# Checklist:

The pull request:
- [x] **Is self-contained**


ffulin:
Thanks for the PR. It's certainly the intent that the files in a folder be deterministically ordered and perhaps this logic isn't working quite right.

ffulin:
This PR has been integrated and will be in the next release (v1.14).

Thanks for the fix!

PR #1046, opened 2024-07-09
Fix solution version written in the sln when using vs2022, so explorer shows vs2022 icon and vs version selector opens vs2022 and not vs2019

belkiss:
Same as in VS2017.bff and VS2019.bff, we only write the major version in the field


ffulin:
Thanks for the fix!

PR #1049, opened 2024-07-10
Fix warning C26462 treated as error in x64-Analyze with latest MSVC, two pointers can be made const

belkiss:
# Description:

Small compile fix for latest MSVC 17.10.3.


ffulin:
Thanks for the fix!

PR #1050, opened 2024-07-10
Copy worker exe to tmp dir instead of next to the executable to allow running it from a NAS

belkiss:
# Description:

Worker exe copy is by default copied next to the original exe, which makes it impossible to store the worker on a NAS for instance and have users run it from there.

This change copies instead the worker to the temp directory, which fixes that use case.

# Checklist:

The pull request:
- [X] **Is self-contained**


belkiss:
Please hold, missed a part while recreating this branch to prepare the PR, a little too hasty :D

ffulin:
I think this is a desirable change, but as implemented this will cause all users to have their settings reset because the settings are stored side-by-side with the executable by default.

For users where I know how the worker is being used, the worker is submitted to source control rather than being run from the network (which allows workers to be easily auto-updated).

I think it's necessary to add some logic to retrieve the settings from the old location if not found in the new location in order to not invalidate user settings.

belkiss:
Will do!

PR #1052, opened 2024-07-27
Add support for vs2022 to Github Actions CI

belkiss:
# Description:

This fixes the paths so the Github Actions CI also run the builds, tests, and analyze with vs2022.

Note that it relies on the fix #1049

# Checklist:

The pull request:
- [X] **Is self-contained**

### DIFF
--- a/.github/workflows/Windows.yml
+++ b/.github/workflows/Windows.yml
@@ -14,11 +14,17 @@ jobs:
       matrix:
         include:
           - cfg: Build & Tests
+            vs: 2022
+            os: windows-2022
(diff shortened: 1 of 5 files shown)


ffulin:
Thanks for the updates. I've integrated this and it's now active on the CI.
As a followup, I've:
* made all versions of VS2022 use the same version (to match what GitHub CI is using)
* extended the VS2022 config so it can detect both "Community" and "Enterprise" versions
* deleted support for building FASTBuild with VS2017 to reduce future maintenance

belkiss:
Excellent, thanks!

PR #1054, opened 2024-08-07
Dev

romanova-ekaterina:
This is the first small patch containing some utility functions needed for contributing Distributed ThinLTO support for FASTBuild.
It contains several simple functions for working with strings and file path names.

This is my first contribution to FASTBuild. The patch is intentionally very small, that's a good start to understand requirements for committing to FASTBuild project.

### DIFF
--- a/Code/Core/FileIO/IOStream.h
+++ b/Code/Core/FileIO/IOStream.h
@@ -6,10 +6,7 @@
 //------------------------------------------------------------------------------
 #include "Core/Containers/Array.h"
 #include "Core/Env/Types.h"
-
-// Forward Declarations
-//------------------------------------------------------------------------------
-class AString;
+#include "Core/Strings/AString.h"
 
 // IStream
 //------------------------------------------------------------------------------
@@ -60,6 +57,35 @@ public:
     // manage memory-mapped aligned data
     void AlignRead( size_t alignment );
     void AlignWrite( size_t alignment );
+
+    inline uint64_t WriteStringRaw( const AString& s ) { return WriteBuffer( s.begin(), s.GetLength() ); }
+
+   // Write several strings into IOStream.
+   //------------------------------------------------------------------------------
+   template < typename T0 = AString, typename... T >
+   bool WriteV( const T0& part0, const T&... parts )
+   {
+       bool rc = true;
+       rc = rc && Write( part0 );
+       if constexpr( sizeof...( parts ) > 0 ) {
+           rc = rc && WriteV( parts... );
+       }
+       return rc;
+   }
+
+   // Write several raw strings into IOStream (raw - means no length information will be stored in the stream)
+   //------------------------------------------------------------------------------
+   template < typename T0 = AString, typename... T >
+   uint64_t WriteRawV( const T0& part0, const T&... parts )
+   {
+       uint64_t rc = 0;
+       rc += WriteStringRaw( part0 );
+       if constexpr( sizeof...( parts ) > 0 ) {
+           rc += WriteRawV( parts... );
+       }
+       return rc;
+   }
+   //------------------------------------------------------------------------------
 };
 
 // Read ( Array< T > )
@@ -91,33 +117,4 @@ bool IOStream::Write( const Array< T > & a )
     return true;
 }
 
-//------------------------------------------------------------------------------
-
-inline uint64_t WriteStringRaw( const AString& s ) { return WriteBuffer( s.begin(), s.GetLength() ); }
-
-// Write several strings into IOStream.
-//------------------------------------------------------------------------------
-template < typename T0 = AString, typename... T >
-bool WriteV( const T0& part0, const T&... parts )
-{
-    bool rc = true;
-    rc = rc && Write( part0 );
-    if constexpr( sizeof...( parts ) > 0 ) {
-        rc = rc && WriteV( parts... );
-    }
-    return rc;
-}
-
-// Write several raw strings into IOStream (raw - means no length information will be stored in the stream)
-//------------------------------------------------------------------------------
-template < typename T0 = AString, typename... T >
-uint64_t WriteRawV( const T0& part0, const T&... parts )
-{
-    uint64_t rc = 0;
-    rc += WriteStringRaw( part0 );
-    if constexpr( sizeof...( parts ) > 0 ) {
-        rc += WriteRawV( parts... );
-    }
-    return rc;
-}
 //------------------------------------------------------------------------------
--- a/Code/Core/FileIO/IOStream.h
+++ b/Code/Core/FileIO/IOStream.h
@@ -92,3 +92,32 @@ bool IOStream::Write( const Array< T > & a )
 }
 
 //------------------------------------------------------------------------------
+
+inline uint64_t WriteStringRaw( const AString& s ) { return WriteBuffer( s.begin(), s.GetLength() ); }
+
+// Write several strings into IOStream.
+//------------------------------------------------------------------------------
+template < typename T0 = AString, typename... T >
+bool WriteV( const T0& part0, const T&... parts )
+{
+    bool rc = true;
+    rc = rc && Write( part0 );
+    if constexpr( sizeof...( parts ) > 0 ) {
+        rc = rc && WriteV( parts... );
+    }
+    return rc;
+}
+
+// Write several raw strings into IOStream (raw - means no length information will be stored in the stream)
+//------------------------------------------------------------------------------
+template < typename T0 = AString, typename... T >
+uint64_t WriteRawV( const T0& part0, const T&... parts )
+{
+    uint64_t rc = 0;
+    rc += WriteStringRaw( part0 );
+    if constexpr( sizeof...( parts ) > 0 ) {
+        rc += WriteRawV( parts... );
+    }
+    return rc;
+}
+//------------------------------------------------------------------------------
--- a/Code/Core/FileIO/PathUtils.cpp
+++ b/Code/Core/FileIO/PathUtils.cpp
@@ -290,4 +290,39 @@
     outRelativeFileName += pathB;
 }
 
+// GetBaseName
 //------------------------------------------------------------------------------
+
+/* static */ void PathUtils::GetBaseName( const AString& pathName, AString& baseName )
+{
+    const char* slashPos = nullptr;
+    if ( nullptr != ( slashPos = pathName.FindLast( NATIVE_SLASH ) ) )
+        baseName = slashPos + 1;
+    else
+        baseName = pathName.Get();
+}
+
+// GetDirectoryName
+//------------------------------------------------------------------------------
+
+/* static */ void PathUtils::GetDirectoryName( const AString& pathName, AString& dirName )
+{
+    const char* slashPos = nullptr;
+    if ( pathName.EndsWith( NATIVE_SLASH ) )
+    {
+        dirName = pathName;
+        return;
+    }
+    if ( nullptr != ( slashPos = pathName.FindLast( NATIVE_SLASH ) ) )
+        dirName = AString( pathName.begin(), slashPos + 1 );
+    else
+        dirName = AString::GetEmpty();
+}
+
+// JoinPath
+//------------------------------------------------------------------------------
+
+/* static */ void PathUtils::JoinPath( AString& pathName, const AString& baseName ) {
+    EnsureTrailingSlash( pathName );
+    pathName += baseName;
+}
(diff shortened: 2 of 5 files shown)


Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +290,39 @@
     outRelativeFileName += pathB;
 }
 
+// GetBaseName
 //------------------------------------------------------------------------------
+
+/* static */ void PathUtils::GetBaseName( const AString& pathName, AString& baseName )
+{
+    const char* slashPos = nullptr;
+    if ( nullptr != ( slashPos = pathName.FindLast( NATIVE_SLASH ) ) )
```
ffulin, line 299: The two lines above line could be simplified to:
```
if ( const char * slashPos = pathName.FindLast( NATIVE_SLASH ) )
```

Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +290,39 @@
...
+        baseName = slashPos + 1;
+    else
+        baseName = pathName.Get();
```
ffulin, line 302: Both baseName and pathName are AStrings, so this can be:
```
basetName = pathName;
```

Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +290,39 @@
...
+
+/* static */ void PathUtils::GetDirectoryName( const AString& pathName, AString& dirName )
+{
+    const char* slashPos = nullptr;
+    if ( pathName.EndsWith( NATIVE_SLASH ) )
+    {
+        dirName = pathName;
+        return;
+    }
+    if ( nullptr != ( slashPos = pathName.FindLast( NATIVE_SLASH ) ) )
```
ffulin, line 316: As above, the slashPos variable can be combined into this line:
```
if ( const char * slashPos = pathName.FindLast( NATIVE_SLASH ) )
```

Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +290,39 @@
...
+        dirName = AString( pathName.begin(), slashPos + 1 );
```
ffulin, line 317: AString has an Assign method that can be used to avoid a memory allocation and string copy here:
```
dirName.Assign( pathName.Get(), slashPos + 1 ); // Include slash
```

Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +290,39 @@
...
+    else
+        dirName = AString::GetEmpty();
```
ffulin, line 319: AString has a Clear() function which can be used to clear a string:
```
dirName.Clear();
```

Code/Core/FileIO/PathUtils.h
```diff
@@ -51,6 +51,9 @@ class PathUtils
     static void GetRelativePath( const AString & basePath,
                                  const AString & fileName,
                                  AString & outRelativeFileName );
+    static void GetBaseName( const AString& pathName, AString& baseName );
+    static void GetDirectoryName( const AString& pathName, AString& dirName );
+    static void JoinPath( AString& pathName, const AString& baseName );
```
ffulin, line 56: It would be nice to include some tests for these core functions.

I can help with those, but if you want to get familiar with the test framework (which we'll certainly want to utilize for the more complex functionality) then these trivial tests would be a good place to start.

Code/Core/Math/Random.cpp
```diff
@@ -24,4 +26,15 @@ uint32_t Random::GetRand()
     return( (uint32_t)( m_Seed / ( ( CORE_RAND_MAX + 1 ) * 2 ) ) % ( CORE_RAND_MAX + 1 ) );
 }
 
+// GetRandString
+//------------------------------------------------------------------------------
+uint32_t Random::GetRandString( AString& str )
+{
+    uint32_t rand_num = GetRand();
+    AStackString<128> buf;
+    snprintf( buf.Get(), buf.GetLength(), "%x", rand_num );
```
ffulin, line 35: This is a buffer overflow, writing to the "no mans land" of the AStackString.

AString has string formatting functionality built in, so you can replace the 2 lines above and the one below with:
```
str.Format( "%x", randNum );
```

Code/Core/Math/Random.cpp
```diff
@@ -24,4 +26,15 @@ uint32_t Random::GetRand()
     return( (uint32_t)( m_Seed / ( ( CORE_RAND_MAX + 1 ) * 2 ) ) % ( CORE_RAND_MAX + 1 ) );
 }
 
+// GetRandString
+//------------------------------------------------------------------------------
```
ffulin, line 30: I don't think this function belongs in the Random class.
It formats a random number in a way that is likely not very general purpose and can be trivially implemented with a single line in the calling code (as shown below) or with 2 lines if you need the "randNum".

Code/Core/Math/Random.cpp
```diff
@@ -24,4 +26,15 @@ uint32_t Random::GetRand()
...
+uint32_t Random::GetRandString( AString& str )
```
ffulin, line 31: Minor: FASTBuild's style includes a space before the &.

I can easily fix these during integration, but it'll be easier if that's not needed.

Code/Core/Math/Random.cpp
```diff
@@ -24,4 +26,15 @@ uint32_t Random::GetRand()
...
+{
+    uint32_t rand_num = GetRand();
```
ffulin, line 33: FASTBuild's coding style uses camel case, so this variable would usually be called "randNum".

This variable could also be const.

I can of course fix minor issues like this, but the fewer there are the easier things will be.

Code/Core/FileIO/IOStream.h
```diff
@@ -60,6 +57,35 @@ class IOStream
     // manage memory-mapped aligned data
     void AlignRead( size_t alignment );
     void AlignWrite( size_t alignment );
+
+    inline uint64_t WriteStringRaw( const AString& s ) { return WriteBuffer( s.begin(), s.GetLength() ); }
+
```
ffulin, line 62: The WriteStringRaw function above seems quite reasonable, but it's unclear to me how the templates below would be leveraged.

Usually when dealing with multiple items, you'd keep them in an Array and could write those with a simple loop already.

Adding a function to take an Array<T> as a convenience doesn't seem unreasonable, but I'm hesitant to use a template to "unroll" a loop in this way.

Do you have an example of the intended use case for this you could share?

Code/Core/FileIO/IOStream.h
```diff
@@ -60,6 +57,35 @@ class IOStream
...
+       return rc;
+   }
+
+   // Write several raw strings into IOStream (raw - means no length information will be stored in the stream)
+   //------------------------------------------------------------------------------
+   template < typename T0 = AString, typename... T >
+   uint64_t WriteRawV( const T0& part0, const T&... parts )
+   {
+       uint64_t rc = 0;
+       rc += WriteStringRaw( part0 );
```
ffulin, line 82: The function takes a template param for the type, but since it calls WriteStringRaw, it could only ever be used with an AString, so the template param seems unnecessary.

Code/Core/FileIO/IOStream.h
```diff
@@ -60,6 +57,35 @@ class IOStream
...
+       if constexpr( sizeof...( parts ) > 0 ) {
```
ffulin, line 83: minor: FASTBuild's standard always balances the braces

Code/Core/FileIO/PathUtils.cpp
```diff
@@ -290,4 +290,39 @@
...
 }
 
+// GetBaseName
 //------------------------------------------------------------------------------
+
+/* static */ void PathUtils::GetBaseName( const AString& pathName, AString& baseName )
+{
+    const char* slashPos = nullptr;
+    if ( nullptr != ( slashPos = pathName.FindLast( NATIVE_SLASH ) ) )
+        baseName = slashPos + 1;
```
ffulin, line 300: minor: FASTBuild's standard never omits braces for if statements.

ffulin:
Thanks for the first PR! Please let me know if I can clarify any of my comments or questions.

Code/Core/FileIO/PathUtils.cpp
romanova-ekaterina: Fixed

Code/Core/FileIO/PathUtils.cpp
romanova-ekaterina: Fixed.

Code/Core/FileIO/PathUtils.cpp
romanova-ekaterina: Fixed.

Code/Core/FileIO/PathUtils.cpp
romanova-ekaterina: Fixed.

Code/Core/FileIO/PathUtils.cpp
romanova-ekaterina: Fixed

Code/Core/FileIO/PathUtils.h
romanova-ekaterina: I will do it. Is there any documentation/description on how to run the tests?

Code/Core/Math/Random.cpp
romanova-ekaterina: I deleted the entire function (as suggested in your next comment).

Code/Core/Math/Random.cpp
romanova-ekaterina: I deleted the entire function

Code/Core/Math/Random.cpp
romanova-ekaterina: Done here and in a few more places.

Code/Core/Math/Random.cpp
romanova-ekaterina: I removed this entire function following your earlier suggestion.

Code/Core/FileIO/IOStream.h
romanova-ekaterina: Done.

Code/Core/FileIO/PathUtils.cpp
romanova-ekaterina: Done.

Code/Core/FileIO/IOStream.h
romanova-ekaterina: It was done to implement a variadic function (i.e. a function that can take a variable number of parameters). For example sometimes this function is called with 4 parameters, sometimes with 2 parameters:

mapFileStream.WriteRawV(GetThinltoModuleId(), AString("\t"), sourceFilePath, AString("\n"));
mapFileStream.WriteRawV(dependencyFile, AString("\t"));

Usually 2 methods are implement variadic function: via using a std::initializer_list or via variadic templates (which was done in this case). 

Alternatively, we could call Write function several times (either several consequent calls or in a loop). I checked in the code, we always invoke WriteV function with 2 to 4 arguments. But this will bloat the code at the call site.

Let me know which solution do you prefer




i

Code/Core/FileIO/IOStream.h
romanova-ekaterina: A template (more specifically, a variadic template) was only used here to create functions that take variable number of arguments (see my comment above). As you correctly noted, it could only be used with AString.

Code/Core/FileIO/IOStream.h
romanova-ekaterina: I wanted to ping you about which solution do you prefer, taking into consideration that this function could be invoked with 2-4 parameters (all of them of the AString type.
E.g. 
mapFileStream.WriteRawV(GetThinltoModuleId(), AString("\t"), sourceFilePath, AString("\n"));
or
mapFileStream.WriteRawV(dependencyFile, AString("\t"));

See all different options in the previous comment. 

I'm sorry I was away on vacation for more than 3 weeks, but I am back, and I will start submitting PR more actively now. However, I prefer to wrap up a first "utility" PR first.

Code/Core/FileIO/PathUtils.h
romanova-ekaterina: I sent tests for the the functions above as a separate followup commit. Let me know if this is something needs to be changed.

romanova-ekaterina:
> Thanks for the first PR! Please let me know if I can clarify any of my comments or questions.

I think I did all the changes that you have requested. I would like to send the next patch, but I suspect this "introductory" patch needs to get merged first, right?

Issue #1030, opened 2024-03-28
 integration Distributed ThinLTO with FASTBuild


romanova-ekaterina:
Hello,

Note: This is an RFC, not patch.

We would like to propose implementation of a new functionality in FASTBuild that will allow its seamless integration with Distributed ThinLTO technology (from now on referred to as DTLTO).

Introduction to ThinLTO and DTLTO:

ThinLTO (Thin “Link Time Optimization” Framework) has been implemented in LLVM for number of years, supported by several linkers, including lld, gold, ld64 and gnu linker, and widely used to achieve a better runtime performance.

In brief, ThinLTO enables scalable and incremental whole program optimization.
Here are presentations from the author, Teresa Johnson, explaining ThinLTO framework in general:
https://www.youtube.com/watch?v=p9nH2vZ2mNo
https://www.youtube.com/watch?v=VAMvr1rXmg8

Here is documentation related to ThinLTO:
https://clang.llvm.org/docs/ThinLTO.html

ThinLTO can be performed in distributed fashion as well. This technology is called Distributed ThinLTO (DTLTO). Though DTLTO has been supported in LLVM for several years, it is not widely used, because of multiple constraints. High-level build system such as Bazel from Google and Buck2 from Facebook have built-in support DTLTO and can handle the additional tasks required to bypass these limitations. Regrettably, unless a user is prepared to restructure their build projects for Bazel or Buck, employing DTLTO technology is not feasible for substantial projects designed for makefiles, ninja, autotool, cmake, and so on.
Here at Sony, we recently implemented “Integrated Distributed ThinLTO” framework, that will allow to use DTLTO coupled with many different build/distribution systems, including but not limited to DistCC, IceCream, FASTBuild, Incredibuild. 

For additional information about DTLTO Framework integrated with build/distribution systems please refer to our slides from LLVM developers conference.
https://discourse.llvm.org/uploads/short-url/A2a3LQ12a2Wzy75ZY36OB2ueOeJ.pdf

Currently, integrated DTLTO Framework is coupled with Sony’s proprietary distributing build system, SN-DBS. This code is in production, well tested by our internal CI test environment, and actively used for game development by Sony’s toolchain users.
We would like to expand the scope of our project and add DTLTO support to the FASTBuild. To do this, several new features must be added to FASTBuild.

Here is the list of problems that needs to be solved for DTLTO integration with the FastBuild. 

[...]

PR #1067, opened 2025-03-11
Add support for C++ Dynamic Debugging introduced in Visual Studio 17.14 Preview 2

xiangfan-ms:
# Description:
Add support for C++ Dynamic Debugging introduced in Visual Studio 17.14 Preview 2.
I'm a developer from Microsoft Visual C++ team. We are working on the [C++ Dynamic Debugging](https://aka.ms/vcdd)  feature which was [announced](https://aka.ms/dynamicdebugging) recently.
We worked with various build engines to add support for this feature. This PR allows FastBuild to support various files introduced by the feature. This initial change disables distributed build if the feature is enabled to keep backward compatibility (old remote worker won't copy the extra files back).
If you have any questions, we can discuss the details here or via email.

# Checklist:

The pull request:
- [ x ] **Is created against the Dev branch**
- [ x ] **Is self-contained**
I built on Windows only.
Does the validation pipeline build on other platforms? The added code is C++ standard conformant and doesn't rely on a specific OS or compiler to compile. So, I don't expect any build issue.
The feature requires 17.14 Preview 2. Looks that the test environment used by the repo only has 17.10.
We tested the change in an internal fork. Let me know how you want to proceed.
- [ x ] **Passes existing tests**
- [ x ] **Keeps Windows, OSX and Linux at parity**
- [ x ] **Follows the code style**
We can discuss how we want to document this support.

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Cache/ICache.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Cache/ICache.cpp
@@ -16,7 +16,7 @@
                                     AString & outCacheId )
 {
     // cache version - bump if cache format is changed
-    const char cacheVersion( 'G' );
+    const char cacheVersion( 'H' );
 
     // format example: 2377DE32AB045A2D_FED872A1_AB62FEAA23498AAC-32A2B04375A2D7DE.7
     outCacheId.Format( "%016" PRIX64 "_%08X_%016" PRIX64 "-%016" PRIX64 ".%c",
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1036,13 +1036,16 @@ bool ObjectNode::ProcessIncludesWithPreProcessor( Job * job )
         // 3) pch files can't be built from preprocessed output (disabled acceleration), so can't be distributed
         // 4) user only wants preprocessor step executed
         // 5) Distribution of /analyze is not currently supported due to preprocessor/_PREFAST_ inconsistencies
-        // 6) Source mappings are not currently forwarded so can only compiled locally
+        // 6) Distribution of /dynamicdeopt is not currently supported because old remote workers won't know to
+        //    return the extra obj files.
+        // 7) Source mappings are not currently forwarded so can only compiled locally
         if ( !usingCLR && !usingPreprocessorOnly )
         {
             if ( isDistributableCompiler &&
                  !usingWinRT &&
                  !( flags.IsCreatingPCH() )&&
                  !( flags.IsUsingStaticAnalysisMSVC() ) &&
+                 !( flags.IsUsingDynamicDeopt() ) &&
                  !hasSourceMapping )
             {
                 flags.Set( CompilerFlags::FLAG_CAN_BE_DISTRIBUTED );
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1024,6 +1024,10 @@ bool ObjectNode::ProcessIncludesWithPreProcessor( Job * job )
                     flags.Set( CompilerFlags::FLAG_STATIC_ANALYSIS_MSVC );
                 }
             }
+            else if ( IsStartOfCompilerArg_MSVC( token, "dynamicdeopt" ) )
+            {
+                flags.Set( CompilerFlags::FLAG_DYNAMIC_DEOPT );
+            }
         }
 
         // 1) clr code cannot be distributed due to a compiler bug where the preprocessed using
@@ -1324,6 +1328,20 @@ void ObjectNode::GetGCNOPath( AString & gcnoFileName ) const
     const char * extPos = m_Name.FindLast( '.' ); // Only last extension removed
     gcnoFileName.Assign( m_Name.Get(), extPos ? extPos : m_Name.GetEnd() );
     gcnoFileName += ".gcno";
+}
+
+// GetAltObjPath
+//------------------------------------------------------------------------------
+void ObjectNode::GetAltObjPath( AString& altObjName ) const
+{
+    ASSERT( IsUsingDynamicDeopt() );
+
+    const AString & sourceName = m_PCHObjectFileName.IsEmpty() ? m_Name : m_PCHObjectFileName;
+
+    // TODO:B The suffix ('.alt') can be manually specified with /dynamicdeopt:suffix
+    const char * extPos = sourceName.FindLast( '.' ); // Only last extension removed
+    altObjName.Assign( sourceName.Get(), extPos ? extPos : sourceName.GetEnd() );
+    altObjName += ".alt.obj";
 }
 
 // GetObjExtension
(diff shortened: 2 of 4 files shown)


ffulin:
Thanks for the PR - this seems like an exciting feature. You may have seen that FASTBuild already support a similar featue, allowing "deoptimization" of files which are writable (for "open for edit" from Perforce) or from a provided list (which can be populated from Git), but having something that works more seamlessly/automatically as you debug would be very nice.

I took a look at the PR and it seems good overall, but there are a few things we'll need to modify:

**1) Cache Version**
The PR correctly adds the additional file to GetExtraCacheFilePaths, but this will alter the payload of cached objects which means old cache files will not contain the new file and new versions will crash trying to read this mismatched payload.
The cache version can be freely bumped between releases so the fix is easy. We just need to bump:
```
    // cache version - bump if cache format is changed
    const char cacheVersion( 'G' );
```
in ICache.cpp.

**2) Distributed Compilation**
Distributed compilation will fail if this new option is present on the command line because remote workers won't know to return the new file. Since this is essentially already broken, it's not an issue with this PR, but we'll want to solve for that.

The initial thing we should do is simply flag the presence of this option as inhibiting distributed compilation (as is already done for various other things) in:
```
ObjectNode::DetermineFlags
```
In the longer term, to continue to support distributed compilation, more changes will be needed, but perhaps those can be addressed later so as to allow this initial change to be integrated (and when adding the disabling the current version will be safe).

That will require a little non-trivial work because FBuildWorkers and FASTBuild clients must be both forwards and backwards compatible. We'll need to have new clients that understand this option only distribute work to workers with the same minor protocol version or later. Similarly, we'll need to make new workers that understand the option not return the new file as part of the payload unless the client is a compatible version or later. Some other functionality already has similar plumbing for things like this, so similar plumbing is possible but it takes a little work to get right.

Again, I think we can just disable distribution initially.

**3) reference to c2dd.dll in VS2022,bff**

The changes to VS2022.bff reference a dll that doesn't exist in old versions of Visual Studio. Just as you mentioned that you can't write effective tests yet, we can't update this file for the same reason.

We'll need to omit this file until FASTBuild can be built against a release of the new Visual Studio that supports this features, including it being available on GitHub.

This won't inhibit users from using this feature as they will have their own Compiler() configs, with their own set of files which they could add this to. (Technically omitting the file won't prevent local compilation anyway, we'd just be missing the dependency should that dll change without others changing, which isn't likely)

-----

Will any more work be require for this to be integrated into the VS IDE, or is everything automatic as long as the objects have been compiled with the appropriate flags and a new VS IDE is used?

xiangfan-ms:
Thanks for the review!
I updated the PR to address the review feedback.

>Will any more work be require for this to be integrated into the VS IDE, or is everything automatic as long as the objects have been compiled with the appropriate flags and a new VS IDE is used?

AFAIK, FastBuild only caches .obj files (the feature generates other .alt files). As long as the extra .obj files are correctly cached / distributed, the feature will work automatically when you build with the appropriate flags and use the new VS debugger.

xiangfan-ms:
I also updated the PR description to reflect the new changes.

ffulin:
Thanks for the updated PR! The changes have been integrated and support for /dynamicdeopt will be included in the next version (v1.15).

I will look into exactly what might be required to allow distributed compilation support while retaining worker compatibility.

xiangfan-ms:
Thanks a lot!

ffulin:
I've made some followup changes: https://github.com/fastbuild/fastbuild/commit/cc5a1e27f937f013737f01fb05cde2f997a44604. These will also be in the next release (v1.15)

**Detection of VS2022 Preview version**
The VS2022.bff now detects the Preview version (either vendorized or in the default install location) and adds the necessary DLL for distributed compilation. The setup is a little clunky, but one this VS version becomes the non-Preview version, the added complexity can be remove. 

**Add test for Caching**
I added a test to cover the caching cases.

**Add support for Distributed Compilation**
I added the necessary code for remote workers to similarly bundle the additional file and return it to a host. Compilation tasks using the /dynamicopt flag will only be passed to workers that are capable of building them and returning the additional file (as controlled by the worker protocol minor version, keeping fwds/backwards compat). Some future work might rework this logic to make it easier to extend for future cases. Tests are also added for /dynamicdeopt to ensure the extra object is returned.

xiangfan-ms:
I ran the tests locally.

Looks that 'MSVC_DYNAMIC_DEOPT_AVAILABLE' is not defined in 'Code\Tools\FBuild\FBuildTest\Data\TestCache\ExtraFiles_DynamicDeopt\fbuild.bff' and the test fails with:

`FASTBuild Error #1030 - Unknown directive 'error'.`

I don't see the check for 'MSVC_DYNAMIC_DEOPT_AVAILABLE' in 'Code\Tools\FBuild\FBuildTest\Data\TestDistributed\MSVCDynamicDeoptimization\fbuild.bff'.
And the test code is already protected by the following before accessing the test data.

`#if defined( _MSC_VER ) && ( _MSC_VER >= 1944 ) // VS 2022 17.44.x`

Any ideas on why the test fails?

The tests can pass (and I confirm they will fail if I change '.alt' to '.alt1') if I remove the `#if` in fbuild.bff.

BTW, 'GetNativeAnalysisXMLPath' has similar inconsistent style (it uses 'AString&').

ffulin:
**> Looks that 'MSVC_DYNAMIC_DEOPT_AVAILABLE' is not defined in**
It's possible the path being checked in the VS2022.bff doesn't match the exact version you have. If so, then the version would not be detected and MSVC_DYNAMIC_DEOPT_AVAILABLE would not be defined.

**> I don't see the check for 'MSVC_DYNAMIC_DEOPT_AVAILABLE' in 'Code\Tools\FBuild\FBuildTest\Data\TestDistributed\MSVCDynamicDeoptimization\fbuild.bff'**
This check is not strictly required as the test is ultimately guarded by the _MSC_VER check. It was however intended to make failures a bit more explcit. The two tests are inconsistent, but as they are temporarily guarded, I'm not super worried about that.

**> Any ideas on why the test fails?**
If VS2022 is not being correctly detected and MSVC_DYNAMIC_DEOPT_AVAILABLE is not defined as a result, the .ExtraFiles of the Compiler() will be missing the new dll and I would expect tests for distributed compilation to fail.

**> BTW, 'GetNativeAnalysisXMLPath' has similar inconsistent style (it uses 'AString&').**
Some existing code has code style issues, missed during initial integration or implementation. Those will very likely be fixed in a future compliance pass. For now, the aim is to limit the introduction of new style variations when integrating new changes.

Holmes98:
> I've made some followup changes: [cc5a1e2](https://github.com/fastbuild/fastbuild/commit/cc5a1e27f937f013737f01fb05cde2f997a44604). These will also be in the next release (v1.15)

I think this change breaks backwards compatibility with older workers when *not* using /dynamicdeopt?
https://github.com/fastbuild/fastbuild/commit/cc5a1e27f937f013737f01fb05cde2f997a44604#diff-abbae728d19e2e6802040aee7896646baf10dacf5503c634baa00930903cabffR463-R478

ffulin:
@Holmes98 Yes, you're right. I've applied a fix here: https://github.com/fastbuild/fastbuild/commit/64b3b515e6efbf588493a986d7a7cfde861ec1cc

That fix will be part of the v1.15 release.

Thanks for letting me know about this bug.

PR #1070, opened 2025-04-15
Add new UseDeterministicPaths_Experimental flag to CompilerNode

jblazquez:
# Description:

This PR is a first step towards supporting deterministic builds.

The key issue that this change solves is the source filenames that Clang and other compilers embed in object files. When building a preprocessed file, FASTBuild used to create a temporary directory with a name that depended on the CPU core that was executing the build step, resulting in different build outputs across runs. With this change, the user can ask for those temporary directories to have a deterministic name that is based on the hash of the compiler command line.

No new tests are added because it wasn't clear how to test this functionality in an automated way.

# Checklist:

The pull request:
- [X] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -1382,26 +1382,7 @@ const AString & ObjectNode::GetCacheName( Job * job ) const
 
     // hash the build "environment"
     // TODO:B Exclude preprocessor control defines (the preprocessed input has considered those already)
-    uint32_t commandLineKey;
-    {
-        Args args;
-        const bool useDeoptimization = false;
-        const bool showIncludes = false;
-        const bool useSourceMapping = false; // Source mapping compiler flags contain local paths, so we treat them specially
-        const bool finalize = false; // Don't write args to response file
-        BuildArgs( job, args, PASS_COMPILE_PREPROCESSED, useDeoptimization, showIncludes, useSourceMapping, finalize );
-
-        if ( job->IsLocal() )
-        {
-            // Append the source mapping destination only, so different machines with different
-            // working directory local paths compute consistent keys.
-            const AString & sourceMapping = job->GetNode()->CastTo<ObjectNode>()->GetCompiler()->GetSourceMapping();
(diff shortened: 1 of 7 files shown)


Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
```diff
@@ -1382,26 +1382,7 @@ const AString & ObjectNode::GetCacheName( Job * job ) const
 
     // hash the build "environment"
     // TODO:B Exclude preprocessor control defines (the preprocessed input has considered those already)
-    uint32_t commandLineKey;
```
jblazquez, line 1385: Extracted into its own function with no changes so we can call it elsewhere.

Code/Tools/FBuild/FBuildTest/Tests/TestDistributed.cpp
```diff
@@ -306,7 +306,7 @@ void TestDistributed::ErrorsAreCorrectlyReported_Clang() const
     TEST_ASSERT( false == fBuild.Build( "ErrorsAreCorrectlyReported-Clang" ) );
 
     // Check that error is returned
-    TEST_ASSERT( GetRecordedOutput().Find( "fatal error: expected ';' at end of declaration" ) );
```
jblazquez, line 309: Clang 20 no longer says "fatal", just "error".

ffulin:
Thanks for the change. This is certainly desirable and I think should be the default behavior. Using an Experimental opt-in flag is a great choice as we can isolate any issues that might crop up.

I think as long as every object with a unique combination of cmd line args including output params is in a unique folder, this should be safe and the further separation of the CPU cores should be unnecessary.

The only case I can think of as potentially being problematic is if additional side-car files are emitted and those are somehow pathed upwards via relative paths on the cmd line. I don't think we could do anything about that, or detect it in a reasonable way, and I think you'd almost have to be trying to hurt yourself to do that, so I'm not super worried about that (and in essence that problem already existed anyway).

*>No new tests are added because it wasn't clear how to test this functionality in an automated way.*

I think what we can do is enable this option in the testcommon.bff for FASTBuild and then it will apply to all tests. I think that makes sense given I think it should be the default behavior. I can make this change.

PR #1071, opened 2025-04-15
Address non-determinism in temporary object file hash suffix

jblazquez:
# Description:

This is a followup change to #1070. There was one more source of nondeterminism related to the name of temporary object files. These object files are suffixed with the XXH32 hash of their source file paths, but these source file paths can vary across machines. The `.UseRelativePaths_Experimental` option can be used to allow for cache hits across machines with different workspace roots, but different machines could generate different object files due to the different names they'd use for this temporary object file. While probably unlikely to cause issues, I think it would be ideal if we can guarantee that different machines generate bit-identical outputs for the same inputs, even if the workspace root differs.

Using the *relative path* to the source file when assembling the temporary object filename is another step towards different machines generating bit-identical outputs.

Currently, this change only applies the relativization of source paths when running the job locally, because that's when we have access to the `FBuild` object to get the working dir. It's not clear to me whether anything needs to be done for remote scenarios, as those paths tend to be consistent across machines as far as I understand.

# Checklist:

The pull request:
- [X] **Is self-contained**
- [ ] **Has accompanying tests** (functionality is exercised by existing tests)
- [ ] **Includes documentation** (no documentation needed, since it's not a user-facing feature)

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/ObjectNode.cpp
@@ -2120,7 +2120,21 @@ bool ObjectNode::WriteTmpFile( Job * job, AString & tmpDirectory, AString & tmpF
     ASSERT( job->GetData() && job->GetDataSize() );
 
     const Node * sourceFile = GetSourceFile();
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the improvement! This change will be in the next release (v1.15)

PR #1083, opened 2025-07-28
Fix JobQueueRemote::ReadResults error handling.

webbju:
# Description:

- Prior to [`466a542`](https://github.com/fastbuild/fastbuild/commit/466a542c3be2e42ae2063234d9787f356002b9f8) `JobQueueRemote::ReadResults` would [return early](https://github.com/fastbuild/fastbuild/commit/466a542c3be2e42ae2063234d9787f356002b9f8#diff-87b2490a001cfb6e450a672dd355ba2261b510de59dfaeeb420fd8f2ce328047L465) if filesystem errors occurred. This PR reinstates the behaviour.
- Shortly after bumping to `v1.15` we encountered various problems related to corrupted output on remote workers (see snippet below). In the process of diffing which code-change may have caused this I noticed that this particular function was missing an early return. Currently we've reverted to using `v1.14` with no issues. Doubt that this PR fixes the corruption but felt like making sure it was upstreamed regardless as it seems to be an oversight.

```
27>ld.lld: error: /build/games/Games/Engine/Plugins/[redacted]/Intermediate/Build/Linux/x64/UnrealEditor/Development/[redacted]/amd_l_defaults.c.o:1: unknown directive: 8z
  >>> 8z27>Execution failed. Error: 1 (0x01) Target: '/build/games/Games/Engine/Plugins/[redacted]/Binaries/Linux/libUnrealEditor-EmbarkRBD.so'
```

# Checklist:

The pull request:
- [x] **Is self-contained**
- [x] **Includes documentation** (n/a)

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueueRemote.cpp
+++ b/Code/Tools/FBuild/FBuildCore/WorkerPool/JobQueueRemote.cpp
@@ -483,6 +483,7 @@ void JobQueueRemote::FinishedProcessingJob( Job * job, Node::BuildResult result 
     {
         job->Error( "Error reading file: '%s'", fileNames[ problemFileIndex ].Get() );
         FLOG_ERROR( "Error reading file: '%s'", fileNames[ problemFileIndex ].Get() );
+        return false;
     }
 
     // Compress result


ffulin:
Thanks for the fix! This has been integrated and will be in the next release (v1.16)

If this case was being hit on a remote worker, it would have been returning invalid results to the host which conceivably could potentially cause invalid data to be written locally. Additionally it's possible the cache would be poisoned with invalid results. In addition to integrating your fix, I've bumped the cache version (https://github.com/fastbuild/fastbuild/commit/f372f4df3192eb0ef6b541cbd14890ce02d8872e)

**> Currently we've reverted to using v1.14 with no issues. Doubt that this PR fixes the corruption ...**

I'm not sure I can reliably speak to the behavior you've seen from the context given alone, but perhaps at some point your cache has been poisoned by v1.15 (assuming you are using the cache) and switching back to v1.14 isolates you from that (as v1.15 contained a cache version change).

On the other hand, if you are reliably seeing failures with causing that link issue, I'd welcome any additional info you can share about your setup or compile time params that might be relevant. Running with v1.15 but with your fix would be ideal to ensure any problems this code is supposed to catch are caught at their origin of course.

PR #1107, opened 2025-09-23
Fix Unity build with UseRelativePaths_Experimental

Noxybot:
# Description:
The following BFF fails to compile without this fix:
```
  Unity('test_unity')
  {
    .UnityOutputPath = 'out_unity'
    .UnityOutputPattern = 'test_unity.cpp'
    .UnityInputFiles = 'cpps/source.cpp'
    .UseRelativePaths_Experimental = true
  }

  ObjectList ('test_obj')
  {
    .Compiler = '/usr/bin/clang++'
    .CompilerOptions = '-o %2 -c %1'
    .CompilerOutputPath = 'out_obj'
    .CompilerInputUnity = 'test_unity'
  }
```

The issues is that the formed `test_unity.cpp` looks like this:
```
// Auto-generated Unity file - do not modify

#pragma message( "cpps/source.cpp" )
#include "cpps/source.cpp"
```
And  `FBuild test_obj` fails with:
```
5> Uni: test_unity                                             
2> Obj: /tmp/out_obj/test_unity.o
2> Failed to build Object. Error: 1 (0x01) Target: '/tmp/out_obj/test_unity.o'
2> PROBLEM: /tmp/out_obj/test_unity.o
/tmp/out_unity/test_unity.cpp:3:9: warning: cpps/source.cpp [-W#pragma-messages]
    3 | #pragma message( "cpps/source.cpp" )
      |         ^
/tmp/out_unity/test_unity.cpp:4:10: fatal error: 'cpps/source.cpp' file not found
    4 | #include "cpps/source.cpp"
      |          ^~~~~~~~~~~~~~~~~
1 warning and 1 error generated.
FBuild: Error: BUILD FAILED: test_obj                          
Time: 0.040s
```

With the fix the generated `test_unity.cpp` looks like this:
```
// Auto-generated Unity file - do not modify

#pragma message( "../cpps/source.cpp" )
#include "../cpps/source.cpp"
```
And compiles fine:
```
4> Uni: test_unity                                             
9> Obj: /tmp/out_obj/test_unity.o
9> WARNING: /tmp/out_obj/test_unity.o
/tmp/out_unity/test_unity.cpp:3:9: warning: ../cpps/source.cpp [-W#pragma-messages]
    3 | #pragma message( "../cpps/source.cpp" )
      |         ^
1 warning generated.
FBuild: OK: test_obj                                           
Time: 0.080s
```

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestUnity.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestUnity.cpp
@@ -919,13 +919,13 @@ void TestUnity::CacheUsingRelativePaths() const
         }
 
         // Check __FILE__ paths are relative
-        // Slash direction changed in Clang 18.x.x from forward slash to backslash
-        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_1(./Subdir/Header.h)FILE_MACRO_END_1" ) ||
-                     buffer.Find( "FILE_MACRO_START_1(.\\Subdir/Header.h)FILE_MACRO_END_1" ) );
 #if defined( __WINDOWS__ )
-        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_2(.\\File.cpp)FILE_MACRO_END_2" ) );
+        TEST_ASSERT(
+            buffer.Find( "FILE_MACRO_START_1(..\\out\\..\\Code\\Subdir/Header.h)FILE_MACRO_END_1" ) );
+        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_2(..\\out\\..\\Code\\File.cpp)FILE_MACRO_END_2" ) );
 #else
-        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_2(./File.cpp)FILE_MACRO_END_2" ) );
+        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_1(./Subdir/Header.h)FILE_MACRO_END_1" ) );
+        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_2(../out/../Code/File.cpp)FILE_MACRO_END_2" ) );
 #endif
     }
 
--- a/Code/Tools/FBuild/FBuildCore/Graph/UnityNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/UnityNode.cpp
@@ -359,7 +359,7 @@ UnityNode::~UnityNode()
     AStackString includeBasePath;
     if ( m_UseRelativePaths_Experimental )
     {
-        includeBasePath = FBuild::Get().GetOptions().GetWorkingDir();
+        includeBasePath = m_OutputPath;
         PathUtils::EnsureTrailingSlash( includeBasePath );
     }
 


Dandielo:
I've investigated a bit in regards to this, it seems that the problem is with the `UnityOutputPath` property from the `Unity` function.
I've not thought about this too much yet, but maybe it helps.

You can reproduce this easily on Win too:
root/
├─ fbuild.bff
├─ out_unity/ (generated)
│   ├─ test_unity.cpp (#include "cpps/source")
├─ cpps/
│   ├─ source.cpp


At least in this case, the resulting unity file does not take into consideration the output path when generating the `#includes`.

If I set the 'UnityOutputPath' to '.' then the `test_unity.cpp` obviously gets generated in the `root/` however, what was not expected for me to see is that the `.obj` file was also there, even though the `ObjectList` function has it's `CompilerOutputPath` set to `out_obj`. 

This resulted in another issue, where fastbuild failed to find the expected object file.

```
BFF file 'asd\test\fbuild.bff' has changed (reparsing will occur).
1> Obj: asd\test\out_obj\test_unity.obj
1> WARNING: asd\test\out_obj\test_unity.obj
1> File missing despite success for 'asd\test\out_obj\test_unity.obj'
FBuild: Error: BUILD FAILED: all
Time: 0.248s
```

ffulin:
I initially misunderstood your directory layout.
This is the correct way to specify the UnityInputFiles:
```
    .UnityInputFiles = 'cpps/source.cpp'
```
(relative to the working directory)

When I compile your test in Windows, I get this fatal error:
```
C:\Test\Unity\out_unity\test_unity.cpp(4): fatal error C1034: cpps\source.cpp: no include path set
```
which hints at the problem.

If I change the command line as follows the test passes:
```
    .CompilerOptions = '/Fo%2 -c %1 /I.'
```
(note the -o was also replaced with /Fo for msvc compat and removing the space fixes the "file missing despite success" error)

Adding the include path as the root of the working directory means the same paths can be used for includes as when you specified files in the bff.

However, I think that's more a work-around than a fix. I think your proposed PR to change the relative path base dir to be the location of the generated unity file is the right one. The test failure you had seems to be because a particular test is assuming the specific current behavior, so it might just be a case of needing to update that test.

Code/Tools/FBuild/FBuildTest/Tests/TestUnity.cpp
```diff
@@ -919,13 +919,21 @@ void TestUnity::CacheUsingRelativePaths() const
...
         // Check __FILE__ paths are relative
-        // Slash direction changed in Clang 18.x.x from forward slash to backslash
-        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_1(./Subdir/Header.h)FILE_MACRO_END_1" ) ||
-                     buffer.Find( "FILE_MACRO_START_1(.\\Subdir/Header.h)FILE_MACRO_END_1" ) );
 #if defined( __WINDOWS__ )
-        TEST_ASSERT( buffer.Find( "FILE_MACRO_START_2(.\\File.cpp)FILE_MACRO_END_2" ) );
+        TEST_ASSERTM(
+            buffer.Find( "FILE_MACRO_START_1(..\\out\\..\\Code\\Subdir/Header.h)FILE_MACRO_END_1" ),
+            "File content:\n%s",
+            buffer.Get() );
```
ffulin, line 926: The buffer is binary data, so I don't think we want to print out the output here.

I guess that was useful for tracking down the problem if you don't have access to all the environments, but I think it'd strange to print all manner of non-printable characters.

PR #1119, opened 2025-10-22
Avoid hashing large compiler files for CMake "Try Compiles"

billhoffman:
# Description:
### Add `-fasthash` option to speed up CMake configure times

**Problem:**
The `cmake -GFASTBuild` configure step is significantly slower than other generators like Ninja. This is because CMake performs numerous "TryCompile" checks. Each check causes FASTBuild to read and hash the full contents of the entire compiler toolchain, resulting in a major I/O bottleneck.

**Solution:**
This change introduces a new command-line option, **`-fasthash`**.

When this flag is used, FASTBuild skips reading and hashing the content of toolchain files (like `cl.exe`, `link.exe`, etc.). Instead, it checks only the file's **timestamp**. If the file has changed, it generates a new "fake" hash based on the file's **name**, not its content.

This avoids the expensive I/O and makes the "TryCompile" checks almost instantaneous.

**Performance Impact:**
This option is intended to be used by the CMake generator during the configure/TryCompile steps.

Here is a comparison of configure times for the same project:

| Generator | Command | Time |
| :--- | :--- | :--- |
| Ninja | `cmake -GNinja` | 27.024s |
| **FASTBuild (Before)** | `cmake -GFASTBuild` | **55.936s** |
| **FASTBuild (After)** | `cmake -GFASTBuild` (with generator passing `-fasthash`) | **33.765s** |

As shown, this change (when used by the generator) cuts the configure time significantly, bringing it much closer to the performance of Ninja.

# Checklist:

The pull request:
- [ x] **Is created against the Dev branch**
- [ x] **Is self-contained**
- [x ] **Compiles on Windows, OSX and Linux**
- [ x] **Passes existing tests**
- [x ] **Keeps Windows, OSX and Linux at parity**
- [ x] **Follows the code style**
- [ x] **Includes documentation**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/CompilerNode.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Graph/CompilerNode.cpp
@@ -369,7 +369,8 @@ CompilerNode::~CompilerNode()
 //------------------------------------------------------------------------------
 /*virtual*/ Node::BuildResult CompilerNode::DoBuild( Job * /*job*/ )
 {
-    if ( !m_Manifest.DoBuild( m_StaticDependencies ) )
+    const bool skipHashing = ( !m_AllowCaching && !m_AllowDistribution );
+    if ( !m_Manifest.DoBuild( m_StaticDependencies, skipHashing ) )
     {
         return BuildResult::eFailed; // Generate will have emitted error
     }
--- a/Code/Tools/FBuild/FBuildCore/Helpers/ToolManifest.cpp
+++ b/Code/Tools/FBuild/FBuildCore/Helpers/ToolManifest.cpp
@@ -109,7 +109,7 @@ void ToolManifestFile::StoreCompressedContent( const void * uncompressedData, co
 
 // DoBuild
 //------------------------------------------------------------------------------
-bool ToolManifestFile::DoBuild()
+bool ToolManifestFile::DoBuild( bool skipHashing )
 {
     // Name should be set
     ASSERT( m_Name.IsEmpty() == false );
@@ -118,6 +118,30 @@ bool ToolManifestFile::DoBuild()
     ASSERT( m_CompressedContent == nullptr );
     ASSERT( m_CompressedContentSize == 0 );
 
+    if ( skipHashing )
+    {
+        // Get the file's timestamp (a fast 'stat' call)
+        const uint64_t lastWriteTime = FileIO::GetFileLastWriteTime( m_Name );
+
+        // Check if our 'fake' hash is already up to date
+        if ( ( m_Hash != 0 ) && ( m_TimeStamp == lastWriteTime ) )
+        {
+            return true; // Nothing to do
+        }
+
+        // File missing?
+        if ( lastWriteTime == 0 )
+        {
+            FLOG_ERROR( "Error: opening file '%s' in Compiler ToolManifest. File not found.\n", m_Name.Get() );
+            return false;
+        }
+
+        // File has "changed" or is new, re-calculate our "fake" hash
+        // We hash the NAME instead of the CONTENT.
+        m_Hash = xxHash::Calc32( m_Name );
+        m_TimeStamp = lastWriteTime;
+        return true; // We're done, skip the expensive file I/O below
+    }
     // Do we already have a hash?
     if ( m_Hash != 0 )
     {
@@ -182,7 +206,7 @@ void ToolManifest::Initialize( const AString & mainExecutableRoot, const Depende
 
 // Generate
 //------------------------------------------------------------------------------
-bool ToolManifest::DoBuild( const Dependencies & dependencies )
+bool ToolManifest::DoBuild( const Dependencies & dependencies, bool skipHashing )
 {
     ASSERT( m_Files.GetSize() == dependencies.GetSize() );
     (void)dependencies;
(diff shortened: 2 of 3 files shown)


ffulin:
Regarding the hash used, the individual file hashes are actually persisted to remote workers as well (in addition to being used to form the ToolChainId) so a little extra work would be required to keep worker compatibility.

billhoffman:
> It can also be a hidden / undocumented environment variable... This way we can set it on the generator side and it will just start magically work faster once users update FASTBuild executable.

I really like this idea as I have run into this breaking the build when I get the wrong fbuild that does not have the flag.  I can rework the PR to do that.

billhoffman:
I tested this with the env var and was able to get it to work

This is the change:
```
-    if ( FBuild::Get().GetOptions().m_FastHash )
+    AStackString tmp;
+    if ( Env::GetEnvVariable( "FASTBUILD_SKIP_TOOLCHAIN_HASH", tmp ))
     {
         // Get the file's timestamp (a fast 'stat' call)
         const uint64_t lastWriteTime = FileIO::GetFileLastWriteTime( m_Name );
```
I was able to get the CMake side of things to work.  Again, this would only be used in the special case of the cmake try_compile not for "real" projects just in the setup of "real" projects.

billhoffman:
@ffulin does this make sense?  Should I update the PR to use the env var approach?  It really makes a big difference in the configure times.  In some cases it is 2x slower without doing this. It should not affect actual builds as it won't be used for them.

Code/Tools/FBuild/FBuildCore/Helpers/ToolManifest.cpp
```diff
@@ -118,6 +118,34 @@ bool ToolManifestFile::DoBuild()
...
+        }
+
+        // File has "changed" or is new, re-calculate our "fake" hash
+        // We hash the NAME instead of the CONTENT.
+        m_Hash = xxHash::Calc32( m_Name );
+        m_TimeStamp = lastWriteTime;
+        m_UncompressedContentSize = 0; // We don't know or care
+        m_CompressedContentSize = 0;
+        FREE( m_CompressedContent );
+        m_CompressedContent = nullptr;
```
ffulin: It should be impossible to get here with compressed content.

billhoffman:
How do I get to the new flags from `CompilerNode::DoBuild` into `ToolManifestFile::DoBuild`?

ffulin:
I think something like this would be nice and tidy:

In CompilerNode::DoBuild:
```
// If we don't need to distribute or cache, we can bypass hashing file contents
const bool skipHashing = !m_AllowCaching && !m_AllowDistribution;
if ( !m_Manifest.DoBuild( m_StaticDependencies, skipHashing  ) )
```
which allows ToolManifest to pass it down to ToolManifestFile:
```
bool ToolManifest::DoBuild( const Dependencies & dependencies, bool skipHashing )
{
    ASSERT( m_Files.GetSize() == dependencies.GetSize() );
    ...
        if ( !file.DoBuild( skipHashing ) )
```

Let me know if I've overlooked something.

billhoffman:
@ffulin  does this look good now?  We have it working on the CMake side.

ffulin:
The change looks pretty good to me. Just a few minor clang formatting issues.

Code/Tools/FBuild/FBuildCore/Graph/CompilerNode.cpp
```diff
@@ -369,7 +369,12 @@ CompilerNode::~CompilerNode()
 //------------------------------------------------------------------------------
 /*virtual*/ Node::BuildResult CompilerNode::DoBuild( Job * /*job*/ )
 {
-    if ( !m_Manifest.DoBuild( m_StaticDependencies ) )
+    bool skipHashing = false;
+    if ( !m_AllowCaching && !m_AllowDistribution )
+    {
+      skipHashing = true;
+    }
```
ffulin: This is one of the bits failing clang format (bad indents).

Could all be one line as well:
```
const bool skipHashing = ( !m_AllowCaching && !m_AllowDistribution );
```

billhoffman:
> If your IDE doesn't support clang format, you can use the script Scripts/ClangFormat.py

I just did another force push and ran clang-format 19.1.7 on the code.  I think it should be good now.

billhoffman:
@ffulin  FWIW the Scripts/ClangFormat.py is pretty much windows only. I ran clang-format by hand.

ffulin:
> **FWIW the Scripts/ClangFormat.py is pretty much windows only.**

I think the location it searches for the clang-format exe can probably be extended to work on other platforms (adding /usr/bin/clang-format or similar). I'll look into that. Thanks for letting me know.

ffulin:
I've integrated the changes and they will be in the next release (v1.18). Thanks for making this improvement!

I submitted some minor followup tweaks here: https://github.com/fastbuild/fastbuild/commit/462931da132f14626e9289cd9883a156f84b7528

billhoffman:
@ffulin  thanks for merging. @Noxybot and I have published a blog about the cmake integration:
https://www.kitware.com/cmake-fastbuild-distributed-cached-and-fast/

ffulin:
@billhoffman Very nice!

I noticed that your blog post seems to revolve around v1.17 (and the timings presumably against that version).

It might be interesting to compare timings of v1.17 vs v1.18 once that is released, not only for the "try compiles" optimization we've landed in this PR, but also other optimizations coming through (including some more I just submitted today).

PR #1133, opened 2025-11-30
Add #elif directive

StrangePan:
# Description:

Add `#elif` directive. Like the C/C++ preprocessor directive, an `#elif` must follow an `#if` directive or another preceeding `#elif`, and must be followed by either an `#else` or `#endif`.

There is no limit to the number of chained `#elif` directives. Only the first `#if` or `#elif` directive in the chain whose expression resolves to `true` will have its body parsed. Subsequent chained `#elif` expressions will be checked for syntax errors, but their expressions will not be resolved (meaning that files and environment variables would not actually be tested for existence).

The `#elif` directive has all the same features as the `#if` directive because it uses the same parsing and resolving logic.

Also, to satisfy a new and more strict test, fixed that a malformed directive within an `#if` or `#else` block would produce error 1030: "Unknown directive" instead of error 1010: "Unknown construct".

# Checklist:

The pull request:
- [X] **Is self-contained**

### DIFF
--- a/.github/workflows/OSX.yml
+++ b/.github/workflows/OSX.yml
@@ -14,7 +14,7 @@ jobs:
       matrix:
         include:
           - cfg: Build & Tests
-            os: macos-13
+            os: macos-14
 
     name: OSX (${{ matrix.cfg }})
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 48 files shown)


StrangePan:
Ok, I've updated the documentation, and have formally updated errors 1046 and 1047 to support the `elif` directive in addition to `if`. If you decide against this approach, those changes should be easy to revert.

I believe this PR is ready for a proper review.

ffulin:
Thanks for the updated PR. This has been integrated and will be part of the next release (v1.19)

StrangePan:
Thanks! I look forward to updating our build files 😄

Issue #218, opened 2017-03-19
Support Visual Studio Ctrl+F7 (Compile Single File) Workflow with FastBuild .vcxproj generation


missmah:
This is a feature request to add support for the Visual Studio Ctrl+F7 (Compile Single File) Workflow with FastBuild .vcxproj generation.

Currently FastBuild .vcxproj simply generates the following XML for a .cpp file:
`<CustomBuild Include="File.cpp" />`

I believe the solution to supporting CTRL+F7 to build an individual .cpp file would involve setting up a build step for each .cpp file, with an associated command to run (e.g. FBuild.exe $(ProjectName)-$(FileName)-$(Platform)-$(Configuration)).

I believe that this per-file build step should also not run during normal builds, because those are run through the NMake Build and Rebuild command lines, though I'm not fully positive of this.

@ffulin Is this something you have on your radar? Is it something we could implement and submit a pull request to you after we're done, with a little guidance on the best way to accomplish it?

Thanks!

missmah:
It looks like in theory, something like this should work:

```
<CustomBuild Include="File.cpp"> 
      <Message>Compiling %(Identity) [%(FullPath)]</Message>  
      <Command>$(FASTBUILD_ROOT_DIR)\FBuild.exe $(ProjectName)-%(Identity)-$(Platform)-$(Configuration)</Command>  
      <Outputs>$(OutDir)%(Identity)</Outputs>       
</CustomBuild>
```

However, this seems to be ignored by Visual Studio in my preliminary tests. It's possible that VS by default makes a custom build command incompatible with an NMake project?

missmah:
So, if you make a configuration which is a 'Utility' project, instead of a 'Makefile' project, then the above works. You also need to make sure headers are flagged as C++ Headers in that case, so that the system doesn't attempt to build them. I'm tempted to make an Iteration configuration, which works this way.

In this case, file paths also need to be relative, not absolute, or else it won't work.

However, there's still a remaining snag - I can't write the following code, because FastBuild can't get all the .cpp files in the project into a string for me:

```
ForEach( .cppFile in .ProjectCppFiles )
{
	// Compile Single Module (CTRL+F7 behavior in Visual Studio)
	ObjectList( '$ProjectName$-$cppFile$-OBJ-$Platform$-$Config$' )
	{
		.CompilerOptions			+ .ConfigurationCompilerOptions
[...]

PR #1141, opened 2025-12-29
VSProject - Add support for CompileFileCommand

leo-polymorph:
# Description:
This add support for the VS 2022 NMake new "Compile File Command Line" option. This is triggered by right clicking on a file in the Solution Explorer, and then clicking on "Compile", or CTRL+F7.

The feature has previously been discussed here: https://github.com/fastbuild/fastbuild/issues/218

# Checklist:

The pull request:
- [X] **Is self-contained**
- [X] **Compiles on Windows**
- [ ] **Compiles on OSX and Linux**

### DIFF
--- a/Code/Tools/FBuild/FBuildTest/Tests/TestProjectGeneration.cpp
+++ b/Code/Tools/FBuild/FBuildTest/Tests/TestProjectGeneration.cpp
@@ -1091,9 +1091,9 @@ void TestProjectGeneration::VCXProj_ProjectRelativePaths() const
     // Check vcxproj
     {
         AStackString proj( pg.GenerateVCXProj( projectFileName, configs, Array<VSProjectFileType>(), Array<VSProjectImport>() ) );
-        TEST_ASSERT( proj.Replace( "<ClCompile Include=\"..\\ProjectSourceFiles\\File.cpp\" />", "" ) == 1 );
-        TEST_ASSERT( proj.Replace( "<ClCompile Include=\"..\\ProjectSourceFiles\\SubDir\\File.cpp\" />", "" ) == 1 );
-        TEST_ASSERT( proj.FindI( "<ClCompile " ) == nullptr );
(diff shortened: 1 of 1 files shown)


ffulin:
Thanks for the change. This all seems good to me in general. I think we should make some small tweaks though:

**Older VS compatibility**
As far as I know, there's no reason we can't use ClCompile unconditionally, but I'm concerned that there may be issues with older Visual Studio versions. I think we can avoid any risk here by conditionally using ClCompile only when .CompileFileCommand is set, and use CustomBuild as before when not.

**Remove %1 fix up**
The support for %1 I think is a little confusing, because it implies that it works along the lines of FASTBuild's %1 when passed to the compiler. I think it would be better to remove this fix up entirely, and update the documentation to suggest the use of $(SelectedFiles) directly. I think that would also make it clearer when something else is needed, such as when the relative path to projects is more complex (it seems for FASTBuild itself, something like '''tmp/VisualStudio/Projects/$(SelectedFiles)''' would be needed for example).

(see note below about this ideally being a temporary concern either way though)

**Version Bump**
The NodeGraphHeader::kCurrentVersion needs an increment due to the new serialized PROPERTY.

---

Most of this is covered in the original issue you linked, but that's got quite a bit of history on it, so while it's fresh in my mind, here are my thoughts about future steps we could take in this space.

If we added a new pair of command line options to FASTBuild (something like "-sourcefiles=$(SelectedFiles)" and "-sourcetarget=$(Project)-$(Config)-$(Platform)") then:
a) FASTBuild could support multiple selected files ($(SelectedFiles) appears to be ; delimited)
b) FASTBuild coudl support multiple configurations (the object file which is the output for the source files could be derived from the target ObjectList/Library) so Ctrl+F7 could build a given file in multiple contexts such as in different libraries or build configs
c) there would be no need for the environment variable work around (and we've avoid re-parsing the bff file when switching contexts such as compiling different files or switching from solution builds to single file builds)

ffulin:
Thanks for the updated PR. This change has been integrated and will be part of the next release (v1.19).

PR #1176, opened 2026-03-28
[Fix] Save errno before close() clobbers it in SystemMutex::TryLock

cyberkm:
# Description:                                                                                                                                          
                                                                                                                                                             
   In `SystemMutex::TryLock` on Linux, when `flock()` fails because another process holds the lock, `errno` is set to `EWOULDBLOCK`/`EAGAIN`. However   `close(handle)` is called before checking `errno`, and a successful `close()` clobbers it. The subsequent check on line 69 then reads an undefined `errno` value and may fail to detect the "already locked" condition, causing `TryLock` to incorrectly report an unexpected failure instead of returning   
    `false`.                                                                                                                                                 
                                                                                                                                                             
   The fix saves `errno` from `flock()` before calling `close()`.                                                                                            
                                                                                                                                                             
   # Checklist:                                                                                                                                              
                                                                                                                                                             
   The pull request:                                                                                                                                         
   - [x] **Is self-contained**                                                                                                                               

### DIFF
--- a/Code/Core/Process/SystemMutex.cpp
+++ b/Code/Core/Process/SystemMutex.cpp
@@ -65,8 +65,9 @@ bool SystemMutex::TryLock()
     int rc = flock( handle, LOCK_EX | LOCK_NB );
     if ( rc )
     {
+        const int flockErrno = errno;
         VERIFY( close( handle ) == 0 );
-        if ( errno == EWOULDBLOCK || errno == EAGAIN )
+        if ( flockErrno == EWOULDBLOCK || flockErrno == EAGAIN )
         {
             return false; // locked by another process
         }


ffulin:
This does look a little wrong in that the code doesn't clearly communicate the intended logic, though I don't think there is a bug here. The close() should succeed, which I believe means it would leave the errno from flock untouched. Both code paths (when the assert is removed) also return false, so functionality is unchanged.

Were you seeing this fail in some situation (and the assert firing)? As far as I can tell, that would only happen if the close() failed, or the flock had an unexpected return code, which would indicate a different problem I think. I want to make sure I'm not overlooking something here.

In terms of the current logic, I think this is probably a clearer expression of the intended logic/would remove the ambiguity:
```
int rc = flock( handle, LOCK_EX | LOCK_NB );
if ( rc )
{
    // flock should only fail due to failing to acquire the lock
    ASSERT( ( errno== EWOULDBLOCK ) || ( flockErrno == EAGAIN ) );
    return false; // locked by another process
}
```

cyberkm:
I didn't see it directly. It’s something the Clang static analyzer found, and I thought it was worth fixing just in case, even though it's a minor edge case.

ffulin:
Thanks for confirming. I agree it's good to fix/improve the code as we can remove any ambiguity about the behavior.

(btw - I have a Clang static analysis config, but have not worked through the issues yet in order to be able to activate it in CI).

ffulin:
There was a problem with my suggested alternative (I accidentally removed the close) but I've gone an applied the intended fixup here: https://github.com/fastbuild/fastbuild/commit/43e64bf0f95f81d91a217c452997bb53dd9ec656

This change will be part of the next source code release (v1.20)

Thanks for reporting the issue!

PR #1183, opened 2026-04-14
Config parsing and general optimizations

StrangePan:
# Description:

I'm working on overhauling one of my project's FastBuild config files and have encountered some unignorable performance problems while parsing the config file. Specifically, whenever dealing with large arrays, large or deeply nested structs, lengthy strings, and stack frames with large numbers of variables. The performance bottlenecks I'm encountering seem to stem from excessive deep copies of data, or from lack of proper variable indexing within a large variable scope.

So I am experimenting with optimizing FastBuild's parsing (and some of its core data structures) in order to improve config parsing time.

So far in this experimental branch, I have:
- rewritten AString and Array to share underlying data with other instances as long as no mutations are made
- rewritten UnorderedMap to use Arrays
- rewritten BFFVariable to be usable directly within Arrays
- rewritten BFFVariable operations to operate on data in-place instead of creating a copy, mutating the copy, and assigning back to the variable
- added BFFVariableScope to hold and index BFFVariables within a scope, and to only switch to using an UnorderedMap after growing past a certain size
- reduced UnorderedMap bucket count from 65,536 to 256
- probably more

As a result, parsing of BFF config files containing large structs, arrays, and strings has significantly improved. For simpler BFF config files, there are still some speed ups, but they are less noticeable.

There are still obvious shortcomings in this experimental branch. I took some shortcuts with style, conventions, and templating in order to make quick progress and test the results. I've also left some IDE files in there for my own convenience. However, I wanted to get the maintainers' input on the general direction these changes are heading so I can decide whether or not to continue.

- Are the maintainers interested in optimizing this area of the code?
- Do these optimizations seem like reasonable solutions to the parsing performance problems?
- Is the introduced complexity an acceptable tradeoff?

Thank you for your time.

# Checklist:

The pull request:
- [ ] **Is self-contained**

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,5 @@
+# Default ignored files
+/shelf/
+/workspace.xml
+# Editor-based HTTP Client requests
+/httpRequests/
--- a/.idea/codeStyles/codeStyleConfig.xml
+++ b/.idea/codeStyles/codeStyleConfig.xml
@@ -1,0 +1,5 @@
+<component name="ProjectCodeStyleConfiguration">
+  <state>
+    <option name="PREFERRED_PROJECT_CODE_STYLE" value="Factorio" />
+  </state>
+</component>
--- a/.idea/copilot.data.migration.agent.xml
+++ b/.idea/copilot.data.migration.agent.xml
@@ -1,0 +1,6 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="AgentMigrationStateService">
+    <option name="migrationStatus" value="COMPLETED" />
+  </component>
+</project>
--- a/.idea/customTargets.xml
+++ b/.idea/customTargets.xml
@@ -1,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="CLionExternalBuildManager">
+    <target id="0c8706e8-11ce-42ed-afb1-fe9d5eb0f7a8" name="fbuild CoreTest" defaultType="TOOL">
+      <configuration id="2f558642-76b9-4ec8-bf63-26d40ab2c85e" name="fbuild CoreTest">
+        <build type="TOOL">
+          <tool actionId="Tool_External Tools_fbuild CoreTest" />
+        </build>
+      </configuration>
+    </target>
+    <target id="bccd6642-1b6c-432a-a905-81a0397af4a4" name="fbuild" defaultType="TOOL">
+      <configuration id="08f47f9d-e73b-4a24-b229-7f2cdc6bb4ce" name="fbuild">
+        <build type="TOOL">
+          <tool actionId="Tool_External Tools_fbuild" />
+        </build>
+      </configuration>
+    </target>
+  </component>
+</project>
--- a/.idea/editor.xml
+++ b/.idea/editor.xml
@@ -1,0 +1,125 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="BackendCodeEditorSettings">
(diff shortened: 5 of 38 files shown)


ffulin:
**> Are the maintainers interested in optimizing this area of the code?**

Generally speaking, bff parsing performance is less important than other things (such as build-time performance) (the bff is not normally parsed during builds; only when bff files are edited). However, it's still highly desirable for it to be performant to improve the experience when editing, or for unusual work-flows that modify bff files more often than expected.

So performance gains are desirable, as long as they don't negatively impact maintainability or the performance of other operations.

To this end, there have been some optimizations to bff parsing in recent versions, including some significant gains for some scenarios in v1.20. There are a few other optimizations in flight too (not visible in github)

**> Do these optimizations seem like reasonable solutions to the parsing performance problems?**
**> Is the introduced complexity an acceptable tradeoff?**

In your branch all the changes are mixed together, and I can't tell concretely what magnitude of benefit each one brings and in what situations, so at the moment it's not possible to judge perf vs complexity trade-offs. It seems your changes could likely be decoupled without too much trouble so we could measure and talk about them fairly independently. I would probably try to start with the simplest ones first.

If you could share a bit more about your bff file structure that might help direct performance work as well.

---

I took a glance at the PR as well and have some thoughts on the various aspects you mentioned:

**> rewritten AString and Array to share underlying data with other instances as long as no mutations are made**

I have experimented with copy-on-write strings in BFFParser some time ago, but at the time it was not a performance improvement. At the time I saw regressions in some cases due to the additional overheads.
However, this was some time ago and much about the code has changed since I last tried, and the idea certainly makes sense.
I think it such an optimization would live at the BFFVariable level, rather than at AString or Array (which are generic containers). atomic operations on bff operations seem unnecessary as parsing is single threaded. Focussing on BFFVariable would also avoid adding overhead (memory and runtime) to other AString/Array uses which are prominent in many different places (not just bff parsing)

**> rewritten UnorderedMap to use Arrays**

The current UnorderedMap is not used much and I've not seen it show up on any profiling. I think i'd need to know more about the use-case to comment on whether this was a good change.

**> rewritten BFFVariable to be usable directly within Arrays**

This is one I have in progress, though last time it didn't move the needle much on performance. I do think this is promising area to re-investigate though.

**> rewritten BFFVariable operations to operate on data in-place instead of creating a copy, mutating the copy, and assigning back to the variable**

Most operations should be using move semantics or operating in-place, but I would not be surprised if some cases are doing more work than is necessary. Fixes to that would make sense.

**> added BFFVariableScope to hold and index BFFVariables within a scope, and to only switch to using an UnorderedMap after growing past a certain size**

If I understand this correctly, the idea is to allow stack entries to be references to properties instead of copying them? If so, i think there is some merit to this, though I imagine the benefit of this is greatest when deep copies are expensive (i.e. the value of such a change is reduced if copy-on-write is in place.

**> reduced UnorderedMap bucket count from 65,536 to 256**

As above, the current implementation is not something I've seen show up in profiling, so it's hard to say if this is globally desirable.

StrangePan:
Thank you for your feedback! I plan on returning to this PR in a few weeks, maybe a month or two.

> **> added BFFVariableScope to hold and index BFFVariables within a scope, and to only switch to using an UnorderedMap after growing past a certain size**
> If I understand this correctly, the idea is to allow stack entries to be references to properties instead of copying them? If so, i think there is some merit to this, though I imagine the benefit of this is greatest when deep copies are expensive (i.e. the value of such a change is reduced if copy-on-write is in place.

Actually, this is so that BFFVariables can be indexed by their name for faster lookup and insertion times once a scope grows too large.

> If you could share a bit more about your bff file structure that might help direct performance work as well.

Sure thing, I can do so when I return to this later. I need to re-evaluate the new BFF config structure with a clear head.

StrangePan:
I've decided to drop these changes and focus on my own work. I hope they can serve as a useful reference if someone decides to implement some of the changes, but that someone won't be me. Thank you for your time and feedback.

PR #1192, opened 2026-06-05
CompilerNode: add EscapeSlashesInResponseFile setting

siukosev:
# Description:

Android NDK clang compiler on Windows incorrectly consumes paths from response file.
`C:\ws\file.c` becomes `C:wsfile.c`

Minimal repro:
```
#import NDK_ROOT

.Host = 'windows-x86_64'
.NDKToolchainBinRoot = '$NDK_ROOT$/toolchains/llvm/prebuilt/$Host$/bin'

Compiler('NDKWindowsClang')
{
    .Executable = '$NDKToolchainBinRoot$/clang.exe'
    .ForceResponseFile = true
    //.EscapeSlashesInResponseFile = true
}

.Compiler        = 'NDKWindowsClang'
.CompilerOptions = '-c %1 -o %2'

ObjectList('obj')
{
    .CompilerInputFiles = { 'func.c' }
    .CompilerOutputPath = 'out/'
}
```

Added `Compiler` param `EscapeSlashesInResponseFile`, used existing `Args::SetEscapeSlashesInResponseFile()`.

# Checklist:

The pull request:
- [x] **Is self-contained**


ffulin:
I think it would be nice to see if we can avoid the need for this new property by instead deriving it automatically after knowing that Clang is being used.

Dandielo:
I just force 'clang-orbis' as the compiler family for emscripten clang as it already cover that issue on Windows, but i think you could do the same here. 

However, i did run into problems on Linux with it. So in the end on Windows i set the value to clang-orbis and on Linux i leave it as clang

Maybe an improvement here would be to expose that setting as a compiler function argument, that is resolved by default for known use-cases?

ffulin:
We can alter the behavior on Windows vs other platforms if needed (and this does already happen in some cases).
We might also like to bake this into the "CompilerFamily" rather than having a separate option (and that may include improving compiler family detection)

In the end of course a unique control option can also be exposed, but it'd be ideal to treat that as a last resort: the more that works by default/automatically the better, at least to start with.

Issue #1063, opened 2025-01-12
Using fastbuild with MinGW-w64


brechtsanders:
I tried fastbuild on Windows using native MinGW-w64 (which can come with either GCC or LLVM build tools).

However it seems fastbuild always assumes Windows uses MSVC build tools.

For example in `Code/fbuild.bff` there is the assumption `$VS_PATH$` is set:
```raw
Settings
{
    #if __WINDOWS__
        #import TMP
        .Environment    = { "PATH=$VS_PATH$;$WINDOWS_SDK_2019_PATH$",
                            "TMP=$TMP$",
                            "SystemRoot=C:\Windows" }
    #endif
}
```

And in `External/SDK/Windows/Windows10SDK.bff` it looks for the Windows SDK in fixed location `C:\Program Files (x86)\Windows Kits` while MinGW-w64 provides it's own Windows SDK.

Any chance MinGW-w64 will be supported in the future?

I assume most of the work would be to allow detecting GCC and LLVM on Windows instead of only MSVC, and in the case where both exist maybe adding a command line flag or environment variable to decide which build tools to use in case multiple are available. A good place to start would be the `CC` and `CXX` environment variables.

ffulin:
There is a distinction between the compiler(s) used to compile FASTBuild binaries for development and release of FASTBuild itself, versus those it supports using to build your own code.

The bff you referenced appears to be from the FASTBuild source and is used to build FASTBuild itself. To contribute to or to modify FASTBuild on Windows will indeed require the MSVC compiler (at least for the most obvious path).

However, to compile your own code, you only need the compilers you wish to use. Many common compilers (MSVC, Clang, GCC etc) are supported, while compilers compatible derived from those can usually also be used with some small additional setup.

It should be possible to use Clang or GCC from a MinGW environment if you wish. You just need to configure your build to point to the appropriate things.

To compile a minimal test case of your own code, you generally need:
* A [Compiler](https://fastbuild.org/docs/functions/compiler.html)() to specify the compiler to use. Point this at the compiler you want to use. The "family" is usually detected automatically, but if not can be set explicitly.
* An [ObjectList](https://fastbuild.org/docs/functions/objectlist.html) or [Library](https://fastbuild.org/docs/functions/library.html) to specify which objects to compile. In a larger code base you would usually have many of these, usually equivalent to the granularity of your "libraries".
[...]

PR #1198, opened 2026-07-09
Add MinGW-w64 (GCC/Clang) to the Windows Hello World example

15daksh-2003:
# Description:

The Windows Hello World example was MSVC-only and dated, reinforcing the
impression that FASTBuild on Windows requires MSVC.

Reworked it into a multi-compiler example: modernizes the MSVC config to the
Compiler() function, adds a MinGW-w64 (GCC/Clang) config that needs no Visual
Studio / Windows SDK, and adds a short note clarifying that FASTBuild's own
self-build .bff files don't constrain user builds. Both configs were verified
to compile, link and run (MSVC VS2022 + Windows SDK 10.0.26100.0; MinGW-w64
g++ 15.2.0).

Fixes: #1063

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/Documentation/docs/examples/helloworld_windows.html
+++ b/Code/Tools/FBuild/Documentation/docs/examples/helloworld_windows.html
@@ -30,91 +30,192 @@
       Overview
     </div>
     <div class='newsitembody'>
-<p>A minimal configuration is shown for compiling a simple "Hello World" application on Windows.</p>
-<p>This examples illustrates:
+<p>A minimal configuration for compiling a simple "Hello World" application on Windows.</p>
+<p>The same program is shown built with two common Windows toolchains:</p>
 <ul>
-  <li>Compiling a directory of cpp files</li>
-  <li>Linking those object files into an Executable</li>
-  <li>Linking against the standard libraries</li>
+  <li><b>MSVC</b> - the Microsoft Visual C++ toolchain.</li>
+  <li><b>MinGW-w64</b> - native GCC or Clang, which ship their own headers, import
+      libraries and runtime.</li>
 </ul>
-To make the sample easier to understand, it is more verbose than is required.
-</p>
+<p>Each example illustrates:</p>
+<ul>
+  <li>Declaring a <a href='../functions/compiler.html'>Compiler</a>()</li>
+  <li>Compiling a directory of cpp files with an <a href='../functions/objectlist.html'>ObjectList</a>()</li>
+  <li>Linking the object files into an <a href='../functions/executable.html'>Executable</a>()</li>
+</ul>
+<p>The examples are deliberately explicit; real configurations usually factor shared
+settings into variables to reduce duplication.</p>
+</div>
+
+    <div class='newsitemheader'>Supported compilers on Windows</div>
+    <div class='newsitembody'>
+<p>FASTBuild supports the following compilers on Windows: <b>MSVC</b>, <b>Clang</b>
+(including <code>clang-cl</code>) and <b>GCC</b> - including the GCC and Clang toolchains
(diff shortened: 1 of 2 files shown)


15daksh-2003:
@ffulin  for engagement here

Code/Tools/FBuild/Documentation/docs/examples/helloworld_windows.html
```diff
@@ -30,91 +30,192 @@ <h1>Hello World (Windows)</h1>
...
+<p>FASTBuild supports the following compilers on Windows: <b>MSVC</b>, <b>Clang</b>
+(including <code>clang-cl</code>) and <b>GCC</b> - including the GCC and Clang toolchains
+that come with <b>MinGW-w64</b>. The compiler family is normally detected automatically
+from the executable name, and can be set explicitly via
+<a href='../functions/compiler.html'>.CompilerFamily</a> when needed.</p>
+<p><b>Note:</b> the <code>.bff</code> files in the FASTBuild <i>source</i> repository
+configure how FASTBuild builds <i>itself</i> (which uses MSVC on Windows). They do not
+constrain which compiler is used to build other projects - for that, point a
+<code>Compiler()</code> at the toolchain to use, as shown below.</p>
+<p>FASTBuild does not read the <code>CC</code>/<code>CXX</code> environment variables to
```
15daksh-2003, line 61: i am not sure if this is needed as part of this documentation...but I really found it helpful to understand how fastbuild does things, as opposed to expectations people carry coming from make, autotools, etc.

15daksh-2003:
@ffulin can u take a look at this when u get time

ffulin:
The "Hello World" example for Windows is quite outdated, so it'd certainly be nice to update that.
Similarly, it'd be nice to extend the minimal examples to cover more things (other platforms and compilers)

This PR goes some way towards that, but I think has some problems and raises some interesting questions about how this should perhaps evolve. I'll note all the things below so we can discuss.

**Minimal**
The change extends the minimal example to cover multiple different use cases. The new use case (MinGW) seems like a nice to have, but perhaps what might be best is to keep these examples isolated from each other and additional cases might become unwieldy. For example, perhaps these should be different files.
I'm generally thinking about how this structure might impact some of the issues below and how we think about resolving them.

**Forcing c++ runtime**
The changes add dependencies on the C++ runtime that are not necessary. I suppose it can be useful to see what extra things need to be done if you want to link the C++ runtime (on top of simply using C++) but it does complicate the examples a bit.

**Missing dependencies**
The changes to the config, specifically removing the explicit libs and removing /NODEFAULTLIB means the example will no longer correctly handle rebuilds when the standard library changes.
This problem seems to exist in the MinGW example as well.

**Unnecessary (?) changes**
The changes to the examples have some changes, such as the main signature, include path style, ordering of some options and renaming of targets.
Some or perhaps all of these seem unnecessary. It might be best to avoid unnecessary changes to avoid distracting from the necessary ones.

**LinkerType**
The MinGW seems to specify the linker type unnecessarily (g++ should already be detected as "gcc"). The comment in the bff seems to imply you had issues with that, but a cursory inspection shows the code seems to handle this properly.

**References to CMake-specific things**
You mentioned this yourself, but referenced to CMAKE specific concepts like CXX are a little out of place, since FASTBuild has no dependency on CMake, and in an ideal world would entirely replace it.
The ability of CMake to generate bff files now and provide options for people is great of course, and perhaps there can even be a place for documentation specifically for how to "migrate" or adjust to FASTBuild if familiar with other systems, but I think that probably doesn't belong in the minimal stand-alone example that the "hello world" is intended to be.

---

Please let me know what you think about the above.

Issue #1199, opened 2026-07-09
Dependencies::operator= self-wipes on self-assignment (copy and move)


15daksh-2003:
Both assignment operators in Dependencies.h destroy their own contents when the source and destination are the same object:

Copy-assign — body is Clear(); Add(other);. When &other == this, Clear() empties the object first, so Add copies from the now-empty source → the dependency list is lost.
Move-assign — body is Clear(); m_DependencyList = other.m_DependencyList; other.m_DependencyList = nullptr;. When &other == this, it frees/abandons its own storage and then nulls the pointer → the object is left empty.

Why it's an issue: 

This is a latent bug — no current caller does deps = deps (or deps = Move(deps)), so it never triggers today. More like protection against any future refactor, or an algorithm that assigns a container element to itself (e.g. during in-place sort/dedup/swap), would silently wipe the dependency list with no error or crash — producing an incorrect build graph that's hard to trace back to the assignment.

ffulin:
For more details about the resolution of this issue, please see the comments on you related PR

Please note that FASTBuild has adopted a no-AI policy going forward.

PR #1200, opened 2026-07-09
Dependencies: guard operator= against self-assignment

15daksh-2003:
# Description:

Add an if ( this != &other ) guard to both operators. Extend TestDependencies.OperatorEquals with populated self-copy and self-move cases asserting size, capacity, node order, stamps, and weak flags all survive.

Fixes: #1199 

# Checklist:

The pull request:
- [x] **Is self-contained**

### DIFF
--- a/Code/Tools/FBuild/FBuildCore/Graph/Dependencies.h
+++ b/Code/Tools/FBuild/FBuildCore/Graph/Dependencies.h
@@ -214,17 +214,23 @@ inline void Dependencies::Add( const Dependencies & deps )
 //------------------------------------------------------------------------------
 inline Dependencies & Dependencies::operator=( const Dependencies & other )
 {
-    Clear();
-    Add( other );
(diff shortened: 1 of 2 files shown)


15daksh-2003:
@ffulin  on engagement here

ffulin:
For general purpose containers it's often ideal to gracefully handle self assignment. For custom containers, especially those created explicitly to reduce overhead present in more generic containers, self assignment support may be unnecessary or even undesirable. That is the case for the Dependencies container. It is a specialized container for some code that is performance critical and is never used in a self-assignment.

I have added an assert in https://github.com/fastbuild/fastbuild/commit/edc664b9823e2bc9474df6635bd0b9f0fdb1e370 so that should this requirement be unintentionally violated in the future, it can be caught.

15daksh-2003:
the reasoning looks fair and makes sense to me. this was caught as part of static scan. 

although I do think the handling of PR could have been better...

PR #1201, opened 2026-07-11
Linux symlink fixes

Dandielo:
# Description:

I've found that FastBuild is currently failing to copy simlinks properly when used together with `-clean`, or when a full database rebuild is necessary. 

In addition to that, while reading the code I've also noticed that FastBuild updates the linked files' write time instead of the symlink itself.

# Checklist:

The pull request:
- [x] **Is self-contained**


Dandielo:
I'll close the pull-request as it seems I've taken additional history that was already submitted previously.

